PR #273, opened 2021-02-15
remove read_visium_data, add docs for segmentation features

hspitzer:
### Types of changes

### Description

- remove not needed and unused `squidpy/_io.py`
- add documentation for segmentation feature props

### How has this been tested?

.


michalk8:
Looks great! 2 minor pet peeves:
- I think here https://github.com/theislab/squidpy/pull/273/files#diff-825981da66fa773bb4837ca9fc5a9da899f85a1244ecf2ba6ae246ec648ef4c0R275 you can use :ref:\`Returns\`
- the list format should be
```
- `'area'` - number of pixels of segment.
```
, i.e. dash instead of color and `.` at the end.

hspitzer:
> Looks great! 2 minor pet peeves:
> 
> * I think here https://github.com/theislab/squidpy/pull/273/files#diff-825981da66fa773bb4837ca9fc5a9da899f85a1244ecf2ba6ae246ec648ef4c0R275 you can use :ref:`Returns`
> * the list format should be
> 
> ```
> - `'area'` - number of pixels of segment.
> ```
> 
> , i.e. dash instead of color and `.` at the end.

Ok, I have changed list formatting and added the ref to Returns. I was wondering if that exists, thanks! 
Do you have any reason for preferring `-` over `:`?

michalk8:
>  Do you have any reason for preferring - over :?

Personally, I think it creates more distinction between the value on the left and the description on the right (because of the 1 extra space on the left and that `-` is wider than `:`).

michalk8:
Thanks for the PR @hspitzer , looks great! I think it can be closed after the checks pass.

PR #280, opened 2021-02-16
Fix slow feature calculating with lazy dask image

hspitzer:
### Types of changes

### Description

When using a lazy image (loaded with `ImageContainer(..., chunks=<int>)`), calculation of features was very slow.
I think this is because numpy operations work much more slowly on lazy dask arrays. Adding a `crop._data = crop.data.load()` in `calculate_image_features` makes this much faster. We still keep the advantage of using lazy loading, because we only load the small image crop, right before it is being processed, and every pixel needs to be read anyways.
What do you think @giovp @michalk8 ?

I've also fixed `feature_summary` to not compute mean and std for every quantile.

### How has this been tested?

...

### Closes

...


michalk8:
Sounds good. Should this be also done for `squidpy.im.process` or `squidpy.im.segment`? If yes, how to determine when to use lazy and when not - e.g. depending on the size of the crops, but what would be good values, maybe lazy if > 1M pixels?

giovp:
looks great! If needed in process and segment would also add it there. Re finding optimal value, I woudl wait for this, actually dask probably have something simialr for on-the-fluy processing (also from your last email): https://github.com/dask/dask-image

let's get the release done and then discuss next steps (already a crazy long list :D)

giovp (approved):

michalk8:
Agree, let's merge (but first please #273 , since this contains also the stuff).

giovp:
I think we can merge this when the check passes!

hspitzer:
Ok, I'm merging this then. I think for `process` and `segment` it might be useful as well, but I'm not sure what sensible default values would be. Lets discuss this later in a separate issue!

PR #283, opened 2021-02-16
formulation suggestions

SabrinaRichter:
changed few formulations in the README


giovp:
really like it! We'd need the same in squidpy_notebooks though because the readme are the same. 
and also in index.rst https://github.com/theislab/squidpy/blob/master/docs/source/index.rst

could you make the same changes there? would be really nice, thank you!

SabrinaRichter:
done

PR #286, opened 2021-02-18
Delegate del

### LINKED ISSUE #285: Support deletion of single layers in Image Container
### Type of the feature

- [ ] Other?

### Description
Support deletion of images from image container like 
``` del img['image']```

michalk8:
### Types of changes

### Description

Delegate `del` in ImageContainer.

### How has this been tested?

1 new test for the deletion.

### Closes

closes #285 

In the future, we should consider using a descriptor to delegate other `xarray.Dataset` methods/attributes.

### DIFF
--- a/squidpy/im/_container.py
+++ b/squidpy/im/_container.py
@@ -948,6 +948,9 @@ class ImageContainer(FeatureMixin):
 
         return y, x
 
+    def __delitem__(self, key: str) -> None:
+        del self.data[key]
+
     def __iter__(self) -> Iterator[str]:
         yield from self.data.keys()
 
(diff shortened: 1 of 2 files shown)


hspitzer:
Thanks for this Michal! Do we need to take care that we don't delete the only image from the container? Or can all the subsequent `ImageContainer` functions deal with an empty container?

michalk8:
> Thanks for this Michal! Do we need to take care that we don't delete the only image from the container? Or can all the subsequent ImageContainer functions deal with an empty container?

No, that should be fine, I check in many places for that using: https://github.com/theislab/squidpy/blob/master/squidpy/im/_container.py#L921

Merging this.

PR #294, opened 2021-02-24
remove `githup_repo` arg from conf sphinx

giovp:
### Types of changes

### Description

...

### How has this been tested?

...

### Closes

...

### DIFF
--- a/docs/source/_ext/edit_on_github.py
+++ b/docs/source/_ext/edit_on_github.py
@@ -1,9 +1,20 @@
 """Based on gist.github.com/MantasVaitkunas/7c16de233812adcb7028."""
 from typing import Any, Dict, Optional
 from sphinx.application import Sphinx
+import os
 import warnings
 
 __licence__ = "BSD (3 clause)"
+
+
+def get_github_repo(app: Sphinx, path: str) -> str:
+    if path.endswith(".ipynb"):
+        return str(app.config.github_nb_repo)
+    if "auto_examples" in path:
+        return str(app.config.github_nb_repo)
+    if "auto_tutorials" in path:
+        return str(app.config.github_nb_repo)
+    return str(app.config.github_repo)
 
 
 def _html_page_context(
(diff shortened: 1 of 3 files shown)


michalk8:
@giovp I've fixed it so that the links to tutorials/examples/rst are correct.
However, for auto-generated files by sphinx, such as `https://squidpy.readthedocs.io/en/latest/api/squidpy.gr.centrality_scores.html#squidpy.gr.centrality_scores` which are not commited to the repo, it's still broken.
I will fix it later - I remember scVelo has this: https://github.com/theislab/scvelo/blob/master/docs/source/conf.py#L224
and we can have a simplified version (without the line from, line to, since we already use an extension for it).

michalk8 (approved):

giovp:
> I will fix it later - I remember scVelo has this: https://github.com/theislab/scvelo/blob/master/docs/source/conf.py#L224
and we can have a simplified version (without the line from, line to, since we already use an extension for it).

those are really fine, not pressing. Thanks for addressing this real quick!

Have a related question actually, I know we discussed this already but I noticed the repo is ~50 Mb and a lot comes from the notebooks (~46 Mb are in  `.git` ). This is because the way we set up running and rendering the notebooks right? There is no workaround to save them there and render them on the docs from there? sorry if repeating myself

michalk8:
@giovp this is because we still have that old branch https://github.com/theislab/squidpy/tree/feature_scaling2,
the repo cleaner did its job. I've created my own fork and deleted the branch, then the .git/objects are only 3MB.
@LouisK92 could you please recreate the changes from on branch based-off the current master and delete the old one? That should fix it.

Regarding the repo size of `squidpy_notebooks`, unfortunately, there we commit the generated .ipynb so that they are downloadable/viewable through binder. There are 63MB in the .git/objects, 39MB are in the docs (mostly matplotlib figures).
A workaround would be to upload all generated stuff as artifacts on github (e.g. as a compressed tar), instead of commiting them and then download these when building the main docs (should be fairly easy). Downside is that we'd lose links to binder + unforseen consequences (not sure about rate limits and retention policy).

LouisK92:
deleted the branch (and made a copy of the changes). Is it fixed?

michalk8:
Yes, it's fixed (just have to reclone), thanks a lot.

PR #302, opened 2021-03-05
Fixes for interaction_matrix

ivirshup:
### Types of changes

### Description

Reimplements `interaction_matrix` as discussed in 301. Additionally added `weight` parameter so graph weights can be used here.

Added checks in tests for correctness of weighted and un-weighted versions.

Current solution does not work if there are null values in the categorical, but the previous one wouldn't either. Not sure how this should be handled. Would null values not be included, or be considered their own category? Encoding functions in scikit-learn seem to have different behavior around this, but it might be nice to just pick one and use it's behavior.

### Closes

Fixes 301


michalk8:
Thanks a lot @ivirshup , looks good!

> Not sure how this should be handled. Would null values not be included, or be considered their own category?

I think `NaN` should just be excluded (if they are not in `.cat.categories`), so I'd just subset all non-NaN. @giovp any preference for this?

giovp:
> I think NaN should just be excluded (if they are not in .cat.categories), so I'd just subset all non-NaN. @giovp any preference for this?

Agree on this. 
Thanks a lot @ivirshup for the PR! I think we could get reimplement the other networkx metrics we use with numba as well actually, something to think about for later releases.

Tests are failing because of plotting, I should have time to address this next week for sure, sorry for delay

michalk8:
I've fixed the plotting tests, merging this, thanks again @ivirshup .

PR #304, opened 2021-03-10
refactor `sq.gr.moran`, remove pysal depend

giovp:
### Types of changes

### Description
Re-implement Moran's I to remove libpysal dependency

### How has this been tested?
not yet

### Closes

TODO:

- [ ] add row wise transform?
- [ ] improve speed ?
- [ ] change docs and reference

@michalk8 implementation is really trivial see [wiki](https://en.wikipedia.org/wiki/Moran%27s_I) and original code https://github.com/pysal/esda/blob/642880e0b49913de07e5b7fafb3dd0b5855ed2ef/esda/moran.py#L234

unfortunately, it's slower than original, not sure why. one interpretation could be that numba vectorized dot product is still slower than `np.dot` and the gain in permutation is lost. do you see any major performance issues? otherwise could also revert to plain numpy.
anyway, this will allow us to remove pysal+esda from dependency. top 100 genes by pysal are consistent
image

### DIFF
--- a/squidpy/gr/_ppatterns.py
+++ b/squidpy/gr/_ppatterns.py
@@ -1,7 +1,6 @@
 """Functions for point patterns spatial statistics."""
 from typing import Any, Tuple, Union, Iterable, Optional, Sequence
 from itertools import chain
-import warnings
 
 from scanpy import logging as logg
 from anndata import AnnData
@@ -25,18 +24,6 @@ from squidpy.gr._utils import (
     _assert_non_empty_sequence,
 )
 from squidpy._constants._pkg_constants import Key
-
-try:
-    with warnings.catch_warnings():
-        warnings.simplefilter("ignore", UserWarning)
-        from libpysal.weights import W
-        import esda
-        import libpysal
-except ImportError:
-    esda = None
-    libpysal = None
-    W = None
-
 
 __all__ = ["ripley_k", "moran", "co_occurrence"]
 
--- a/squidpy/gr/_ppatterns.py
+++ b/squidpy/gr/_ppatterns.py
@@ -254,7 +254,6 @@ def _compute_moran(
         i_indices = indices[s]
         i_data = data[s]
         zl[i] = np.sum(i_data * z[i_indices])
-    # zl = adj * z
     inum = (z * zl).sum()
     z2ss = (z * z).sum()
     return len(counts) / data.sum() * inum / z2ss
--- a/squidpy/gr/_ppatterns.py
+++ b/squidpy/gr/_ppatterns.py
@@ -218,7 +218,7 @@ def _moran_helper(
 
     moran_list = []
     for g in gen:
-        counts = adata.obs_vector(g, layer=layer).copy().astype(fp)
+        counts = adata.obs_vector(g, layer=layer).astype(fp, copy=True)
         indptr = adj.indptr.astype(ip)
         indices = adj.indices.astype(ip)
         data = adj.data.astype(fp)
--- a/squidpy/gr/_ppatterns.py
+++ b/squidpy/gr/_ppatterns.py
@@ -28,11 +28,11 @@ from squidpy._constants._pkg_constants import Key
 __all__ = ["ripley_k", "moran", "co_occurrence"]
 
 
-it = nt.int32
-ft = nt.float32
+it = nt.int64
+ft = nt.float64
 tt = nt.UniTuple
-ip = np.int32
-fp = np.float32
+ip = np.int64
+fp = np.float64
 
 
 @d.dedent
@@ -235,7 +235,7 @@ def _moran_helper(
 
 
 @njit(
-    ft(ft[:], it[:], it[:], ft[:]),
+    ft(ft[:], it[:], it[:], ft[:], ft[:], ft, ft),
     parallel=False,
     fastmath=True,
 )
@@ -244,9 +244,11 @@ def _compute_moran(
     indptr: np.ndarray,
     indices: np.ndarray,
     data: np.ndarray,
+    z: np.ndarray,
+    data_sum: fp,
+    z2ss: fp,
 ) -> Any:
 
-    z = counts - counts.mean()
     zl = np.empty(len(z))
     N = len(indptr) - 1
     for i in range(N):
@@ -255,8 +257,8 @@ def _compute_moran(
         i_data = data[s]
         zl[i] = np.sum(i_data * z[i_indices])
     inum = (z * zl).sum()
-    z2ss = (z * z).sum()
-    return len(counts) / data.sum() * inum / z2ss
+
+    return len(counts) / data_sum * inum / z2ss
 
 
 @njit(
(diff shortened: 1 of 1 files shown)


squidpy/gr/_ppatterns.py
```diff
@@ -251,20 +234,54 @@ def _moran_helper(
...
+    z2ss = (z * z).sum()
+    return len(counts) / data.sum() * inum / z2ss
 
 
-def _set_weight_class(adata: AnnData, key: str) -> W:
-    X = adata.obsp[key]
-    if not isspmatrix_lil(X):
-        X = X.tolil()
+@njit(
+    ft[:](ft[:], it[:], it[:], ft[:], it),
```
ivirshup, line 263: Indices of sparse matrices are not always `int32`, if you have enough entries they will become `int64`.

squidpy/gr/_ppatterns.py
```diff
@@ -251,20 +234,54 @@ def _moran_helper(
     return pd.DataFrame(moran_list, columns=["I", "pval_sim", "VI_sim"], index=gen)
 
 
-def _compute_moran(y: np.ndarray, w: W, transformation: str, permutations: int) -> Tuple[float, float, float]:
-    mi = esda.moran.Moran(y, w, transformation=transformation, permutations=permutations)
-    return mi.I, mi.p_z_sim, mi.VI_sim
+@njit(
+    ft(ft[:], it[:], it[:], ft[:]),
+    parallel=False,
```
ivirshup, line 239: In theory, is parallelizable, but I've run into some issues with this recently

squidpy/gr/_ppatterns.py
ivirshup: Also, probably good to collect to `float64` if you're expecting a lot of samples.

squidpy/gr/_ppatterns.py
```diff
@@ -239,7 +218,11 @@ def _moran_helper(
 
     moran_list = []
     for g in gen:
-        mi = _compute_moran(adata.obs_vector(g, layer=layer), weights, transformation, permutations)
+        counts = adata.obs_vector(g, layer=layer).copy().astype(fp)
```
ivirshup, line 221:
```suggestion
        counts = adata.obs_vector(g, layer=layer).astype(fp, copy=True)
```

ivirshup (commented):
Hey, saw this and had a couple points.

I wouldn't be surprised if the overhead of joblib was significant here. I could give a few more pointers on performance if you want (I'd have to look closer, but I think there's a few computations that are repeated more than needed).

Any relationship to https://github.com/theislab/scanpy/issues/1698?

giovp:
thanks a lot @ivirshup ! so re performance, I calculated including joblib in both cases (their original implementation and mine with numba). It's slower nonetheless. I was wondering if there is some improvement I could do on the numba code itself, I think using joblib is necessary anyway to parallelize across genes.

re theislab/scanpy#1698 yeah thanks for pointing it out, so one option is to submit this PR directly there, and wrap `scanpy.metrics` here. Could also wrap your implementation of geary's C there if you agree. 

would like to get feedback on this from @michalk8 @hspitzer : shall we move this to scanpy metrics module and wrap it as moran here?

squidpy/gr/_ppatterns.py
```diff
@@ -251,20 +234,54 @@ def _moran_helper(
...
+    zl = np.empty(len(z))
+    N = len(indptr) - 1
+    for i in range(N):
+        s = slice(indptr[i], indptr[i + 1])
+        i_indices = indices[s]
+        i_data = data[s]
+        zl[i] = np.sum(i_data * z[i_indices])
+    inum = (z * zl).sum()
+    z2ss = (z * z).sum()
+    return len(counts) / data.sum() * inum / z2ss
```
ivirshup, line 259: `counts - counts.mean()`, `z2ss = (z * z).sum()`, and `data.sum()` should be the same w.r.t. each permutation, so you can probably compute these outside the permutation loop and pass them in.

ivirshup (commented):
I've commented a bit more on the performance, basically some things that you're computing each permutation that I think should be constant across permutations.

Could you share the cases you're using to benchmark?

ivirshup:
> re scanpy#1698 yeah thanks for pointing it out, so one option is to submit this PR directly there, and wrap scanpy.metrics here. Could also wrap your implementation of geary's C there if you agree.

Whatever works for you. I think it could also make sense to have the code live both places for a bit depending on how you feel about requiring newest scanpy/ release schedules.

squidpy/gr/_ppatterns.py
giovp: done, thank you!

squidpy/gr/_ppatterns.py
giovp: ok interesting, but didn't see much improvement with parallel anyway

squidpy/gr/_ppatterns.py
giovp: done, thank you!

squidpy/gr/_ppatterns.py
giovp: good point, done but still not much better

giovp:
```python
genes = adata[:, adata.var.highly_variable].var_names.values[0:100]

# numba
sq.gr.moran(adata, genes=genes, n_jobs=1)
# time 01:30
moran_numba = adata.uns["moranI"].copy()

# switch to master, pysal
sq.gr.moran(adata, genes=genes, n_jobs=1)
# time 00:18
moran_pysal = adata.uns["moranI"].copy()

g = sns.scatterplot(moran_pysal["I"], moran_numba["I"])
g.set_xlabel("pysal")
g.set_ylabel("numba")
```
image

re moving to scanpy, brief talk with @hspitzer and she is fine with moving it there, I think with the ratio of scanpy being core of squidpy it makes sense. 
We would than wrap both this and geary's c in `sq.gr.moran` and `sq.gr.gearysc` for convenience to handle spatial neighbors. How does it sounds @ivirshup @michalk8 ?

squidpy/gr/_ppatterns.py
```diff
@@ -239,7 +218,11 @@ def _moran_helper(
 
     moran_list = []
     for g in gen:
-        mi = _compute_moran(adata.obs_vector(g, layer=layer), weights, transformation, permutations)
+        counts = adata.obs_vector(g, layer=layer).astype(fp, copy=True)
```
michalk8, line 221: This can be moved out of the loop.

squidpy/gr/_ppatterns.py
```diff
@@ -251,20 +234,61 @@ def _moran_helper(
...
 
+    perms = np.empty(n_perms, dtype=ft)
+    res = np.empty(3, dtype=ft)
 
-def _set_weight_class(adata: AnnData, key: str) -> W:
-    X = adata.obsp[key]
-    if not isspmatrix_lil(X):
-        X = X.tolil()
+    z = counts - counts.mean()
+    data_sum = data.sum()
```
michalk8, line 281: `data_sum` can be computed in the caller.

michalk8:
>  We would than wrap both this and geary's c in sq.gr.moran and sq.gr.gearysc for convenience to handle spatial neighbors. How does it sounds @ivirshup @michalk8 ?

Sounds good to me. In terms of performance, 2 things come to mind:
- look at the esda implementation
- switch the joblib par. loops (perms/genes instead of genes/perms)
- maybe try implementing the crucial parts in Cython? not sure if there would be a speed-up

giovp:
this was addressed in #317

PR #313, opened 2021-03-24
init sepal

giovp:
Hi @almaan , as mentioned per email, here's a first stab at implementing sepal in squidpy. I decided to start from scratch here in order to access standard anndata data format:
- sparse spatial graph
- sparse counts

and implemented core functions in numba. 
I think leveraging both sparse representation and numba would be best down the line in order to run the method on large spatial data. However, this is only a first try and would really appreciate any feedback/contribution from your side! I might have missed core ingredients of the implementation.

Here you can find scripts to benchmark this implementation against original. You should be able to convert them to notebook or run them as is, just need to clone the repo, switch to this branch, and install dev version `pip install -e'.[dev,test]'`

<details>
<summary>Synthetic data</summary>

```python
# ---
# jupyter:
#   jupytext:
#     text_representation:
#       extension: .py
#       format_name: percent
#       format_version: '1.3'
#       jupytext_version: 1.11.0
#   kernelspec:
#     display_name: Python 3
#     language: python
#     name: python3
# ---

# %%
import numpy as np
import scanpy as sc
import squidpy as sq
from anndata import AnnData
import sepal
import sepal.datasets as d
import sepal.models as m
import sepal.utils as ut
sc.logging.print_header()
sq.__version__
# %load_ext autoreload
# %autoreload 2
# %load_ext lab_black
pth = "./sepal/data/synthetic/image-based.tsv"
raw_data = d.RawData(
    pth,
)
raw_data.cnt = ut.filter_genes(raw_data.cnt, min_expr=10, min_occur=5)
data = m.ST1K(raw_data, eps=0.1)

# %%
times = m.propagate(data, normalize=True, scale=False, num_workers=8)

# %%
snidx = data.get_satnbr_idx(data.saturated)
unidx = data.get_unsatnbr_idx(data.unsaturated)

# %%
coords = [i.split("x") for i in raw_data.cnt.index.values]
spatial = np.stack(coords).astype(int)

# %%
spatial.shape

# %%
ncnt = raw_data.cnt.values
ncnt = ut.normalize_expression(ncnt, c=2)
colMax = np.max(np.abs(ncnt), axis=0).reshape(1, -1)
ncnt = np.divide(ncnt, colMax, where=colMax > 0)
ncnt = ncnt.astype(float)

# %%
adata = AnnData(
    ncnt,
)
adata.obs.index = raw_data.cnt.index
adata.var.index = raw_data.cnt.columns
adata.obsm["spatial"] = spatial

# %%
sc.pp.pca(adata)
sc.pp.neighbors(adata)
sc.tl.umap(adata)
sc.tl.leiden(adata)

# %%
adata.obs["unsaturated"] = [1 if (i in unidx) else 0 for i in np.arange(adata.shape[0])]
adata.obs["saturated"] = [1 if (i in snidx) else 0 for i in np.arange(adata.shape[0])]

# %%
sq.gr.spatial_neighbors(adata, radius=1)

# %%
sc.pl.spatial(
    adata,
    spot_size=0.5,
    color=["leiden", "unsaturated", "saturated", "P0-O"],
    neighbors_key="spatial_neighbors",
    edges=True,
    edges_width=4,
    frameon=False,
)

# %%
from squidpy.gr._sepal import compute_idxs
g = adata.obsp["spatial_connectivities"]
spatial = adata.obsm["spatial"]
sat, sat_idx, unsat, unsat_idx = compute_idxs(g, spatial, 4, "l1")

# %%
# idxs are equal
np.testing.assert_array_equal(sat_idx.sum(axis=1), snidx.sum(axis=1))
np.testing.assert_array_equal(unsat_idx, unidx)

# %%
from squidpy.gr._sepal import _diffusion
genes = np.zeros(times.shape[0])
for i in range(genes.shape[0]):
    conc = adata.obs_vector(times.index.values[i], layer=None).copy()
    idx = _diffusion(
        conc,
        2000,
        sat,
        sat_idx,
        sat.shape[0],
        unsat,
        unsat_idx,
        conc.shape[0],
    )
    if idx.shape[0] != 0:
        genes[i] = 0.001 * idx[0]
    else:
        genes[i] = 0

# %%
from scipy.stats import pearsonr
import seaborn as sns
g = sns.scatterplot(x=genes, y=times["average"])
g.set_xlabel("current")
g.set_ylabel("sepal")
p = pearsonr(genes, times["average"])[0]
g.set_title(f"pearson_{round(p, 5)}")
```

image

image

</details>

<details>
<summary>Melanoma data</summary>

```python
# ---
# jupyter:
#   jupytext:
#     text_representation:
#       extension: .py
#       format_name: percent
#       format_version: '1.3'
#       jupytext_version: 1.11.0
#   kernelspec:
#     display_name: Python 3
#     language: python
#     name: python3
# ---

# %%
import numpy as np
import scanpy as sc
import squidpy as sq
from anndata import AnnData
import sepal
import sepal.datasets as d
import sepal.models as m
import sepal.utils as ut
sc.logging.print_header()
sq.__version__
# %load_ext autoreload
# %autoreload 2
# %load_ext lab_black
pth = "./sepal/data/real/melanoma.tsv.gz"
raw_data = d.RawData(
    pth,
)
raw_data.cnt = ut.filter_genes(raw_data.cnt, min_expr=10, min_occur=5)
data = m.ST1K(raw_data, eps=0.1)

# %%
times = m.propagate(data, normalize=True, scale=False, num_workers=8)

# %%
snidx = data.get_satnbr_idx(data.saturated)
unidx = data.get_unsatnbr_idx(data.unsaturated)

# %%
coords = [i.split("x") for i in raw_data.cnt.index.values]
spatial = np.stack(coords).astype(int)

# %%
spatial.shape

# %%
ncnt = raw_data.cnt.values
ncnt = ut.normalize_expression(ncnt, c=2)
colMax = np.max(np.abs(ncnt), axis=0).reshape(1, -1)
ncnt = np.divide(ncnt, colMax, where=colMax > 0)
ncnt = ncnt.astype(float)

# %%
adata = AnnData(
    ncnt,
)
adata.obs.index = raw_data.cnt.index
adata.var.index = raw_data.cnt.columns
adata.obsm["spatial"] = spatial

# %%
sc.pp.pca(adata)
sc.pp.neighbors(adata)
sc.tl.umap(adata)
sc.tl.leiden(adata)

# %%
adata.obs["unsaturated"] = [1 if (i in unidx) else 0 for i in np.arange(adata.shape[0])]
adata.obs["saturated"] = [1 if (i in snidx) else 0 for i in np.arange(adata.shape[0])]

# %%
sq.gr.spatial_neighbors(adata, radius=1)

# %%
sc.pl.spatial(
    adata,
    spot_size=0.5,
    color=["leiden", "unsaturated", "saturated", "ANXA2"],
    neighbors_key="spatial_neighbors",
    edges=True,
    edges_width=4,
    frameon=False,
)

# %%
from squidpy.gr._sepal import compute_idxs
g = adata.obsp["spatial_connectivities"]
spatial = adata.obsm["spatial"]
sat, sat_idx, unsat, unsat_idx = compute_idxs(g, spatial, 4, "l1")

# %%
# idxs are equal
np.testing.assert_array_equal(sat_idx.sum(axis=1), snidx.sum(axis=1))
np.testing.assert_array_equal(unsat_idx, unidx) # these are not, output: Mismatched elements: 14 / 76 (18.4%)

# %%
from squidpy.gr._sepal import _diffusion
# this takes ~15 min, but ok will be parallelized
genes = np.zeros(times.shape[0]) 
for i in range(genes.shape[0]):
    conc = adata.obs_vector(times.index.values[i], layer=None).copy()
    idx = _diffusion(
        conc,
        2000,
        sat,
        sat_idx,
        sat.shape[0],
        unsat,
        unsat_idx,
        conc.shape[0],
    )
    if idx.shape[0] != 0:
        genes[i] = 0.001 * idx[0]
    else:
        genes[i] = 0

# %%
from scipy.stats import pearsonr
import seaborn as sns
gt = times["average"]
g = sns.scatterplot(x=genes, y=gt)
g.set_xlabel("current")
g.set_ylabel("sepal")
p = pearsonr(genes, gt)[0]
g.set_title(f"pearson_{round(p, 5)}")
```

image

image

</details>

Results are broadly reproducible, although not perfect for the `melanoma` data. I think the reason might be the way the nearest saturated nodes are assigned to unsaturated neighbors. What I have now is, for each unsaturated node:
- search if any nhood node is saturated, if yes, assign the first.
- if all nhood of unsaturated node are unsaturated, then compute distance matrix and take min dist saturated. Doing this step only for a bunch of unsaturated makes it more mem efficient, as usually there will be few nodes still unsaturated.
However, in the case of melanoma, the indexes don't match, and I fear that's a large source of discrepancy. Do you think this is a major problem?

I also have a question re laplacian on non-grid graphs. The solution proposed is to build a grid in the tissue area and then assign gexp to the grid. I wonder if there is another solution that doesn't include the linear program part? maybe too much 😅  but would be cool to think about it !

Thanks again for any input!


almaan:
Hi @giovp,

this is nothing but awesome! Amazingly fast and neat work. I've browsed through the code today and also looked at some of the results for the melanoma samples more carefully, and I think your way of selecting saturated neighbors for unsaturated data points actually is superior to my original implementation! 

I also ran through the notebooks you sent and it all makes sense to me, the only thing I noted was that the plots illustrating the saturated vs. unsaturated spots looks a bit weird, but this can be remedied by changing:

```python
adata.obs["unsaturated"] = [1 if (i in unidx) else 0 for i in np.arange(adata.shape[0])]
adata.obs["saturated"] = [1 if (i in snidx) else 0 for i in np.arange(adata.shape[0])]
```
to
```python
adata.obs["unsaturated"] = [1 if (i in data.unsaturated) else 0 for i in np.arange(adata.shape[0])]
adata.obs["saturated"] = [1 if (i in data.saturated) else 0 for i in np.arange(adata.shape[0])]
```
as you've already done a great job with the porting, there's not much to remark on in the code itself. I just have a few comments:

###  Using np.diff instead of for loop

One minor thing I spotted was that the function `get_sat_unsat_idx`

now given as:

```python
@njit(parallel=False)
def _get_sat_unsat_idx(
    g_indptr: np.ndarray, g_indices: np.ndarray, g_shape: int, sat_thres: int
) -> Tuple[np.ndarray, np.ndarray]:
    """ Get saturated and unsaturated nodes based on thres. """

     n_indices = np.zeros(g_shape)
     for i in np.arange(g_shape):
         s = slice(g_indptr[i], g_indptr[i + 1])
        n_indices[i] = g_indices[s].shape[0]
    unsat = np.arange(g_shape)[n_indices < sat_thres]
    sat = np.arange(g_shape)[n_indices == sat_thres]

    return sat, unsat

```
could, I believe, be simplified to:

```python
@njit(parallel=False)
def _get_sat_unsat_idx(
    g_indptr: np.ndarray, sat_thres: int
) -> Tuple[np.ndarray, np.ndarray]:
    """ Get saturated and unsaturated nodes based on thres. """

    n_indices = np.diff(g_indptr)
    unsat = np.arange(g_shape)[n_indices < sat_thres]
    sat = np.arange(g_shape)[n_indices == sat_thres]

    return sat, unsat

```

However, you might have a reason with the initial suggestions that I'm not aware of so I haven't turned this into a PR yet, but rather highlighted it here.

### Variable number of neighbors

I would also suggest that the number of neighbors is given as an argument to the function `_get_nhood_idx` as this varies between ST1K and Visium (even though I believe the small corpus of ST1K data out there will soon be dwarfed by the abundance of Visium samples).  This could then be used in the construction of the `sat_idx` array by doing something like `    sat_idx = np.zeros((sat.shape[0], n_nbrs))`.

### Different Array support

Maybe you already have something in mind here, but the `_diffusion` function could be expanded to support the analysis of different platforms, which somehow would include the choice of an appropriate laplacian and also the number of neighbors to be used.

<hr>

To answer some of your questions:

### Non-structured data
I must be honest here, that platforms with non-structured data were maybe not my "target audience" when developing this method, and hence slightly less focus was devoted to this data type. However, I did try using a "greedy" approach; creating a grid and then first assigning the data point with the shortest distance to it's nearest grid point and so forth until all points were assigned. This did not give very robust results, as some datapoints were moved _far away_ from their original sites.

I can totally see why you would like to avoid the LP approach, it's quite ineffective from a computational perspective. Perhaps an alternative solution would be to use interpolation instead, and approximate the gene expression at the artificial grid points by a weighted nearest neighbor function or using spline interpolation. That would be a fairly simple solution, of course one could try to make something more advanced like a using something like FEM (Fine Element Method) to approximate the diffusion equation. If you have other ideas, I'm happy to discuss this further!

### Exclusion of non-zero condition in the diffusion function

I also saw your question about removing the condition that the concentration must be larger than zero in the `_diffusion` function. This should be fine imo, since you now use a predetermined finite set of iterations.

<hr>

Let me know if I can contribute in any way, or perhaps help in writing up a more seamless API for the use of sepal, you obv. have all the components there already, but I'm more than happy to help out!

Again, such great work - it's so fun to see this cast in the squidpy format!

giovp:
Hi @almaan ,

thanks a ton for the remarks, after very long I picked it up again, now should be closed to be wrapped up. Here's to reply some of your comments and couple of open questions:

> Using np.diff instead of for loop

great catch! added it.

> Variable number of neighbors
> I would also suggest that the number of neighbors is given as an argument to the function _get_nhood_idx as this varies between ST1K and Visium (even though I believe the small corpus of ST1K data out there will soon be dwarfed by the abundance of Visium samples). This could then be used in the construction of the sat_idx array by doing something like sat_idx = np.zeros((sat.shape[0], n_nbrs)).

> Different Array support
> Maybe you already have something in mind here, but the _diffusion function could be expanded to support the analysis of different platforms, which somehow would include the choice of an appropriate laplacian and also the number of neighbors to be used.

Indeed, now I've ported over the `hex_laplacian` as well, works for both square and hex grids, as original. I think the square-like grid is useful not only for old ST but also for new tech like [Dbit-seq](https://www.cell.com/cell/pdf/S0092-8674(20)31390-8.pdf) and [stereoseq](https://www.biorxiv.org/content/10.1101/2021.01.17.427004v2.full) .

> Exclusion of non-zero condition in the diffusion function
> I also saw your question about removing the condition that the concentration must be larger than zero in the _diffusion function. This should be fine imo, since you now use a predetermined finite set of iterations.

ok, removed.

> Non-structured data
> I must be honest here, that platforms with non-structured data were maybe not my "target audience" when developing this method, and hence slightly less focus was devoted to this data type. However, I did try using a "greedy" approach; creating a grid and then first assigning the data point with the shortest distance to its nearest grid point and so forth until all points were assigned. This did not give very robust results, as some datapoints were moved far away from their original sites.

> I can totally see why you would like to avoid the LP approach, it's quite ineffective from a computational perspective. Perhaps an alternative solution would be to use interpolation instead, and approximate the gene expression at the artificial grid points by a weighted nearest neighbor function or using spline interpolation. That would be a fairly simple solution, of course, one could try to make something more advanced like FEM (Fine Element Method) to approximate the diffusion equation. If you have other ideas, I'm happy to discuss this further!

so after some thinking, with this PR we'll drop the support to non grid-like graphs. However, with @MxMstrmn we've been thinking a bit about how to generalize the proposed diffusion process. We got some early results but not yet super sure about it, would be really nice to catch up and get your feedback on it. I'll let you know as soon as we have something to share!

I've also copied over the notebooks to the reproducibility folder, can be found here https://github.com/theislab/squidpy_reproducibility/pull/13
----------------

#### Remaining to do
 - [ ] @almaan would really appreciate your feedback on docstrings, arguments and defaults. I'm not 100% I've chosen the right words to describe the parameters and maybe the one-line description of the function could be improved? thanks in advance if you have the time to take a look.
 - [ ] tests

giovp:
@almaan just reminder if you could give a quick skim to the docstring, whether you think sepal is well explained and arguments are correct 🙏 

I've also played around a bit with changing `dt`/`threshold`, noticed that for visium a large part of high spatially variable genes are genes with very low expression (e.g. only few spots expressing) and tried to find a better heuristic. Did not really managed. If you have any suggestion much appreciated.
Thank you in advance for the help !

also, ready for review @michalk8 , thank you!

squidpy/gr/_sepal.py
```diff
@@ -0,0 +1,341 @@
...
+
+        dcdt = np.zeros(conc_shape)
+        dcdt[sat] = D * d2
+        conc[sat] += dcdt[sat] * dt
+        conc[unsat] += dcdt[unsat_idx] * dt
+        # set values below zero to 0
+        conc[conc < 0] = 0
+        # compute entropy
+        ent[i + 1] = _entropy(conc[sat]) / sat_shape
+        entropy_arr[i] = np.abs(ent[i + 1] - ent[i])  # estimate entropy difference
```
michalk8, line 228: There's no need to store the whole `ent` array, since it's just used to calculate the diff, which is the only thing returned.
You can save space just by saving `ent: float` and `prev_env: float`.

squidpy/gr/_sepal.py
```diff
@@ -0,0 +1,341 @@
...
+def _compute_idxs(
+    g: spmatrix, spatial: np.ndarray, sat_thres: int, metric: str = "l1"
+) -> Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
+    """Get saturated and unsaturated nodes and nhood indexes."""
+    sat, unsat = _get_sat_unsat_idx(g.indptr, g.shape[0], sat_thres)
+
+    sat_idx, nearest_sat, un_unsat = _get_nhood_idx(sat, unsat, g.indptr, g.indices, sat_thres)
+
+    # compute dist btwn remaining unsat and all sat
+    dist = pairwise_distances(spatial[un_unsat], spatial[sat], metric=metric)
```
michalk8, line 286: Maybe `force_all_finite=...` from https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise_distances.html
could be set to something other than default (e.g. allowing NaN and/but no infinities + use, np.nanargmin later).
Just a thought, not sure if makes sense.

squidpy/gr/_sepal.py
```diff
@@ -0,0 +1,341 @@
...
+    """Get saturated and unsaturated nhood indexes."""
+    # get saturated nhood indices
+    sat_idx = np.zeros((sat.shape[0], sat_thres))
+    for idx in np.arange(sat.shape[0]):
+        i = sat[idx]
+        s = slice(g_indptr[i], g_indptr[i + 1])
+        sat_idx[idx] = g_indices[s]
+
+    # get closest saturated of unsaturated
+    nearest_sat = np.zeros(unsat.shape) * np.nan
```
michalk8, line 316: I'd use `np.full_like(unsat, fill_value=np.nan, dtype=np.float64)` or `np.full(unsat.shape, fill_value=np.nan)`.

squidpy/gr/_sepal.py
```diff
@@ -0,0 +1,341 @@
...
+    for idx in np.arange(unsat.shape[0]):
+        i = unsat[idx]
+        s = slice(g_indptr[i], g_indptr[i + 1])
+        unsat_neigh = g_indices[s]
```
michalk8, line 320: I'd remove the slice and just use `g_indices[g_indptr[i]: g_indptr[i + 1]]` unless numba complains, same as above.

squidpy/gr/_sepal.py
```diff
@@ -0,0 +1,341 @@
...
+    elif max_n == 6:
+        fun = _laplacian_hex
+    else:
+        NotImplementedError("Laplacian for `max_nbrs`= {max_n} is not yet implemented.")
+    for i in ixs:
+        conc = vals[:, i].copy()
+        time_iter = _diffusion(
+            conc, fun, n_iter, sat, sat_idx, sat.shape[0], unsat, unsat_idx, conc.shape[0], dt=dt, thres=thres
+        )
+        if time_iter.shape[0] != 0:
```
michalk8, line 179: I'd make `_diffusion` return a float since you're just accessing the 1st elem of the entropy array (if present).
If the array is empty, just return `np.nan` from `_diffusion`. It will save you again some memory + `if` in this loop.

squidpy/gr/_sepal.py
```diff
@@ -0,0 +1,341 @@
...
+        - :attr:`anndata.AnnData.uns` ``['sepal_score']`` - sepal score.
+
+    Notes
+    -----
+    If some genes in :attr:`anndata.AnnData.uns` ``['sepal_score']`` are `NaN`,
+    consider re-running the function with increased `n_iter`.
+    """
+    _assert_connectivity_key(adata, connectivity_key)
+    _assert_spatial_basis(adata, key=spatial_key)
+
```
michalk8, line 95: Please add check for the maximum neighbors if it's in `(4, 6)` and raise ValueError if it isn't.
I know it's being checked later down the code, but not after some operations with non-trivial cost.

squidpy/gr/_sepal.py
```diff
@@ -0,0 +1,341 @@
...
+def _get_nhood_idx(
+    sat: np.ndarray, unsat: np.ndarray, g_indptr: np.ndarray, g_indices: np.ndarray, sat_thres: int
+) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
+    """Get saturated and unsaturated nhood indexes."""
+    # get saturated nhood indices
+    sat_idx = np.zeros((sat.shape[0], sat_thres))
+    for idx in np.arange(sat.shape[0]):
+        i = sat[idx]
+        s = slice(g_indptr[i], g_indptr[i + 1])
+        sat_idx[idx] = g_indices[s]
```
michalk8, line 313: I'd remove the slice and just use g_indices[g_indptr[i]: g_indptr[i + 1]] unless numba complains.

tests/graph/test_sepal.py
```diff
@@ -0,0 +1,54 @@
...
+
+UNS_KEY = "sepal_score"
+
+
+def test_sepal_seq_par(adata: AnnData):
+    """Check whether sepal results are the same for seq. and parallel computation."""
+    spatial_neighbors(adata, coord_type="visium")
+    adata.var["highly_variable"] = np.random.choice([True, False], size=adata.var_names.shape, p=[0.005, 0.995])
+
+    sepal(adata, max_nbrs=6)
```
michalk8, line 16: Could you please parametrize the test with max. neighs = [4, 6]?
Same for the below test
Also both tests are flaky because of the `np.random.choice` without seed (could lead to CI randomly breaking), please add seed.

squidpy/gr/_sepal.py
```diff
@@ -0,0 +1,341 @@
...
+        else:
+            genes = adata.var_names.values
+    genes = _assert_non_empty_sequence(genes, name="genes")
+
+    n_jobs = _get_n_cores(n_jobs)
+
+    vals = _get_obs_rep(adata[:, genes], use_raw=use_raw, layer=layer)
+    vals = _resolve_vals(vals)
+
+    g = adata.obsp[connectivity_key].copy()
```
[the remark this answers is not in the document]
giovp: well, don't need to densify. Now added check if not sparse, and if so sparsify

almaan:
Thanks a lot for this @giovp, I've looked through the docstings  - which you did a wonderful job of assembling - and can give the following suggestions (very minor):

for `sepal`:
``` 
   dt
       time step size in diffusion simulation.
   n_iter
        Maximum number of iterations for the diffusion simulation. If n_iter of iterations are reached the simulation will terminate even though convergence has not been achieved.


```
<hr>

for `_laplacian_rect` one-line description:
`Five point stencil approximation on rectilinear grid`

If you want to include a reference, you could use: https://en.wikipedia.org/wiki/Five-point_stencil

<hr>
for `_laplacian_hex`  : one -line description:

`Seven point stencil approximation on hexagonal grid`

for a reference you could use : Approximate Methods of Higher Analysis (Curtis D. Benster, L.V. Kantorovich, V.I. Krylov)  ISBN-13: 978-0486821603.

<hr>

For `_compute_idxs` and `_get_nhood_idx` you could perhaps change the one-liners to:

` _compute_idxs: """Get saturated and unsaturated nodes and neighborhood indices."""`
`_get_nhood_idx` : `"""Get saturated and unsaturated neighborhood indices."""`
<hr>
As for the remaining helper functions, did you want me to add parameter descriptions there (I see that none is added yet), or is the praxis to mainly add docstrings to the main functions? I'm happy to do it, but if there's no point in doing it I might skip it ;) 
`
<hr>
Regarding the question about choosing good Visium parameters:

> noticed that for visium a large part of high spatially variable genes are genes with very low expression (e.g. only few spots expressing) and tried to find a better heuristic. Did not really managed. If you have any suggestions much appreciated.

I did not quite experience this specifically when looking at the Visium-derived data, but what became quite apparent for me when developing the method was that the normalization had some impact on the performance. I used `log2(x+2)` and division with the max-value for the normalization, but judging from the line:

` vals = _get_obs_rep(adata[:, genes], use_raw=use_raw, layer=layer)`
 
I assume you are either using the active layer of normalized values (if log normalized, I would guess this is rather with pseudocount 1 rather than 2) or the raw values. This was something I addressed in the reviewer's comments, and my conclusion is that it has do to with how the expression gradients sort of getting inflated when you have a single spot with non-zero expression surrounded by spots with zero counts. Below is an excerpt from that answer (though illustrated for 1k arrays, but it conveys the same idea).

image

Having a slightly larger pseudocount quenches this inflation to some extent.

 An *alternative* strategy that works equally well is to apply some filtering to the genes being assessed, to only include those with a percentage of non-zero observations above a certain threshold. In the CLI application, this is what the flag `-mzp` allows you to do.

I think, rather than re-normalizing the data (costs a lot for large data and is inconvenient), this - filtration of genes - could be a good strategy. Implementation-wise, one could perhaps just assign these genes -inf values for the diffusion time or set it to the lowest value found in the data (even though I might prefer the former).

<hr>
Also this:

> so after some thinking, with this PR we'll drop the support to non grid-like graphs. However, with @MxMstrmn we've been thinking a bit about how to generalize the proposed diffusion process. We got some early results but not yet super sure about it, would be really nice to catch up and get your feedback on it. I'll let you know as soon as we have something to share!`

sounds super exciting, I've had a lot of stuff last months but things are starting to clear up now, and I'm super-down for a dicussion! Would love to hear what you guys have come up with!

Sorry for not getting back to you later, and again thanks for all the work on this!

squidpy/gr/_sepal.py
```diff
@@ -0,0 +1,341 @@
...
+
+    max_n = np.diff(g.indptr).max()
+    if max_n != max_nbrs:
+        logg.warning(f"Found node with # neighbors == {max_n}, bu expected `max_nbrs` == {max_nbrs}.")
```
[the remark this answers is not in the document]
giovp: actually not, re-added the value error and fixed the check (also related to your comment down)

squidpy/gr/_sepal.py
```diff
@@ -0,0 +1,341 @@
...
+
+    # get saturated/unsaturated nodes
+    sat, sat_idx, unsat, unsat_idx = _compute_idxs(g, spatial, max_nbrs, "l1")
```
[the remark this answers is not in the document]
giovp: yeah this is just for defining saturated/unsaturated neighbors in an efficient way, can remain fixed.

squidpy/gr/_sepal.py
```diff
@@ -0,0 +1,341 @@
...
+
+    score = []
+    if max_n == 4:
+        fun = _laplacian_rect
+    elif max_n == 6:
+        fun = _laplacian_hex
+    else:
+        NotImplementedError("Laplacian for `max_nbrs`= {max_n} is not yet implemented.")
+    for i in ixs:
+        conc = vals[:, i].copy()
```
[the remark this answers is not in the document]
giovp: it cannot be passed as sparse

squidpy/gr/_sepal.py
```diff
@@ -0,0 +1,341 @@
...
+        time_iter = _diffusion(
+            conc, fun, n_iter, sat, sat_idx, sat.shape[0], unsat, unsat_idx, conc.shape[0], dt=dt, thres=thres
```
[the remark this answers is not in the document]
giovp: you mean for clarity or efficiency?

squidpy/gr/_sepal.py
giovp: I see what you mean but if there is np.nan/inf in arr then we'd not be able to assign it to closest saturated neighbor so would not work I think

squidpy/gr/_sepal.py
giovp: > a float since you're just accessing the 1st elem of the entropy array (if present).

it's not the first element but the first array index. I then use the shape of the index to count the number of steps it required, and multiply it by dt to get the final "time"

squidpy/gr/_sepal.py
giovp: it's now added before any computation (after checking it graph is sparse)

tests/graph/test_sepal.py
giovp: mmh problem is that i would have to parameterize with two dataset (fixtures), can't use the same dataset for both hex and square. I tried a bit but didn't manage

giovp:
thanks a lot for the review @michalk8 ! (hopefully) addressed as requested all comments.

squidpy/gr/_sepal.py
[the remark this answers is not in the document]
giovp: I densify it at get_obs_rep now, and not copying it anymore.

squidpy/gr/_sepal.py
michalk8: Ok, I see. Let's keep it as it is.

squidpy/gr/_sepal.py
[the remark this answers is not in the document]
michalk8: For the former, mostly.

giovp:
thanks for all the comments @almaan , made changes in docstring as you suggested. Looks good to me!

> I think, rather than re-normalizing the data (costs a lot for large data and is inconvenient), this - filtration of genes - could be a good strategy. Implementation-wise, one could perhaps just assign these genes -inf values for the diffusion time or set it to the lowest value found in the data (even though I might prefer the former).

I agree. I think rather than in the main function, I will document this in the example (doing some filtering on features derived with `sc.pp.calculate_qc_metrics` like pct_dropout etc. Thanks for the pointer. Also agree that we should keep the data as is, and not re-process/normalize.

> sounds super exciting, I've had a lot of stuff last months but things are starting to clear up now, and I'm super-down for a dicussion! Would love to hear what you guys have come up with!

that's great! I'll ping you hopefully within a month , now have major deadline myself that need to prioritize.

Thanks again for the comments and the help!

squidpy/gr/_sepal.py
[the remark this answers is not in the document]
michalk8: Still, I think you should densify at `cont = ...`, I don't see why you can't pass sparse array to that function (it's not numba).
It's much more efficient because if densifying at `get_obs_rep`, you get a dense array of `n_cells, n_genes`, whereas in the function, you get at most `n_cells, n_jobs`.

squidpy/gr/_sepal.py
michalk8: Ok, I get it now. I'd still do the below since it's a bit cleaner I think:
```python
# _diffusion
tmp = np.nonzero(entropy_arr <= thresh)[0]
return tmp[0] if len(tmp) else np.nan  # or tmp.shape[0] if tmp.shape[0] else np.nan
# _score_helper
time_iter = _diffusion(...)
scores.append(time_iter * dt)
```

tests/graph/test_sepal.py
michalk8: Ok, I'd say we can leave as it is, since 1 test use 4, the other uses 6.

squidpy/gr/_sepal.py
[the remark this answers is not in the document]
michalk8: Ok, thanks!

squidpy/gr/_sepal.py
michalk8: Done.

squidpy/gr/_sepal.py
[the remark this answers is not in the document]
michalk8: @giovp it turns out you were right, problem was that `adata[:, genes].X` returned a custom type because `AnnData`
views broke the parallelization (e.g. https://github.com/theislab/anndata/blob/0ec97410702f61b29acef80b6377d1232699fa94/anndata/_core/views.py#L92), i.e.
this is ok `adata.X[:, ixs]: csr_matrix, but adata[:, genes].X: SparseCSRView` I've created a custom extraction function, should do the job and will create an issue on `anndata`.

squidpy/gr/_sepal.py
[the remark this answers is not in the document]
michalk8: Done.

michalk8 (approved):

PR #317, opened 2021-04-04
add moran from scanpy

giovp:
### Types of changes

### Description
- wraps morans_i calculation from scanpy (work with @ivirshup )
- add pvalue comp based on standard approx from permutations (previous behaviour) and simply based on permutations.

### How has this been tested?
high reproducibility from pysal implementation

`score`
image

`pvalues`
image

note, the pvalues returned from previous implementation corresponds to current `pval_z_sim` .

2 points to discuss:
- the current implementation parallelize over permutations, and it's slower w/ 4k genes and 100 perms (approx 2 min against ~40 seconds ). However, it's more scalable wrt `# features` as well as leveraging sparse implementation for memory efficiency. I understand it's not very nice, but we are also dropping a dependency, so still worth it?
- I would also like to add `gearys_c` https://github.com/theislab/scanpy/blob/master/scanpy/metrics/_gearys_c.py , which has exactly same interface as `morans_i`. Shall we therefore make a unifying function called something like `sq.gr.spatial_autocorr` ? Or shall we keep them separate , thus resulting in `sq.gr.morans_i, sq.gr.gearys_c` ?
Thoughts @ivirshup @michalk8 ?

p.s. happy easter!


michalk8:
> the current implementation parallelize over permutations, and it's slower w/ 4k genes and 100 perms (approx 2 min against ~40 seconds ).

Did you try benchmarking it by e.g. using `py-spy` where most of the time happens? ~2.5x slowdown seems like a lot (I see a lot of `prange`, which does not always make things faster). Also, what about the `transformation` argument from pysal's impl.?

> I would also like to add gearys_c https://github.com/theislab/scanpy/blob/master/scanpy/metrics/_gearys_c.py , which has exactly same interface as morans_i. Shall we therefore make a unifying function called something like sq.gr.spatial_autocorr ?

I'd go for the common function.

giovp:
> Did you try benchmarking it by e.g. using py-spy where most of the time happens? ~2.5x slowdown seems like a lot (I see a lot of prange, which does not always make things faster).

here's the profile (when run in notebook)
[profile.svg.txt](https://github.com/theislab/squidpy/files/6257556/profile.svg.txt)

<details>
<summary>Details</summary>

```python
import squidpy as sq
sq.__version__

sq.gr.moran(adata, n_perms=100, n_jobs=4)%
```

</details>

I also tried changing `prange` to `range` but didn't get performance improvement. as of now, it takes ~10s for computing the score for ~18k genes. That is much faster than the current implementation which takes ~2 minutes for 4k genes (cause it's not vectorized). However, it looks like the permutation it's still very fast when done per gene, as compared to the proposed strategy that computes score across all genes at each permutation. 
I guess an even better approach would be to include the permutation step in the numba calculation. However, this would require time that I'm not super keen in putting into this yet (would like to see more feedbacks on usage).

>  Also, what about the transformation argument from pysal's impl.?

that's not strictly necessary, it's only useful when analytic p-value calculation is used, in that case row-normaliztion is the preferred choice. However, we only provide perm-based p value calculation so not necessary.

> I'd go for the common function.

great, I'll work towards that

giovp:
ok so I've spent some more time looking into speedup/pvalue computation and decided to settle on the following:
- didn't manage to get speedup, both modyfying numba implementation in scanpy (to compute variables only once and permutes over them) or change parallelization scheme (parallelize over genes, over permutations, over permutations chunks). 
- looked a bit more into p-value computation and decided to provide the analytic p value with normality assumption. Here's some results for Moran:
image

and gearys' C
image

interestingly, the gearys' C pvalues are quite different between simulations and analytic. I'm fairly confident that computation is correct (blindly checked with pysal implementation). 
for Moran's I instead the pval_norm is more conservative than simulations, which imho is a good thing if anything (in general not big fun of pvalues for these types of statistics, and in general).

also for reference, this is the correlation between the two scores (very high)
image

I modified tests, they are passing locally but seems to be failing in CI, it has to do with the fact that I added scanpy@master to requirements, but maybe the env is not refreshed? do you have any pointer on this @michalk8 ? 
also, ready for review, thanks in advance for the help/feedback!

michalk8:
@giovp Based on the errors, using master scanpy broke things because there's no leiden installation
(previously we've used `scanpy[leiden]`

giovp:
> If 'leiden' is not really needed for installation, I'd just add it to tox's requirements

good point, done, it's not striclty required yet but we'll probably add multi-modal partition functionality at some point...

squidpy/gr/_ppatterns.py
```diff
@@ -196,75 +198,109 @@ def moran(
...
-        np.random.seed(seed + ix)
 
-    moran_list = []
-    for g in gen:
-        mi = _compute_moran(adata.obs_vector(g, layer=layer), weights, transformation, permutations)
-        moran_list.append(mi)
+    score_perms = np.empty((len(perms), vals.shape[0]))
+    if seed is not None:
+        ix += seed
+    rng = default_rng(ix)
```
michalk8, line 289: Wrong seed - should be `default_rng(seed)`.

squidpy/gr/_ppatterns.py
```diff
@@ -196,75 +198,109 @@ def moran(
             genes = adata.var_names.values
     genes = _assert_non_empty_sequence(genes, name="genes")
 
+    params: Dict[str, Any] = {"mode": mode, "transformation": transformation, "two_tailed": two_tailed}
+
+    if params["mode"] == "moran":
```
michalk8, line 203: I'd use the `ModeEnum` class we have and rewrite this as e.g.:
```python
if mode == SpatialAutocorr.MORAN:
  ...
elif mode == SpatialAutocorr.GEARY:
  ...
else:
   raise NotImplementedError(f"Mode `{mode}` is not yet implemented.")
```

squidpy/gr/_ppatterns.py
```diff
@@ -196,75 +198,109 @@ def moran(
...
+        normalize(g, norm="l1", axis=1, copy=False)
+
+    res = np.empty((vals.shape[0], 6), dtype=np.float32) * np.nan
+
+    score = params["func"](
+        g,
+        vals,
+    )
+
+    start = logg.info(f"Calculating for `{n_perms}` permutations using `{n_jobs}` core(s)")
```
michalk8, line 229: I'd add which statistic we're computing: `Calculating {mode}'s statistic for ...` or similar.

squidpy/gr/_ppatterns.py
```diff
@@ -471,3 +507,109 @@ def _find_min_max(spatial: np.ndarray) -> Tuple[float, float]:
...
+        p_z_sim = np.empty(z_sim.shape)
+
+        p_z_sim[z_sim > 0] = 1 - stats.norm.cdf(z_sim[z_sim > 0])
+        p_z_sim[z_sim <= 0] = stats.norm.cdf(z_sim[z_sim <= 0])
+
+        var_sim = np.var(sims, axis=0)
+    else:
+        p_sim = p_z_sim = var_sim = np.empty(score.shape[0]) * np.nan
+
+    p_norm, var_norm = _analytic_pval(score, weights, params)
```
michalk8, line 561: If a division by 0 can occur in this function, maybe consider wrapping the call in https://numpy.org/doc/stable/reference/generated/numpy.errstate.html
Not necesary though.

michalk8 (changes requested):

michalk8:
> good point, done, it's not striclty required yet but we'll probably add multi-modal partition functionality at some point...

Ok, sounds good. I'm going to fix the CI now, still the problem persists.

requirements.txt
```diff
@@ -5,7 +5,7 @@ ipywidgets>=7.5.1  # progress bar
 omnipath>=1.0.4
 pandas>=1.2.0
 rasterio>=1.1.8
-scanpy[leiden]>=1.7.0
+scanpy @ git+https://github.com/theislab/scanpy@master # scanpy[leiden]>=1.7.0
```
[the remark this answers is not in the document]
giovp: should be fine I think

giovp:
thanks a lot @michalk8 for review, addressed all comments and adopted all your suggestions.

re using the `ModeEnum`, had to resort to one type ignore here:
```python
mode: Literal[SpatialAutocorr.MORAN.s, SpatialAutocorr.GEARY.s] = SpatialAutocorr.MORAN.s,  # type: ignore
```

main reason is that (I think) explictly setting as str w/ `.s` is better because:
- without specifyin `.s` , the doc would not render as str but as a variable
- the if/else statment down below would fail if explicit `sq.gr.spatial_autocorr(..., mode="moran")` was passed

but of course mypy would throw error for `ModeEnum.s` , but not for `ModeEnum`. I think I'm missing something though cause this is not best practice (?). Curious to hear your point (also, I think this popped up somwehre else before but now I'm forgetting what we settled on).

Another point where I ignored type was here:

```python
p_sim = p_z_sim = var_sim = None  # type: ignore
```	

but this should be ok.

If test pass and you agree with changes, can be merged imho.

michalk8 (approved):
I've fixed the docs (they weren't rendered properly + some pet peeves with mypy, etc.).
LGTM, thanks a lot @giovp .

PR #320, opened 2021-04-26
enable visualisation of images with variable scale

hspitzer:
### Types of changes

### Description

if an ImageContainer is downscaled, currently interactive visualisation does not work as expected.
This PR fixes this by changing the Interactive to respect image scale stored in `img.data.attrs['scale']`.
In addition, a scale kwarg was added to `ImageContainer.add_img`.

### How has this been tested?

needs a test

### Closes

...

### DIFF
--- a/squidpy/im/_container.py
+++ b/squidpy/im/_container.py
@@ -80,6 +80,9 @@ class ImageContainer(FeatureMixin):
     Parameters
     ----------
     %(add_img.parameters)s
+    scale
+        Scalefactor of the image with respect to the spatial coords saved in
+        the accompanying :class:`ad.AnnData`.
 
     Raises
     ------
@@ -90,12 +93,13 @@ class ImageContainer(FeatureMixin):
         self,
         img: Optional[Input_t] = None,
         layer: str = "image",
+        scale: float = 1.0,
         **kwargs: Any,
     ):
         self._data: xr.Dataset = xr.Dataset()
         self._data.attrs[Key.img.coords] = _NULL_COORDS  # can't save None to NetCDF
         self._data.attrs[Key.img.padding] = _NULL_PADDING
-        self._data.attrs[Key.img.scale] = 1
+        self._data.attrs[Key.img.scale] = scale
         self._data.attrs[Key.img.mask_circle] = False
 
         chunks = kwargs.pop("chunks", None)
@@ -246,7 +250,7 @@ class ImageContainer(FeatureMixin):
 
         suffix = img.suffix.lower()
 
-        if suffix in (".jpg", ".jpeg"):
+        if suffix in (".jpg", ".jpeg", ".png"):
             return self._load_img(imread(str(img)))
 
         if img.is_dir():
--- a/squidpy/pl/_interactive/_model.py
+++ b/squidpy/pl/_interactive/_model.py
@@ -35,6 +35,8 @@ class ImageModel:
     def __post_init__(self) -> None:
         self.symbol = Symbol(self.symbol)
         self.coordinates = self.adata.obsm[self.spatial_key][:, ::-1]
+        self.library_id = Key.uns.library_id(self.adata, self.spatial_key, self.library_id)
+        self.spot_diameter = Key.uns.spot_diameter(self.adata, self.spatial_key, self.library_id)
 
         if self.container.data.attrs.get("coords", _NULL_COORDS) != _NULL_COORDS:
             c: CropCoords = self.container.data.attrs["coords"]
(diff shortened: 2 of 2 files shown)


michalk8:
I've added the test, have only 1 question: is the `scale` going to be used elsewhere as well? If not, I'd consider it just passing it to `.interactive()`.

hspitzer:
> I've added the test, have only 1 question: is the `scale` going to be used elsewhere as well? If not, I'd consider it just passing it to `.interactive()`.

The way I currently think about this, it should be used in some other places, eg with generate_spot_crops.  

Let me explain with some context:
I am currently using lower res png images in the image container (the „hires“ pngs from adata). This seems to be a common use-case as Original-scale res tiffs are not always acquired. To make this work, I set the correct scale on initialization. Visualization then works, but generating spot crops does not (but could easily be adapted I think). 

Thinking a bit further we might also need a way to combine old+new scale (after cropping). 

I think it’s worth the effort to support though, because that enables you to fully use image features with smaller res images.

giovp:
> This seems to be a common use-case as Original-scale res tiffs are not always acquired

just want to jump in on this, original-scale resolution image is always acquired, and passed to spaceranger as input. They can be in several formats (have seen tiffs, png, jpg so far).
People don't use them/share them becasue so far visium images have been used only for visualization purposes. 

Whether all ImgCont functionalities should also work with some type of scale factor I think it's an interesting enhancement, but not sure what's the priority/effort required.

hspitzer:
> > This seems to be a common use-case as Original-scale res tiffs are not always acquired
> 
> just want to jump in on this, original-scale resolution image is always acquired, and passed to spaceranger as input. They can be in several formats (have seen tiffs, png, jpg so far).
> People don't use them/share them becasue so far visium images have been used only for visualization purposes.

Thanks for the clarification, that makes a lot of sense! 

> Whether all ImgCont functionalities should also work with some type of scale factor I think it's an interesting enhancement, but not sure what's the priority/effort required.

Currently when cropping, a scaled ImageContainer might be returned. But atm it does not behave as expected on squidgy processing functions. So its probably well worth the effort to fix this. I'll update the code here.

hspitzer:
feature extraction from scaled images now works. 

Interactive visualisation does not yet work for extracted crops that have beed downscaled during crop extraction: 
I think the reason is that crop offsets are saved in original image space, nor current (downscaled) crop pixel space, and cannot just be scaled with the saved scale factor, because this refers to the scaling factor between the adata coordinates and the current crop pixels.  

I'm not quite sure how to fix this, as this would mean changing how CropCoords works. Also, interactive visualisation of a downscaled image crop is probably not a very common use-case so imo it would be fine to not address it.

Open points:
- [ ] update or overwrite existing CropCoords attribute with new one?

michalk8:
> I'm not quite sure how to fix this, as this would mean changing how CropCoords works. Also, interactive visualisation of a downscaled image crop is probably not a very common use-case so imo it would be fine to not address it.

@hspitzer I've fixed it, should work now. There was additional bug when shifting the coordinates (x/y was swapped, padding was not taken into account when crop was used). Not sure how this happened, but it seems to be broken on master as well.
Furthermore, I've disallowed different scales in `uncrop` (doesn't make sense) and made sure it propagates to the newly created container.

Some code I've played around:
```python
import squidpy as sq
import napari

img = sq.datasets.visium_fluo_image_crop()
adata = sq.datasets.visium_fluo_adata_crop()

c1 = img.crop_corner(0, 0, (3000, 2500), scale=0.5).crop_corner(-500, -500, (2000, 1000), scale=2)
c2 = img.crop_corner(3000, 2500, (3000, 2500), scale=0.5 * 2)
c1, c2

with napari.gui_qt():
    c1.interactive(adata)

with napari.gui_qt():
    sq.im.ImageContainer.uncrop([c1, c2]).interactive(adata)
```

hspitzer:
Thanks a lot @michalk8 , napari works great now! 

I've fixed some bugs in cropping when curing already cropped images.
Tests are added and running. The PR can be merged from my side. 

- added functions to correctly chain cropping of cropped images
- fixed generating crop spots from cropped images
- added tests for chaining cropping, scaling, and generate_spot_crops from scaled/cropped images.

I had one issue with pushing though: 
First, this ran:
```
Strip unnecessary `# noqa`s..............................................Failed
- hook id: yesqa
- exit code: 1
- files were modified by this hook

Rewriting squidpy/im/_utils.py
```
, effectively removing `# noqa: A003` from line 109 in `squidpy/im/_utils.py` and then, flake8 was complaining about:
```
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

squidpy/im/_utils.py:108:5: A003 class attribute "slice" is shadowing a python builtin
```
Why is the first check removing the "noqa"?

michalk8:
Thanks @hspitzer , looks great. Once the tests pass, let's merge.

>, effectively removing # noqa: A003 from line 109 in squidpy/im/_utils.py and then, flake8 was complaining about:
> Why is the first check removing the "noqa"?

The 1st check remove's `noqas` that have no effect, so they shouldn't be there. Based on the above message, it looks to me that the `noqa` was expected on line `108`, but it was found on line `109` (however, when I checked the code, it was correctly on line `109` before, so not sure why it was removed, possibly some version mismatch).

Issue #314, opened 2021-03-24
Provide new graph building methods


giovp:
### Type of the feature

- [ ] New function in `sqduipy.im`?
- [x] New function in `squidpy.gr`?
- [ ] New function in `squidpy.pl`?
- [ ] Change an existing functionality, such as default behavior?
- [ ] Other?

### Description
General improvements for neighbor graph construction:
- [ ] Delaunay (@MxMstrmn still up to take a stab ? )
- [ ] Faster nn search with umap (pot just call `sc.pp.neighbors` )
- [ ] square grid-like graph (e.g. dbitseq) ?

MxMstrmn:
In ```_build.py``` something like would work for Delaunay: 
```
from scipy.spatial import Delaunay


def _build_connectivity(
    coords: np.ndarray,
    set_diag: bool = False,
) -> csr_matrix:
    """Build delaunay connectivity matrix from spatial coordinates."""
    N = coords.shape[0]

    tri = Delaunay(coords)
    adj_indices = []
    for sim in tri.simplices:
        for i in range(3):
            row = sim[i]
            col = sim[(i + 1) % 3]
            adj_indices.append([row, col]) if not [row, col] in adj_indices else None
            col = sim[(i + 2) % 3]
            adj_indices.append([row, col]) if not [row, col] in adj_indices else None
    adj_indices = np.array(adj_indices)
    row_indices = adj_indices[:, 0]
    col_indices = adj_indices[:, 1]

    if set_diag:
        row_indices = np.concatenate((row_indices, np.arange(N)))
        col_indices = np.concatenate((col_indices, np.arange(N)))

    conns_m = csr_matrix((np.ones(len(row_indices)), (row_indices, col_indices)), shape=(N, N))

    return conns_m
```

Wanted to add this to a new branch but tests failed wenn pushing...

PR #322, opened 2021-04-27
Delaunay graph computation

MxMstrmn:
### Types of changes

### Description
This PR adds the commutation of the Dealunay Triangulation graph form the ``scipy.spatial`` implementation. The graph is stored a adjacency matrix, similar to ``build_connectivities`` in ``_build.py`` 

### How has this been tested?
- ToDo -

### Closes
Related to #314

### DIFF
--- a/squidpy/gr/_build.py
+++ b/squidpy/gr/_build.py
@@ -10,7 +10,7 @@ from numba import njit
 from scipy.sparse import spmatrix, csr_matrix, isspmatrix_csr, SparseEfficiencyWarning
 from scipy.spatial import Delaunay
 from sklearn.neighbors import NearestNeighbors
-from sklearn.metrics.pairwise import cosine_similarity
+from sklearn.metrics.pairwise import cosine_similarity, euclidean_distances
 import numpy as np
 
 from squidpy._docs import d, inject_docs
@@ -110,7 +110,7 @@ def spatial_neighbors(
             Dst = None
 
     elif coord_type == CoordType.GENERIC:
-        Adj, Dst = _build_connectivity(coords, n_neigh, radius, return_distance=True)
+        Adj, Dst = _build_connectivity(coords, n_neigh, radius, delaunay=delaunay, return_distance=True)
     else:
         raise NotImplementedError(coord_type)
 
@@ -168,16 +168,24 @@ def _build_connectivity(
         tri = Delaunay(coords)
         col_lst = []
         row_lst = []
-        for i in np.arange(N + 1):
+        dists_lst = []
+        for i in np.arange(N):
             idx = np.argwhere(i == tri.simplices)[:, 0]
             idx_col = np.unique(np.setdiff1d(tri.simplices[idx.squeeze(), ...], i)).tolist()
             col_lst.append(idx_col)
             row_lst.append(np.repeat(i, len(idx_col)))
+            dists_lst.append(
+                euclidean_distances(
+                    coords[idx_col, :],
+                    coords[np.newaxis, i, :],
+                )
+            )
 
         col_indices = np.array(list(chain(*col_lst)))
         row_indices = np.array(list(chain(*row_lst)))
+        dists = np.array(list(chain(*dists_lst))).squeeze()
 
-    else:
+    elif radius is None and delaunay is False:
         results = tree.kneighbors()
         dists, row_indices = (result.reshape(-1) for result in results)
         col_indices = np.repeat(np.arange(N), n_neigh or 6)
--- a/squidpy/gr/_build.py
+++ b/squidpy/gr/_build.py
@@ -53,6 +53,8 @@ def spatial_neighbors(
         Number of rings of neighbors for Visium data.
     n_neigh
         Number of neighborhoods to consider for non-Visium data.
+    delaunay
+        Whether to compute the graph from Delaunay triangulation.
     radius
         Radius of neighbors for non-Visium data.
     transform
--- a/squidpy/gr/_build.py
+++ b/squidpy/gr/_build.py
@@ -1,5 +1,6 @@
 """Functions for building gr from spatial coordinates."""
 from typing import Tuple, Union, Optional
+from itertools import chain
 import warnings
 
 from scanpy import logging as logg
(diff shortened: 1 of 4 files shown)


giovp:
Hi @MxMstrmn ,

thanks a lot for the contribution! I had a look a bit at the current implementation and I think there is a faster solution without iterating across idxs for each simplex, you can try it out with this code

<details>
<summary>Details</summary>

```python
import scanpy as sc
import squidpy as sq
import scipy
import numpy as np
import itertools

from scipy.spatial import Delaunay, delaunay_plot_2d

points = adata.obsm["spatial"]
tri = Delaunay(points)

%%timeit
sq.gr.delaunay(
    adata,
)
# current 8.16 s ± 351 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

adj_current = adata.obsp["spatial_connectivities"]
sc.pl.spatial(
    adata,
    neighbors_key="spatial_neighbors",
    img_key=None,
    edges_width=1,
    edges=True,
)

%%timeit

N = adata.shape[0]
col_lst = []
row_lst = []
for i in np.arange(N + 1):
    idx = np.argwhere(i == tri.simplices)[:, 0]
    idx_col = np.unique(np.setdiff1d(tri.simplices[idx.squeeze(), ...], i)).tolist()
    col_lst.append(idx_col)
    row_lst.append(np.repeat(i, len(idx_col)))

col_indices = list(itertools.chain(*col_lst))
row_indices = list(itertools.chain(*row_lst))

# new 272 ms ± 8.25 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

adj = scipy.sparse.csr_matrix(
    (np.ones(len(row_indices)), (row_indices, col_indices)), shape=(N, N)
)
adata.obsp["spatial_connectivities"] = adj
sc.pl.spatial(
    adata,
    neighbors_key="spatial_neighbors",
    img_key=None,
    edges_width=1,
    edges=True,
)

np.testing.assert_array_equal(
    adj.toarray(),
    adj_current.toarray(),
)
# they are same
```

</details>

I'm gonna go on and cleanup a bit but there are couple of points that I'd like to discuss:

### qhull params exposed for users?
in https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.Delaunay.html there is note a mentioning that:

> Unless you pass in the Qhull option “QJ”, Qhull does not guarantee that each input point appears as a vertex in the Delaunay triangulation. Omitted points are listed in the coplanar attribute.

this refers to the `qhull_options`. Have you experimented a bit with this? do you think it's relevant? I'm wondering if it should be exposed to users, but if so a more detail explanation would be needed (which scipy doesn't provide).

### calculate distances for neighbors
Do you think it would be useful to calculate distances for neighbors? right now, it only outputs a `connectivities` obj but maybe `distances` would also be useful? I'm thinking because this method would be most likely used in non-grid structures where the distance actually have a meaning.

thanks in advance for inputs/suggestions

MxMstrmn:
Hi @giovp,

Thanks for taking up the PR! I am much in favour of making the code faster, the speed up looks good to me but have not yet checked in detail. For the initial commit I wanted the code to be readable and tried to stay close to the original implementation for the KNN connectivities. If the results are the same and checked, that is great, thanks! 

## QHULL
I have not yet experimented with ``qhull_options`` and I am not sure when point do not show up in the simplexes, might be worth checking the ``coplanar`` attribute then? For the moment, I think this is not an issue but should definitely be kept in mind.

## Distances 
I agree, at least being able to provide the same functionality as for KNN makes sense. Especially as the distances might vary across simplexes.

giovp:
ok I think this is good to go. There is some work to be done in the way arguments are handled in order to generalize graph construction but can be done in another PR (will summarize in an issue).

@MxMstrmn I added the euclidean distances as well, indeed re QULL useful to keep in mind,although don't have use case now to try it out, also documentation in qhull package is a bit cryptic

squidpy/gr/_build.py
```diff
@@ -161,7 +166,28 @@ def _build_connectivity(
         row_indices = np.concatenate(results[1])
         lengths = [len(x) for x in results[1]]
         col_indices = np.repeat(np.arange(N), lengths)
-    else:
+    elif delaunay is True:
```
michalk8, line 169: `elif delaunay`

squidpy/gr/_build.py
```diff
@@ -161,7 +166,28 @@ def _build_connectivity(
...
+                    coords[idx_col, :],
+                    coords[np.newaxis, i, :],
+                )
+            )
+
+        col_indices = np.array(list(chain(*col_lst)))
+        row_indices = np.array(list(chain(*row_lst)))
+        dists = np.array(list(chain(*dists_lst))).squeeze()
+
+    elif radius is None and delaunay is False:
```
michalk8, line 190: `and not delaunay`

michalk8: can be removed

michalk8 (changes requested):

michalk8:
Few comments:
- the condition starting https://github.com/theislab/squidpy/blob/delauny/squidpy/gr/_build.py#L163 seems to be broken,
  should be e.g.
```python
if delaunay:
   ...
else:
  tree = NearestNeighbors(n_neighbors=n_neigh or 6, radius=radius or 1, metric="euclidean")
  tree.fit(coords)
  if radius is None:
    ...
```

giovp:
> the condition starting https://github.com/theislab/squidpy/blob/delauny/squidpy/gr/_build.py#L163 seems to be broken,
should be e.g.

done

> is this fast? https://github.com/theislab/squidpy/blob/delauny/squidpy/gr/_build.py#L176
generally, I'm a bit distrustful when using numpy + sets (or was this done to preserve the order + uniqueness?)

I think so yes, essentially I need to remove from the idx array the current `i` index. The whole graph building function on ~3k obs takes:
`701 ms ± 39.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`

> should delauney be also passed here? https://github.com/theislab/squidpy/blob/delauny/squidpy/gr/_build.py#L93

good point, although logic should be change (see disc in #325 )

> could you also include this so that docs are being built? #323 (files)

done

michalk8 (approved):

PR #324, opened 2021-04-28
Dask image integration

### LINKED ISSUE #296: dask and dask-image to improve ImageContainer
@GenevieveBuckley thanks a lot for the discussion today! Pinging @hspitzer, key points that I got (please add more if missing):
- dask-image could replace rasterio to read images? It's using pims under the hood, and it should be able to handle multi page tiffs https://github.com/soft-matter/pims/
- dask-image has good handling of edge cases in e.g. filters etc. should look into it for pre-processing that we currently do with scikit-image (related to #101, segmentation at crop boundaries)

### LINKED ISSUE #327: Add label layer in napari
### Type of the feature

- [x] Change an existing functionality, such as default behavior?
- [ ] Other?

### Description

We can add a layer in `napari` as `viewer.add_labels(...)` to display the result from e.g. watershed segmentation.
Would need to add some metadata to `xarray` so that we know it's a label image.

### LINKED ISSUE #328: Allow channel=None in sq.im.segment
### Type of the feature

- [x] Change an existing functionality, such as default behavior?
- [ ] Other?

### Description

...

currently working on external segmentation methods that run on multiple channels.
https://squidpy.readthedocs.io/en/latest/api/squidpy.im.segment.html#squidpy.im.segment

could we allow `channel=None` which then runs the segmentation on multiple channels instead just one?

e.g. stardist has an H&E model that runs on 3 channels which would be nice to add as custom model

### LINKED ISSUE #333: Segmenting a crop expects wrong dimensions
### Description

In #320, propagating absolute crop coords was fixed (i.e. if a crop had coords `(100, 100, 200, 200)` and another crop was made with coords [rel to previous crop] `(50, 50, 100, 100)`, it would be correctly updated to `(150, 150, 200, 200)`, as absolute coords w.r.t. to the original image).
However, in the tutorial below, a crop is directly passed, gets cropped again (with `size=None` it's the same image) and gets correctly updated with the absolute crop coords. This causes an issue in `uncrop`, because the now absolute positin is `(1000, 1000, 2000, 2000)` instead of the relative crop `(0, 0, 1000, 1000)`, causing it to expect an image of shape `(2000, 2000)` and not of `(1000, 1000)`.
In #324 , we no longer rely on our cropping, so it should be fine.
@hspitzer does it make sense to have relative crop coords w.r.t. to the image we're cropping from, rather then the original uncropped image? It would mean essentially to add an option to not add the old+new coords in https://github.com/theislab/squidpy/blob/master/squidpy/im/_utils.py#L163.
I think the way things are currently should be fine.

### Minimal reproducible example
https://github.com/theislab/squidpy_notebooks/blob/master/examples/image/compute_segment_fluo.py

### Version

master

### LINKED ISSUE #336: Allow non-interactive plotting of segmentation mask
### Type of the feature

- [ ] Other?

### Description

Currently, we only have `as_mask: bool = True` which simply does thresholding. I'd adapt this to have it similar behaviour as `labels` layer in napari, i.e. coloring the individual segmentation objects.
Related notebooks issue: https://github.com/theislab/squidpy_notebooks/issues/61

### LINKED ISSUE #337: Crop image on ImageContainer and data
Good day,

Thanks for developing this great tool. I want to use tangram to deconvolve each spot, but I have problems cropping using the same coordinates on the ImageContainer object and the data with the ST data. I do not understand how to subset the ST data to match the image. When I use `crop_coor` within the scanpy workflow to display the plot (for instance `sc.pl.spatial(adata, img_key="hires", color="clusters",  crop_coord=[15000, 25000, 20000, 10000], alpha=0, size=1.3)`) and try to use those on the ImageContainer, it does not seem to work. How can I subset the data and crop the image to match the same coordinates?

Thanks in advance!

michalk8:
### Types of changes

### Description

Add dask support.

### How has this been tested?

Many times.

### Closes

closes #296 
closes #327 
closes #328 
closes #333 
closes #336 
closes #337


michalk8:
@giovp First version of watershed using dask is done, see below the comparison (chunks/size=1000):
current (note the vertical line in the middle for both labels and the image - chunking artifacts for preproc./segmentation)
2021-04-29T19:24:39+02:00
map_blocks (note that there's a vertical line in the middle - chunking artifact for segmentation)
2021-04-29T19:07:12+02:00
map_overlap (note no such vertical line)
2021-04-29T19:03:29+02:00

Performance-wise, it took 10s, compared to 20s using 8 cores on my machine, will need to benchmark this more thouroughly though, but the runtime seems fine.
However, there are other design decisions to make:

- [ ] remove segmentation blob? (don't really want to port it to dask
- [ ] how to handle custom models? before, the crops were passed (possibly in parallel) to the function
  we can keep the old cropping approach, but it has drawbacks:
  1. boundaries - we could do `dask.array.map_overlap`, as it's done for `watershed` right now
  2. DL-models + thread/process parallelization won't play together nicely - if above is used, we could rely on user not to parallelize over chunks, TF/PT models should be fine w.r.t. mem/cores
  relate to #326 
- [ ]  currently, `ImageContainer.apply` handles lazy computation very clumsily, always calling compute after getting the result
  we could potentially lazy-chain some computations, possible ideas:
  1. LazyImageContainer - does everything lazily unless you call `.compute()` or e.g. napari needs the data
  2. expose `lazy: bool = ...` to `ImageContainer.apply` + high level methods like `squidpy.im.process`
  3. create `apply_lazy` which computes stuff lazily, whereas if a lazy array is supplied to `.apply`, it is computed
  currently, I'd go for 1 or 2 (think 1st is a bit too heavy approach)

giovp:
> remove segmentation blob? (don't really want to port it to dask

yes sounds great

>  how to handle custom models? before, the crops were passed (possibly in parallel) to the function
>  boundaries - we could do dask.array.map_overlap, as it's done for watershed right now
DL-models + thread/process parallelization won't play together nicely - if above is used, we could rely on user not to parallelize over chunks, TF/PT models should be fine w.r.t. mem/cores
relate to #326

I agree that we should rely on user to not parallelize, and yse agree with map overlap...

>  currently, ImageContainer.apply handles lazy computation very clumsily, always calling compute after getting the result
we could potentially lazy-chain some computations, possible ideas:
LazyImageContainer - does everything lazily unless you call .compute() or e.g. napari needs the data
expose lazy: bool = ... to ImageContainer.apply + high level methods like squidpy.im.process
create apply_lazy which computes stuff lazily, whereas if a lazy array is supplied to .apply, it is computed
currently, I'd go for 1 or 2 (think 1st is a bit too heavy approach)

I would go for 2 as well, this essentially mean exposing lazy to all `process` and `segment` right? and for napari not necessary cause is able to handle lazyness right?
@hspitzer thoughts on this?

michalk8:
@giovp @hspitzer I think I've finished the integration, would you mind testing it/giving feedback as well?

To summarize the changes:
- using dask/dask_image whereever possible to avoid artifacts when chunking
- add `channelwise: bool = True` to `ImageContainer.show`
- add `ImageContainer.__setitem__`, .e.g `img['foo'] = some_array`, just for convenience at the cost of some control
- add `labels` layer to napari (use 
- allow channels to be shown separately in napari (automatically detected, can be set as `img[<layer>].attrs['luminance'] = True/False`
- allow `channel_dim=None` in `add_img` (creates automatically iff name/dim mismatch happens, e.g. `channels_0`)
- add lazy loading on png/jpegs
- add `ImageContainer.rename` to rename a layer (could also extend this for channels)
- add tab completion for ImageContainer (try tab-completing `img['im`)
- remove `SegmentationBlob`

TODOs:
- [ ] deal with `pims` + very large images (thinks it's a zip bomb)
- [ ] test the new additions/changes + feedback
- [ ] deal with non Z-dim TODOs in the code
- [ ] update docstrings
- [ ] write tests

Here's some starting code I've used, but more thorough testing is needed (lazy=True, seems to work as well):
```python
import squidpy as sq
import numpy as np
import napari

img = sq.datasets.visium_fluo_image_crop()
adata = sq.datasets.visium_fluo_adata_crop()
sq.im.process(img, chunks=1000, method='smooth', sigma=[1, 1, 0])
img.show("image_smooth", channelwise=True, dpi=180, cmap='gray')

sq.im.segment(img, layer='image_smooth', chunks=1000)

with napari.gui_qt():
    img.interactive(adata)
```

squidpy/im/_container.py
```diff
@@ -172,14 +177,16 @@ def add_img(
         Parameters
         ----------
         img
-            In memory array or path to on-disk *TIFF*/*JPEG* image.
```
hspitzer, line 175: the general ImageContainer doctoring needs to be updated as well

squidpy/im/_process.py
```diff
@@ -74,22 +90,26 @@ def process(
     if channel_dim is None:
         channel_dim = img[layer].dims[-1]
     layer_new = Key.img.process(method, layer, layer_added=layer_added)
+    map_kwargs: Dict[str, Any] = {"lazy": lazy}
```
hspitzer, line 93: allow passing of `depth` to `map_kwargs`?

hspitzer (commented):
Great work @michalk8 
Its lovely to see all the dask image functionalities here — especially the possibility of using `map_overlap` to take care of crop boundaries. And the tab-completion is amazing!! How is this implemented?

Ok, for removing SegmentationModelBlob. It was not very good anyways. I’d rather have support for Stardist etc in the future than some old models that no one uses.

I’ve played around with the functions a bit and have a couple of comments:
* in `im.process`, there is no way to pass kwargs to `map_overlap`, because all kwargs are passed to the processing function. I could imagine a use-case though, when you have a custom processing function, and would like to use the overlap functionalities. Could you expose the `map_overlap` arguments somehow (or at least a couple of them)?
* for the `SMOOTH` processing, I did not quite understand how this is implemented. You are setting chunks to 0, before passing the gaussian filter to `img.apply`. Why? Could you explain what is happening there?
* the new segmentation is great as well. I observe less border artifacts! I noticed, that you can see the chunks structure from the labels, and was wondering if you have tested somehow that cells are not labelled with the same label? 
* is it possible to tell dask to use several processes when applying functions to chunks?
* did you test the lazy loading for big pngs/jpegs? I don’t have a large image like this atm to test this.

And just as a reminder (also for myself): we should update with docs with these changes. In particular changes are needed in 
* ImageContainer example
* processing, segmentation example with pointers to how borders between chunks are handled
* we could add an example for `img.show` which now has some nice functionalities

michalk8:
> Could you expose the map_overlap arguments somehow (or at least a couple of them)?

Good idea, will do that.

> You are setting chunks to 0, before passing the gaussian filter to img.apply. Why? Could you explain what is happening there?

That's because `dask-image`'s `gaussian_filter` already contains `map_overlap` inside. In `ImageContainer.apply`, if `chunks != None` controls whether we use dask or not.

> I noticed, that you can see the chunks structure from the labels, and was wondering if you have tested somehow that cells are not labelled with the same label?

Yes, there's this [test](https://github.com/theislab/squidpy/pull/324/files#diff-7c41484d346aac7a9605c1dfc81e3d1726aaf1cf190b5728b4e6557abe54fcd6R184) which assign label `1` to 1 position for each of 4 (or 16) blocks and then checks if the full image has only unique labels. In general, we the labels are saved as `u32` and the lowest `k` bits encode the block id, the rest is the original label. So if number of labels/blocks is reasonable, there shouldn't be any overflow, but we can always preeemptively bump the dtype to `u64`.

> is it possible to tell dask to use several processes when applying functions to chunks?

Yes, should be double with this: https://docs.dask.org/en/latest/scheduling.html#configuration
One problem I've found is that when using processing + large images, PIL's `MAX_IMAGE_PIXELS` is reset, resulting in an error.

> did you test the lazy loading for big pngs/jpegs? I don’t have a large image like this atm to test this.

I've tried on 30k x 30k jpeg and found a bug, thanks! (seems like `pims` loads the array when fetching metadata, I will either delay it as well or find another way).

hspitzer:
> > I've tried on 30k x 30k jpeg and found a bug, thanks! (seems like pims loads the array when fetching metadata, I will either delay it as well or find another way).
> 
> Actually, I wanted to talk to you about this @hspitzer - currently, the ugly code in `_io.py` stems from handling tiffs. In our case,
> the `visium_hne_image_crop.tiff` is saved as `(3, 7272, 7272, 1)`, so 3 frames, 1 channel whereas the `visium_hne_image_crop.tiff` is saved as `(1, 3527, 3527, 3)`, 1 frame, 3 channels. What about reading tiffs always as `Z, Y, X, C` (+ add an option to swap `C <-> Z` in the `ImageContainer`? This would be backwards incompatible, since we save the fluo image as it has 3 channels though.

Hm, this is difficult. I just checked an original fluorescence tiff and and original h&e tiff. The fluo tiff has 3 pages and 1 channel. The h&e tiff has 1 page and 3 channels. This makes sense if you consider what is saved in the image, the fluo image encodes a different kind of information in the pages (multiple markers) than the h&e image encodes in the channels (RGB). We just decided to handle both of these types of information in the same way by mapping it to the channels dimension -- and I think this is a good option and we should make sure that if you have these types of images they are read "correctly" by default. 

I'm not sure how 3D or z-stacks data would look like, as I haven't seen examples of this saved in a single tiff -- if its saved in several files then its no problem to handle this for us. I think the code you have right now tries to disambiguate very well if there are several z stored in the same tiff. It might be a bit ugly, but as I said above, I would not change the default behaviour for fluo tiffs with one z. Imo the different pages should still be read in different channels. 

I'm very happy to discuss though, if you have a different opinion!

hspitzer:
> > and I think this is a good option and we should make sure that if you have these types of images they are read "correctly" by default.
> 
> Ok, then I'd propose to add an option when reading images called e.g. `determine_dims: Literal['auto', 'default'] or : Literal['prefer_channels', 'prefer_z', 'default']` where `default` is to keep the dims as they are in the tiff (i.e. `3, y, x, 1` for fluo, `1, y, x, 3` for H&E), `auto/prefer_channels` is what's done now and `prefer_z` would be like `prefer_channels`, but in the Z-dim.
> If you agree, I'd go for the 2nd option (`prefer_channels, prefer_z, default`), since it's more flexible and not too much work.

This is a great proposal, I love it! This makes it very clear what is happening, and the user can also decide on their own what they would like to use. Lets go for option 2 then.

michalk8:
I've also removed `as_mask: bool` from `ImageContainer.show` and replaced it with `segmentation_layer: Optional[str]` and `segmentation_alpha: float = 0.75`, it looks e.g. as follows (alpha=1.0):
segmentation

michalk8:
#337 will also be handled as (still need to write tests):
```python3
adata = ...
img = sq.im.ImageContainer(...)
adata = img.crop(...).subset(adata)
```

giovp:
I swear I'll test it out tomorrow or friday, sorry for super late reply.

qwuick q: these artifacts are due to `map_overlap` ?
image

or, are they artefacts at all?

michalk8:
> or, are they artefacts at all?

They are not artifacts of `map_overlap`, but of watershed. Artifacts of using `map_blocks`, our previous custom chunking is seen in https://github.com/theislab/squidpy/pull/324#issuecomment-829467853

squidpy/im/_container.py
```diff
@@ -71,13 +74,13 @@
...
     """
-    Container for in memory :class:`numpy.ndarray`/:class:`xarray.DataArray` or on-disk *TIFF*/*JPEG* images.
+    Container for in memory :class:`numpy.ndarray`/:class:`xarray.DataArray` or on-disk *TIFF*/*JPEG*/*PNG* images.
 
     Wraps :class:`xarray.Dataset` to store several image layers with the same x and y dimensions in one object.
     Dimensions of stored images are ``(y, x, channels)``. The channel dimension may vary between image layers.
 
-    Allows for lazy and chunked reading via :mod:`rasterio` and :mod:`dask`, if the input is a *TIFF* image.
-    This class is given to all image processing functions, along with :class:`anndata.AnnData` instance, if necessary.
+    This class also llows for lazy loading and processing using :mod:`dask`, and is given to all image
```
hspitzer, line 82:
```suggestion
    This class also allows for lazy loading and processing using :mod:`dask`, and is given to all image
```

hspitzer (approved):
Looks really amazing! The subset function is super cool!
Good to merge for me.

I've played around with the `infer_dimensions` fn, and it seems to do the right thing in most cases, but I'd like to ask you to document its behaviour properly somewhere. I find the code a bit hard to follow with all this switching of indices, so It would be good do either add a lot of comments in it, or to have a doc somewhere that explains how dimensions are guessed. 
If its not there already, we should also add a line or two in the docstring of `add_img` to explain that dimensions are guessed (but maybe lets do that after #329, where I add explicit dims as well).

squidpy/im/_container.py
michalk8: Done, thanks.

michalk8:
> I find the code a bit hard to follow with all this switching of indices, so It would be good do either add a lot of comments in it, or to have a doc somewhere that explains how dimensions are guessed.

I've added a docstring for the internal method, hopefully it's clearer now. Will update the main docstring in #329 to accomodate for explicit passing of dims.

>  Good to merge for me.

Thanks for the review, merging this.

PR #329, opened 2021-05-03
ImageContainer z-stacks

### LINKED ISSUE #321: support for z-stacks
### Type of the feature

- [x] New function in `squidpy.pl`?
- [ ] Other?

### Description

We should support datasets with a z-dimension properly. 
Broadly, the following functions are needed:
- infrastructure:
  - [ ] imagecontainer adaption to z-stacks
  - [ ] adata adaption to z-stacks
- analysis methods:
   - [ ] graph statistics with 3D coords
- visualisation:
   - [ ]  of results with 3D coords
   - [ ] interactive visualisation; allowing to switch between different z-coords

Below a couple of questions that I have regarding the infrastructure and visualisation parts:
#### How to store z dimension in adata and in img?
- adata: add column to `adata.obsm['spatial']` such that it stores `x, y, z`
- img:
    - (a) add z coord to DataArray such that we store all images in the same DataArray.
        - makes processing single images quite complicated as soon as we have more than one z
    - (b) add z coord as attribute to each DataArray
        - know immediately which z this belongs to, but need to make sure to always track this attribute.

#### How to modify interactive viewer to deal with z
- (a) limit to 1 z coord on initialisation (both imagecontainer and adata)
- (b) prefix point layers with z coord to enable multiple point layers (could visualise layers that have different z coords on top of each other)
     - might be confusing, how can the user know which layers actually belong to each other?
- (c) have multiple z coords, but only allow matching layers to be displayed (img + genes)
    - requires tight coupling between layers that is not present atm

related: #309 

happy to discuss @giovp and @michalk8

hspitzer:
### Types of changes

### Description

Enables ImageContainer to deal with z-stacks by extending representation to `y,x,z,channels`
- feature extraction: 
   - summary+histogram: work with 3D
   - segmentation: can only be extracted from ImageContainer with `len(img.data.z) == 1`
- processing:
  - smoothing: works per z-dim, can apply to several z-dims

merge after #324 (Das image integration)

### TODO
- [x] add concatenation function for creating img with z-stacks
- [x] add `library_id` argument to processing functions
- [x] check processing functions
- [ ] texture feature extraction
- [x] how to deal with ImageContainer attributes? -> check disc below
- [ ] tests:
   - [x] fix existing tests (TODO: segmentation, features, processing, DONE: container, io)
   - [x] test for new `infer_dimensions` option (is z dim loaded correctly)?
   - [x] test for `show` on zstack image
   - z-stack tests:
	- [x] test for `concat`
	- [x] test for cropping different library_ids
	- [x] test for `crop_corner`  with scale and mask_circle
	- [x] test for `generate_spot_crops` with different library ids
	- [x] test for `apply` with different library_ids
	- [x] test for `apply` with overwriting existing layers 
	- [x] test for assigning library_id on loading image
	- [x] 	test `process` / `segment` with more than one library_id - make sure that non-processed library_ids are either identity or 0s.
	- [x] test `img._library_id_list`

### Closes

closes #321


hspitzer:
How should we deal with ImageContainer attributes when concatenating multiple images in one zstack?
E.g. we could have 2 images that have been cropped differently and should now be concatenated.
options:
1. not deal with it, i.e. require the user to update `adata.obsm['spatial']` accordingly before concatenating imgs, and remove existing attrs 
2. keep attributes for each z (library_id) separately, by using sth like `img.data.attrs = {library_id: ATTRS}`

Easiest would be 1. and since we won't provide registration capabilities, we'd expect the user to provide registered images anyways. 2. obviously is way nicer, but also a little bit more involved to implement.

michalk8:
> keep attributes for each z (library_id) separately, by using sth like img.data.attrs = {library_id: ATTRS}

If it's not too much work, I'd go for this, since it's the clean solution.

hspitzer:
> > keep attributes for each z (library_id) separately, by using sth like img.data.attrs = {library_id: ATTRS}
> 
> If it's not too much work, I'd go for this, since it's the clean solution.

I've decided to not go for this at the moment. It quickly becomes a nightmare to keep track of all kinds of different image resolutions + croppings. I still believe that this is the cleanest solution, but this needs to be implemented at a later date. 
For now, I'm just focussing on getting processing and feature extraction working with z stacks. 

@michalk8 as you can see lint failed on my last commit -- this is because I redefine variables in function bodies, but Mypy does not seem to realise and still assumes the original type of the variable. Do you know what I need to do to fix this? I could just give all variables new names, but this does not make the code very readable.

michalk8:
> @michalk8 as you can see lint failed on my last commit -- this is because I redefine variables in function bodies, but Mypy does not seem to realise and still assumes the original type of the variable. Do you know what I need to do to fix this? I could just give all variables new names, but this does not make the code very readable.

We can do 2, either:
1. redefine the variable as:
```python
library_id: Sequence[str] = self.data.coords["z"].values if library_id is None else [library_id]  # type: ignore[no-redef]
````
2. the above + add `allow_redefinition = True` to `.mypy.ini` to remove the type ignore (will need more changes since previous code also sometimes redefines var)

**Comment by hspitzer** [comment]
> > @michalk8 as you can see lint failed on my last commit -- this is because I redefine variables in function bodies, but Mypy does not seem to realise and still assumes the original type of the variable. Do you know what I need to do to fix this? I could just give all variables new names, but this does not make the code very readable.
> 
> We can do 2, either:
> 
> 1. redefine the variable as:
> 
> ```python
> library_id: Sequence[str] = self.data.coords["z"].values if library_id is None else [library_id]  # type: ignore[no-redef]
> ```
> 
>  1. the above + add `allow_redefinition = True` to `.mypy.ini` to remove the type ignore (will need more changes since previous code also sometimes redefines var)

Great, lets go for 1.: redefining,I think this is easiest. I just did not know the exact syntax to do it.

**Comment by hspitzer** [comment]
current status:
* added option to `infer_dimensions` to explicitely pass dimension names. Otherwise, doing sth. like 
  ```
  img = ImageContainer(np.zeros((10,10,3)))
  display(ImageContainer(img['image'].values))
  ```
  does not results in the expected shape of (10,10,1,3). 
  With the new option you can do `ImageContainer(img[‘image’].values, infer_dimensions=img[‘image’].dims)` which results in   the correct shapes. Another option would be to add another literal to `infer_dimensions`, but I like the flexibility of passing a   tuple of dimension names.  Related to this is the semi-broken functionality of `img[‘layer’] = some_array`, if `some_array` has   the form `y,x,z,channels`. I’m not sure how to fix this — using different defaults for add_img, depending on what function you   call seems like a bad idea. @michalk8 (`test_dask_processing` is broken because of this)
* added `library_id` to apply, to allow user to select an image slice. By default, all slices are used.
* added `library_id` to cropping fns, to allow cropping of one library_id only
* changed `as_array` to remove z-dim if there is only 1 dim. This makes it backwards compatible to the existing TF tutorials. I think this is ok to do, and makes using the output for downstream TF analysis easy.
* changed functionality of `add_img` to keep old values when overwriting a layer which has nans. This allows us to consecuteively apply processing functions to different `z` dims (`library_id`), and write results to the same layer. I feel that this is very reasonable behaviour when overwriting / adding information to a layer, but I’m happy to make this optional as well.
Here an example:
```
img = sq.im._container.ImageContainer(np.zeros((10,10,3,3)), infer_dimensions=('y','x','z','channels'))
img.apply(lambda arr: arr + 1, channel=None, copy=False, new_layer='res', library_id='library_id_0', layer='image')
img.apply(lambda arr: arr + 2, channel=None, copy=False, new_layer='res', library_id='library_id_1', layer='image')
img.apply(lambda arr: arr + 3, channel=None, copy=False, new_layer='res', library_id='library_id_2', layer='image')
img.show('res', vmin=0, vmax=10, channel=0)
```

Questions:
* `chunks` argument for dask fns needs to have 4 dims now instead of three. Should we make some sort of default behaviour if z-dim is 1, to allow passing 3 dims only? I’d vote for no, as this usage of chunks is super new anyways, so no backwards compatibility is lost here (I think)

minor:
* fixed bug in `apply`, where the condition for a new `channel_dim` name used a wrong shape for img to determine wheterh the channel dimension changed.

**Comment by michalk8** [comment]
>  does not results in the expected shape of (10,10,1,3).

It's because for 4D, the silent assumtion is it's either `z, y, x, c` or `c, y, x, z`. I can use the heuristic I'm using for 3D (to determine which dim is c/z) and apply it to 4D, but I agree explicitly passing dims is a good idea.
Also, if we were to save the images in the contianer as `z, y, x, c`, it would "work". Maybe we should discuss whether to store `z, y, x, c` or `y, x, z, c` (personal preference is the 1st [afaik, that's how it's going to be transposed nonetheless in `.interactive`).
I'd also rename `infer_dimensions -> dims` since you can now explicitly specify them.

> Related to this is the semi-broken functionality of img[‘layer’] = some_array, if some_array has the form y,x,z,channels. I’m not sure how to fix this 

The heuritic can fix it, in most cases imho (i.e. check what the 2 largest dims are and declare them spatial, rest are c/z) + maybe add method to swap between c/z dims if heuristic didn't work.
Alt would be to optionally specify dims in `img['foo', ('y', 'x', 'c', 'z')] = arr`, but it loses its nice properties.

> changed as_array to remove z-dim if there is only 1 dim. This makes it backwards compatible to the existing TF tutorials. I think this is ok to do, and makes using the output for downstream TF analysis easy.

Maybe this could be generalized to `squeeze: bool = True` to remove all (non-spatial) dims of 1, e.g. `10, 10, 1, 1 -> 10, 10`.

>  chunks argument for dask fns needs to have 4 dims now instead of three. Should we make some sort of default behaviour if z-dim is 1, to allow passing 3 dims only? I’d vote for no, as this usage of chunks is super new anyways, so no backwards compatibility is lost here (I think)

Agree on the bwd compat. Do you expect some functions to also use the Z-dim? I'd say we can keep it flexible (but also convenient) and allow passing both 3D/4D and if 3D chunks are passed, set the chunk for `z` always to 1.
But also passing only 4D is just fine I think.

> changed functionality of add_img to keep old values when overwriting a layer which has nans. This allows us to consecuteively apply processing functions to different z dims (library_id), and write results to the same layer. I feel that this is very reasonable behaviour when overwriting / adding information to a layer, but I’m happy to make this optional as well.

I think this behavior is related more to `apply` than to `add_img`. Not sure if this is any better, but we could make in apply
`func: Union[Callable, Mapping[str, Callable]]`. If just a function, all Z-stacks are passed. You example would be in this case:
```python
img = sq.im.ImageContainer(np.zeros((10,10,3,3)), infer_dimensions=('y','x','z','channels'))
img.apply({
  "library_id_0": lambda arr: arr + 1,
  "library_id_1": lambda arr: arr + 2,
  "library_id_1": lambda arr: arr + 3,  
}, channel=None, copy=False, new_layer='res', library_id='library_id_0', layer='image')
```
If any library id is missing, identity function is implied. The NaN approach worries me because NaN could be also used as a mask (or other purposes).

> fixed bug in apply, where the condition for a new channel_dim name used a wrong shape for img to determine wheterh the channel dimension changed.

Thanks!

michalk8:
I've added the better heuristics in #324 , think that issue should be closed asap and merged into this one.

hspitzer:
> > does not results in the expected shape of (10,10,1,3).
> 
> It's because for 4D, the silent assumtion is it's either `z, y, x, c` or `c, y, x, z`. I can use the heuristic I'm using for 3D (to determine which dim is c/z) and apply it to 4D, but I agree explicitly passing dims is a good idea.
> Also, if we were to save the images in the contianer as `z, y, x, c`, it would "work". Maybe we should discuss whether to store `z, y, x, c` or `y, x, z, c` (personal preference is the 1st [afaik, that's how it's going to be transposed nonetheless in `.interactive`).

Hm, switching dims would be an option. However, its a breaking change if you use `img['image'].values`. I'm not sure if this would maybe break too many things. Maybe @giovp has an opinion?

> I'd also rename `infer_dimensions -> dims` since you can now explicitly specify them.

Yes, good point! Will do.

> The heuritic can fix it, in most cases imho (i.e. check what the 2 largest dims are and declare them spatial, rest are c/z) + maybe add method to swap between c/z dims if heuristic didn't work.
> Alt would be to optionally specify dims in `img['foo', ('y', 'x', 'c', 'z')] = arr`, but it loses its nice properties.

I think a heuristic is fine, the use can always use `img.add_img` and explicitly pass `dims`.

> > changed as_array to remove z-dim if there is only 1 dim. This makes it backwards compatible to the existing TF tutorials. I think this is ok to do, and makes using the output for downstream TF analysis easy.
> 
> Maybe this could be generalized to `squeeze: bool = True` to remove all (non-spatial) dims of 1, e.g. `10, 10, 1, 1 -> 10, 10`. 

yes, that is probably a better idea. Again, I am not sure how the numpy arrays are used downstream, and if its a problem if they are either 2D, 3D, 4D, depending on how many channels, and z we have. Using `squeeze` feels like the cleanest way though. @giovp would this work for the DL example?

> > chunks argument for dask fns needs to have 4 dims now instead of three. Should we make some sort of default behaviour if z-dim is 1, to allow passing 3 dims only? I’d vote for no, as this usage of chunks is super new anyways, so no backwards compatibility is lost here (I think)
> 
> Agree on the bwd compat. Do you expect some functions to also use the Z-dim? I'd say we can keep it flexible (but also convenient) and allow passing both 3D/4D and if 3D chunks are passed, set the chunk for `z` always to 1.
> But also passing only 4D is just fine I think.

Atm I think all functions will use 1 z only. But in the future you might have full 3D volumes and might want to e.g. smooth in 3D. But I think setting the chunk for z to 1 if 3D chunks are passed is a good idea.

> > changed functionality of add_img to keep old values when overwriting a layer which has nans. This allows us to consecuteively apply processing functions to different z dims (library_id), and write results to the same layer. I feel that this is very reasonable behaviour when overwriting / adding information to a layer, but I’m happy to make this optional as well.
> 
> I think this behavior is related more to `apply` than to `add_img`. Not sure if this is any better, but we could make in apply
> `func: Union[Callable, Mapping[str, Callable]]`. If just a function, all Z-stacks are passed. You example would be in this case:
> 
> ```python
> img = sq.im.ImageContainer(np.zeros((10,10,3,3)), infer_dimensions=('y','x','z','channels'))
> img.apply({
>   "library_id_0": lambda arr: arr + 1,
>   "library_id_1": lambda arr: arr + 2,
>   "library_id_1": lambda arr: arr + 3,  
> }, channel=None, copy=False, new_layer='res', layer='image')
> ```
> 
> If any library id is missing, identity function is implied. The NaN approach worries me because NaN could be also used as a mask (or other purposes).

Awesome, this is much cleaner than what I came up with. And we can probably use this in the same way for `segment` as well (with a default of 0 for library_ids that are not segmented).

michalk8:
@hspitzer I've merged `dev` + fixed linting, hopefully all changes are correct (there were many conflics).

> Hm, switching dims would be an option. However, its a breaking change if you use img['image'].values. I'm not sure if this would maybe break too many things. Maybe @giovp has an opinion?

This reminds me of what's done in OpenGL, there you have vectors and can access them as `vec.xyzw, vec.zyxw, ...`.
We could define a desriptor to do the same, e.g. `img.yxcz['image']` or `img['image'].yxcz` as drop-in replacement for values.
In any case, by default, I'd go for `z, y, x, c` (many tests will need to be fixed, but I think it's worth it).

> Yes, good point! Will do.

Thanks!

> Using squeeze feels like the cleanest way though. @giovp would this work for the DL example?

Maybe we could also create a `get(layer: str)` method that always tries to squeeze z/channels (could be more convenient)

Pinging @giovp for opinion.

hspitzer:
> This reminds me of what's done in OpenGL, there you have vectors and can access them as `vec.xyzw, vec.zyxw, ...`.
> We could define a desriptor to do the same, e.g. `img.yxcz['image']` or `img['image'].yxcz` as drop-in replacement for values.
> In any case, by default, I'd go for `z, y, x, c` (many tests will need to be fixed, but I think it's worth it).

Thats a good point, I like this idea. This would also be very verbose, and make sure that the user knows which order they are getting. But first, I'm not sure if it should be high priority to do this now, I'd certainly prioritise getting the interactive plotting to work with zstacks (and there you can just transpose at the beginning, right?). And second, `img.yxcz` is not the fastest sequence of characters to type :-D

> 
> > Using squeeze feels like the cleanest way though. @giovp would this work for the DL example?
> 
> Maybe we could also create a `get(layer: str)` method that always tries to squeeze z/channels (could be more convenient)
> 
> Pinging @giovp for opinion.

I'm not sure if we need a `get` function really - returning numpy arrays directly from the crop iterators is nice, because this makes it compatible with `keras` models. Or could you chain the `get` function to the `generate_spot_crops`?

hspitzer:
@michalk8 @giovp This is ready for a first round of reviews.
From my last comment, I've now renamed `infer_dimensions` -> `dims`, fixed processing and segmentation.

In general, I've added an optional `library_id` argument to cropping, processing, and segmentation functions to control which z dims are being processed. Here a short set of examples of how the high level api now looks like for a zstack (there are no changes if you only have 1 z dim):
```python

import squidpy as sq
import numpy as np

# build example images
arrs = [np.linspace(0,1,10*10*3).reshape(10,10,3), np.zeros((10,10,3))+0.5, np.zeros((10,10,3))]
arrs[1][4:6,4:6] = 0.8
arrs[2][2:8,2:8,0] = 0.5
arrs[2][2:8,2:8,1] = 0.1
arrs[2][2:8,2:8,2] = 0.9
imgs = [sq.im.ImageContainer(arr) for arr in arrs]
img = sq.im.ImageContainer.concat(imgs, library_ids=['1','2','3'])

# img.show shows all z dims
img.show(vmin=0, vmax=1, channelwise=True)

# specify library_id to cropping fns to crop only certain z dims
img.crop_corner(0,0, (5,5), library_id=['2']).show(channel=0)
img.crop_corner(0,0, (5,5), library_id=['1', '3']).show(channel=0)

# use apply with {'library_id': func} to apply function to a specific library_id
img.apply({'2': lambda x: x+2}, layer='image', new_layer='test', copy=False, channel=0)
img.show('test', channelwise=True, vmin=0, vmax=3)

# specify library_id to processing fns to process only certain z dims
sq.im.process(img, method='smooth', layer='image', chunks=5, library_id='2', copy=False)
img.show('image_smooth', channel=0)

# specify library_id to segmentation to segment only certain z dims
sq.im.segment(img=img, layer='image', channel=0, library_id=['2','3'], chunks=None, copy=False)
img.show("segmented_watershed")
```
Let me know what you think!

open TODOs are:
- [ ] fix existing tests (@michalk8 , could you have a look?)
- [ ] add tests for processing / segmenting zstacks
- [ ] squeeze returned numpy arrays in crop generators? (see comment above)
- [ ] reorder internal dims to zyxc? (I'd prefer to do this in separate PR, to be able to close this soon)
- [ ] change napari to work with z dims (again, I'd be very thankful if @michalk8 could help, maybe in a separate PR as well)

I also noticed that in `im._io.py`, we mention `channels` explicitly as dim name. This can sometimes be changes to sth else after you apply preprocessing functions etc. Are we sure that this is working as expected when assuming that there is a `channels` dim? E.g. for `_lazy_load_image`?

michalk8:
>  From my last comment, I've now renamed infer_dimensions -> dims, fixed processing and segmentation.

Thanks! I will take care of the tests+napari.
Re `zyxc` reshape I agree, let's do it (if we do it) in a separate PR).
Re squeezing: I'd say adding this as an option won't hurt.

> Edit: should the base for this PR be master or dev? I saw that the dask_image stuff has been merged to dev, right?

Definitely `dev`.

giovp:
> @michalk8 @giovp This is ready for a first round of reviews.
From my last comment, I've now renamed infer_dimensions -> dims, fixed processing and segmentation.

hey guys, sorry for being totally absent on this discussion. I don't think I will be able to test it before wed and that's probably too late? Anyway, I trust your decision fully.

few things that come to mind (sorry if they sound too naive):
- when changing across zstacks, does it also display the correct annotations (gexp/clusters) in the anndata? e.g. I have 3 datasets in one anndata (with one `batch_key`), with 3 images concated in zstack for the image container. Ig I chnage zstack, is the anndata being subsetted for visualization?
- the `library_id` has a specified name in adata (where it store metadata such a scaling etc.) is this taken into account?

hspitzer:
> * when changing across zstacks, does it also display the correct annotations (gexp/clusters) in the anndata? e.g. I have 3 datasets in one anndata (with one `batch_key`), with 3 images concated in zstack for the image container. Ig I chnage zstack, is the anndata being subsetted for visualization?
> * the `library_id` has a specified name in adata (where it store metadata such a scaling etc.) is this taken into account?

These are good points. This has not been implemented yet, but @michalk8 will now start. I should probably clarify how I designed this and what assumptions I make about the adata:
1. we assume that all images have the same scale and cropping parameters (i.e. only one `scale` and `CropCoords` attribute per image). This can be changed in the future, but is ok for now; as this will always be the case when the images are registered.
2. each z dimension has a specific name `library_id`. (`img['image'].z`). By default this is a dummy name.
3. for the concatenation (creation of zstack from individual images) to work, you need to specify a valid list of `library_ids` (all unique strings).
4. when generating spot crops, the `adata` is assumed to either contain only 1 `library_id`, or the same `library_id`s as present in `img['image'].z. The spot diameter and scale is taken from the `adata.uns['spatial'][library_id]` entry.
5. napari should do the same - i.e. subset the adata to the currently displayed z dim, and read the spot diameter from the appropriate uns entry.

I have a couple of consecutive mouse brain sections that I can put together in an adata, maybe that will help @michalk8 for testing napari? I will also create a tutorial for this to make this a bit more clear.

michalk8:
> I have a couple of consecutive mouse brain sections that I can put together in an adata, maybe that will help @michalk8 for testing napari? I will also create a tutorial for this to make this a bit more clear.

That would be great, thanks.

hspitzer:
> > I have a couple of consecutive mouse brain sections that I can put together in an adata, maybe that will help @michalk8 for testing napari? I will also create a tutorial for this to make this a bit more clear.
> 
> That would be great, thanks.

Here an example of 2 saggital mouse brain visium slices:

```python
import scanpy as sc
import squidpy as sq
import anndata as ad

library_ids = ['V1_Mouse_Brain_Sagittal_Posterior', 'V1_Mouse_Brain_Sagittal_Posterior_Section_2']

adatas = []
imgs = []
use_hires_tiff = False
for library_id in library_ids:
    adatas.append(sc.datasets.visium_sge(library_id, include_hires_tiff=use_hires_tiff))
    adatas[-1].var_names_make_unique()
    if use_hires_tiff:
        imgs.append(sq.im.ImageContainer(adatas[-1].uns['spatial'][library_id]['metadata']['source_image_path']))
    else:
        # as we are using a scaled image, we need to specify a scalefactor to allow correct mapping to adata.obsm['spatial']
        imgs.append(sq.im.ImageContainer(
            adatas[-1].uns['spatial'][library_id]['images']['hires'],
            scale=adatas[-1].uns['spatial'][library_id]['scalefactors']['tissue_hires_scalef']))

adata = ad.concat(adatas, uns_merge='only', label='library_id', keys=library_ids, index_unique='-')
img = sq.im.ImageContainer.concat(imgs, library_ids=library_ids)
```

michalk8:
@hspitzer napari should be mostly fixed (still need to take a look at `Labels` layer), could you try it out e.g. as:
```python
...  # the above code
import napari

with napari.gui_qt():
    img.interactive(adata, library_key='library_id', library_id=None)  # only subset of ids can be used.
```

Currently, there should be a slider and the spots should be correctly displayed in each z-stack (currently, it's a bit hacky: when visualizing some spots belonging to 1 stack but not other, all of them are plotted, but the ones that are in the other stack have `size=0`, couldn't find a better solution). Currently, showing multiple Z stacks will not show all the spots (no Z-dim in spatial), but I think it's fine.
Exporting the cropped shapes with `Shift-E` will mark spots only on the selected Z-stack (there's a small visual bug when the selecting the other z-stack - the cluster name `True` is still being shown, can't do much about it).

Would like to know your feedback w.r.t. whether we should add another panel for library id on the right or keep the slider below (currently, the slider doesn't show you which id is being shown, but I can add this info somewhere).

I've also added `img.library_id = ...` setter so that it can be easily set.

hspitzer:
Great, I just tried it out and it works really great! I had to update napari to `0.4.8` though, and am now getting some Future warnings.
The switching of the slices with the slider works nicely, although a bit slow. How are the images internally loaded? Are all slices loaded in napari and its the displaying that takes such a long time? Imo its fine and expected that this take a bit of time (I was trying with the whole brain slices).

> Currently, showing multiple Z stacks will not show all the spots (no Z-dim in spatial), but I think it's fine.

I'm not sure what you mean with showing multiple Z-stacks?

> Exporting the cropped shapes with `Shift-E` will mark spots only on the selected Z-stack (there's a small visual bug when the selecting the other z-stack - the cluster name `True` is still being shown, can't do much about it).

Ok, that's fine!

> Would like to know your feedback w.r.t. whether we should add another panel for library id on the right or keep the slider below (currently, the slider doesn't show you which id is being shown, but I can add this info somewhere).

hm that's hard to say without actual use-cases. If there are only very few z dims (like in this example that I posted above), a panel would probably be good. On the other hand, if you have a slider its more intuitive, especially for many z dims. I like the slider and would keep it for now. But if possible, I'd add the current library_id somewhere (maybe below or to the side of the slider?). 

> I've also added `img.library_id = ...` setter so that it can be easily set and added `library_id` in the feature mixin (from `sq.im.calculate_image_features`, it doesn't do anything, since the passed crop should have only 1 z, but when calling it from `ImageContainer` directly, it should work).

Awesome, thanks! I had not considered calling features directly from the ImageContainer. 

> Another question: do we want to compute these features lazily? Currently, each spot crop calls `.compute(layer)`. I'd say it's fine for now as it is.

I'm not sure if lazy computation would help us here. We want to save the result in `adata`, which currently does not support lazy data -- or would this work? 
I think its not necessary atm.

hspitzer:
> Also, could you please look at the 2 `TODOs` in `add_img`? In short, default library ids are broken iff some/all of them mismatch (they arrays are silentfly filled with NaNs + z dim is expanded/reduced).

Ok I see the issue. I think we have to do some checks in `add_img` and for the `dims` argument.
x,y are "fixed" dimensions, meaning that all layers need to have the same shape.
We have to decide whether to make z "fixed" as well or to keep it flexible. Meaning, will we allow:
```
img = ImageContainer(np.zeros((10,10,1,1)), dims=('x','y','z','c'), library_id='z1'))
img.add_img(np.zeros((10,10,1,1)), dims=('x','y','z','c'), library_id='z2'))
```
?
I'm not sure if this level of flexibility is needed at the moment, and it is likely to cause problems in other processing functions.
I think for now, we want to keep z "fixed". This means:
- when initialising an empty ImageContainer, `library_id`s are set
- adding new images to this container will use the existing library_ids and raise an error if the number of z dims is not the same (?).
- we could allow passing a permuted list of library_ids in `img.add_img`. 

what do you think?

I also realised that we need to check that the `dims` argument contains `x`, `y`, `z` and a flexible name for the channel dimension. 

Also, the channel name specified in `dims` seems to be ignored? for `img.add_img(lab, dims=['x', 'y', 'z', 'cseg'], layer='foo')` I still get `('y', 'x', 'z', 'channels_0')` for `img['foo'].dims`
,z are "fixed" dimensions, meaning that all layers need to have the same x,y,z. Thus, on adding an img to an empty container, this sets the `library_id`s for this container. Then,

michalk8:
> The switching of the slices with the slider works nicely, although a bit slow. How are the images internally loaded? Are all slices loaded in napari and its the displaying that takes such a long time? Imo its fine and expected that this take a bit of time (I was trying with the whole brain slices).

I think this is the implementation issue right now - I just iterate over all `Points` layers and set the correct sizes (0 for not the current slice). Not sure how best to solve this.

> I like the slider and would keep it for now. But if possible, I'd add the current library_id somewhere (maybe below or to the side of the slider?).

Ok, I will add some info as which slice is being displayed.

> I'm not sure if lazy computation would help us here. We want to save the result in adata, which currently does not support lazy data -- or would this work?
I think its not necessary atm.

I agree, let's not do this now.

>  We have to decide whether to make z "fixed" as well or to keep it flexible. Meaning, will we allow:

In the future, I think non-fixed Z-dim would be better, but I also think this would create large issues now and it's not that necessary.

> when initialising an empty ImageContainer, library_ids are set
> adding new images to this container will use the existing library_ids and raise an error if the number of z dims is not the same (?).
> we could allow passing a permuted list of library_ids in img.add_img.

I like this. I think the only thing that needs change is [here](https://github.com/theislab/squidpy/blob/img_zstacks/squidpy/im/_container.py#L335) by not passing `default_library_id` (or only passing it when adding 1st image).

> Also, the channel name specified in dims seems to be ignored? for img.add_img(lab, dims=['x', 'y', 'z', 'cseg'], layer='foo') I still get ('y', 'x', 'z', 'channels_0')

Yes, I haven't fixed this yet - when passing `dims`, it should not rename channels, [this](https://github.com/theislab/squidpy/blob/img_zstacks/squidpy/im/_container.py#L329) should be removed.

michalk8:
Ready for review @hspitzer .

squidpy/im/_container.py
```diff
@@ -619,25 +698,58 @@ def generate_spot_crops(
...
+
+        if library_id is None:
+            try:
+                library_id = Key.uns.library_id(adata, spatial_key=spatial_key, library_id=None)
+                obs_library_ids = [library_id] * adata.n_obs
+            except ValueError as e:
+                if "Unable to determine which library id to use" in str(e):
+                    raise ValueError(
+                        str(e)
+                        + " Or specify a key in `adata.obs` containing a mapping from observations to library ids."
```
hspitzer, line 724: I tried to make the error message a bit clearer here, as the preferred way of specifying library_id should be an entry in adata.obs (for z-stacks anyways). Not sure if this is the best way of concatenating error messages...

hspitzer (commented):
This is fantastic, @michalk8 !! I tested the interactive visualisation, and its super fast now! It also works flawlessly with more than one image layer. I love it!
I also tested adding image layers with different dims + channels and the behaviour is as expected. The error messages are very clear now I think.
I have nothing to add or complain about, good to merge from my side.

giovp:
just tried with napari and I got this

```python
viewer = img.interactive(adata)
```

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-5-58c2d85826f0> in <module>
----> 1 viewer = img.interactive(adata)

~/Projects/squidpy/squidpy/im/_container.py in interactive(self, adata, spatial_key, library_key, library_id, cmap, palette, blending, symbol, key_added)
   1046         from squidpy.pl import Interactive  # type: ignore[attr-defined]
   1047 
-> 1048         return Interactive(  # type: ignore[no-any-return]
   1049             img=self,
   1050             adata=adata,

~/Projects/squidpy/squidpy/pl/_interactive/interactive.py in __init__(self, img, adata, **kwargs)
     38             raise ImportError(f"Unable to import the interactive viewer. Reason `{_error}`.")
     39 
---> 40         self._controller = ImageController(adata, img, **kwargs)
     41 
     42     def show(self, restore: bool = False) -> "Interactive":

~/Projects/squidpy/squidpy/pl/_interactive/_controller.py in __init__(self, adata, img, **kwargs)
     53         self._view = ImageView(model=self.model, controller=self)
     54 
---> 55         self.view._init_UI()
     56 
     57     def add_image(self, layer: str) -> bool:

~/Projects/squidpy/squidpy/pl/_interactive/_view.py in _init_UI(self)
    128         self.viewer.window.add_dock_widget(self._colorbar, area="left", name="percentile")
    129         self.viewer.window.add_dock_widget(widgets, area="right", name="genes")
--> 130         self.viewer.layers.selection.events.changed.connect(self._move_layer_to_front)
    131         self.viewer.layers.selection.events.changed.connect(self._adjust_colorbar)
    132         self.viewer.dims.events.current_step.connect(update_library)

AttributeError: 'LayerList' object has no attribute 'selection'


```

hspitzer:
> just tried with napari and I got this

@giovp I think you need to update Napari, see above, I had to update napari to 0.4.8 .

michalk8:
> just tried with napari and I got this

@giovp if there's nothing from your side, I think I'd merge this.

squidpy/im/_container.py
michalk8: Thanks, looks great!

giovp:
> @giovp I think you need to update Napari, see above, I had to update napari to 0.4.8 .

my bad! it works perfectly and it's so fast!

> @giovp if there's nothing from your side, I think I'd merge this.

LGTM !

Issue #325, opened 2021-04-28
Refactors and extend graph building


giovp:
### Type of the feature

- [ ] New function in `sqduipy.im`?
- [x] New function in `squidpy.gr`?
- [ ] New function in `squidpy.pl`?
- [ ] Change an existing functionality, such as default behavior?
- [ ] Other?

### Description

It would be nice to refactor the way the spatial graph is computed in order to generalize to other settings (e.g. grids). Follows up from #314 .

The way it's structured now is the following:
- if `coord_type = CoordType.VISIUM`, it builds a graph from hexagonal grid
- if `coord_type = CoordType.GENERIC`, it build a graph either by using the radius, a knn search or (from #322 ) a Delaunay triangulation.

It would be good to change behaviour with the following:
- if `coord_type = CoordType.GRID` (replace `VISIUM`), then builds a regular grid with either 4 (old ST data and dbit-seq) or 6 (visium) neighbors. Potentially allow for 8 (although don't have use case).
- `coord_type = CoordType.GENERIC`, then use an additional `mode` argument (?) to allow either knn, radius or the new delaunay. I'm not sure how this could be best set up but I guess re arguments exposed we can think about it down the line.

Additional points:
- Distances should be consistently passed imho. This is not done currently for visium. Although maybe redundant/not needed, I think it would still be useful and would improve consistency.
- The argument `neigh_correct` is currently in used but since it's not exposed, maybe it can be removed?
- Is there any graph building approach that we havent' considered and that it would be useful for users? Currently we have knn-search, regular grid, radius, delaunay. Pinging @MxMstrmn @hspitzer @AnnaChristina for ideas...

As discussed on MM @Koncopd , would you mind taking a look at this? Thank you!

Koncopd:
`neigh_correct` is an argument from the internal function to build visium graphs, it should be present.

Koncopd:
I don't know, maybe we should split the function, i am not sure how to pack all these changes to the arguments of a single function. Ofc it is possible but will be very messy.

giovp:
I would not split the function because then it comes more messy cause a lot of arguents are redundant I think.

So the way I see it is:
- if `coord_type = CoordType.GRID` , then 2 additional arguments will be used:  
  - `n_rings` which selects then number of neighbors ring to compute
  - `neigh_grid` which selects whether ist's a swuare grid (4) or hexagonal (6) etc.
- if `coord_type = CoordType.GENERIC` , then 2 additional arguments are used
[...]

PR #340, opened 2021-05-14
extend graph build

Koncopd:
https://github.com/theislab/squidpy/issues/325

Replaced `CoordType.VISIUM` with `CoordType.GRID`.
Added `neigh_grid` for grid coordinates. It is `int` now, i think it is the most flexible variant.
Distances are passed for `n_rings=1` also, though it is the same matrix as in connectivities.


giovp:
@Koncopd I don't think the `neigh_grid=4` is working:
image

I would expect this:
image

reprex

```python
import numpy as np
from anndata import AnnData

rng = np.random.default_rng(42)
coord = rng.integers(
    0,
    12,
    size=(200, 2),
)
coord = np.unique(coord, axis=0)
counts = rng.integers(0, 10, size=(coord.shape[0], 10))
adata = AnnData(counts)
adata.obsm["spatial"] = coord

sq.gr.spatial_neighbors(adata, neigh_grid=4)
# or
sq.gr.spatial_neighbors(adata, radius=1.0)
```

giovp:
hey @Koncopd ,

I've experimenting a bit more and still no success

```python
import numpy as np
from anndata import AnnData

rng = np.random.default_rng(42)
coord = rng.integers(0, 12, size=(200, 2))
coord = np.unique(coord, axis=0)
counts = rng.integers(0, 10, size=(coord.shape[0], 10))
adata = AnnData(counts)
adata.obsm["spatial"] = coord

n_rings, n_neigh, sum_dist = (2, 4, 0)
sq.gr.spatial_neighbors(adata, coord_type="grid", n_rings=n_rings, neigh_grid=n_neigh)
np.diff(adata.obsp["spatial_connectivities"].indptr).max() 
# this one gives 11, it should give 8 (it gives eight with radius = 1)

sc.pl.spatial(
    adata, spot_size=0.2, edges=True, edges_width=2, neighbors_key="spatial_neighbors"
)
```
image


beside the fact that `np.diff` gives the wrong result, see for instance the bottom right point that is not connected to anything (although it should)...
I wonder if this si a bug from nearestneighbor from scikit learn? Not sure
Maybe I'll try to have a look again tomorrow....

Koncopd:
And why should the bottom one be connected to anything actually?

giovp:
> Also radius=1. and n_rings=1 are the same (with max diff of 4). I don't understand why you compare n_rings=2 with radius=1.

with `n_rings=2` it should be 8 max diff, not 11 as it is now. I don't understand where that 11 come from

Koncopd:
Taking `radius=2.` is not the same as `n_rings=2` For example, the bottom one becomes connected with `radius=2.` and stays disconnected with `n_rings=2.` because it how the algorithm works now. We walk through neighbors, neighbors of neighbors and so on. If there are no neighbors it stays isolated. This is rather a question of how we build `n_rings`, i am pretty sure we discussed this.

giovp:
thanks @Koncopd !
weird bug in napari test but ok. Will merge this!

giovp (approved):

PR #351, opened 2021-05-27
Fix napari for multiple library_ids

hspitzer:
### Types of changes

### Description
- fix missing library id when rescaling image crops
- fix wrong spot_diameter in napari. There was a bug in the code that lead to napari only working with 2 library_ids. 

### How has this been tested?

...

### Closes

...

### DIFF
--- a/squidpy/im/_container.py
+++ b/squidpy/im/_container.py
@@ -541,7 +541,8 @@ class ImageContainer(FeatureMixin):
 
         if scale != 1:
             attrs = data.attrs
-            data = data.map(_rescale)
+            library_ids = data.coords["z"]
+            data = data.map(_rescale).assign_coords({"z": library_ids})
             data.attrs = _update_attrs_scale(attrs, scale)
 
         if mask_circle:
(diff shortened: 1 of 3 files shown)


hspitzer:
@giovp @michalk8 (sorry for bothering you!)
I noticed that the `label_layer` in napari does not work as expected with multiple library_ids. 

Following example code (using the new mibitof dataset):
```python
import squidpy as sq
adata = sq.datasets.mibitof()

library_ids = ['point8', 'point16', 'point23']
imgs = []
for library_id in library_ids:
    img = sq.im.ImageContainer(adata.uns['spatial'][library_id]['images']['hires'], library_id=library_id)
    img.add_img(adata.uns['spatial'][library_id]['images']['segmentation'], library_id=library_id, layer='segmentation')
    img['segmentation'].attrs['segmentation'] = True
    imgs.append(img)
img = sq.im.ImageContainer.concat(imgs)

img.interactive(adata, library_key='library_id')
```

If you then load the "segmentation" layer (which I gave the attribute segmentation to allow napari to recognise it), we have two sliders for switching the z dims at the bottom. One switches "image" (works as expected, and also switches obs), and the other switches only the label layer. I would expect both image + labels to switch together (and for two normal image layers this works). 

I tried to fix this myself, but unfortunately I'm not an expert in napari. If its not an easy fix, we can also leave it as is for now!

michalk8:
Thanks @hspitzer for the PR, I will look into the problems with segmentation.

Issue #353, opened 2021-05-30
Docstring of sq.im.ImageContainer.apply


hspitzer:
The current docstring for kwargs is not very understandable.
```
kwargs
    Keyword arguments for :func:`dask.array.map_blocks` or :func:`dask.array.map_overlap`, depending whether
    ``'depth'`` is present in ``map_kwargs``. Only used when ``chunks != None``.
```

What are ``map_kwargs``? This argument does not exist anymore.
I'm a bit naive about the differences between `map_overlap` and `map_blocks`, and am wondering in which conditions `map_overlap` is used and when `map_blocks`.

PR #356, opened 2021-06-01
Fix dataset docs, update IC docs, bibliography

### LINKED ISSUE #344: add overview page for class methods
### Type of the feature

- [x] Other?

### Description

we'll add a couple of examples for `ImageContainer` methods. For this, I was wondering 
- if it is possible to add the examples to the `ImageContainer` docs (like we add links to examples for the API functions)
- if we can create an overview page for all class methods. E.g. something like the table that we have in the API atm, where you see the function (class method) and a short description. You can then click on these to go to the method docs (can be on the same page). Or, alternatively we could at least have an entry in the navigation bar on the left for each class method. I played around with sphinx a bit, and think this:
```
.. autosummary::
    :toctree: stubs
    :recursive:
    
    ImageContainer
```
does something very similar to what I'm after. 
image

this also seems to add the gallery links - at least to the overview page, which would also be fine.

michalk8:
### Types of changes

### Description

Slightly update IC docs, fix datasets docs (1 wrong name, 1 duplicate), fix spelling + bibliography,
enable doc checking on CI's linting step. Note for future (again) - the dictionaries on CI might (and usually will) differ from local ones, so even if linting passes locally, it might be necessary to add some extra words to `docs/source/spelling_wordlist.txt`.

### How has this been tested?

Not applicable.

### Closes

closes #344  #353


michalk8:
@hspitzer I've added `ImageContainer` to `API`, it achieves the same effect. If it's fine with you, I'd go ahead merge this.

hspitzer:
> @hspitzer I've added `ImageContainer` to `API`, it achieves the same effect. If it's fine with you, I'd go ahead merge this.

Lovely, looks great! This should do well for the time being. To make things really nice, I'd like to avoid duplicating Imagecontainer docs; right now you can click on API->ImageContainer and on classes->ImageContainer which will lead you to different pages. Ideally I'd like to have what is now in API->ImageContainer, but under classes->ImageContainer. But feel free to merge this, I can play around with the different options later to see if I can make this.

PR #357, opened 2021-06-02
allow diagonal setting in spatial_neighbors

AnnaChristina:
### Types of changes

### Description

exposing `set_diag` to the end-user. Example usage in a complex GAT structure where one needs the diagonal to be 1.


giovp:
@AnnaChristina best would be to create the branch from dev and also merge into current dev. We are keeping master free until release

giovp:
@AnnaChristina should be good to go, but if you feel like adding a test for your use case please go on. Otherwise lgtm.

michalk8:
I think for generic and `n_rings > 1`, the diagonal will be still 0. Is this intentional? If yes, I'd mention it in the docs; if not, I'd just write  ` Adj.setdiag(float(set_diag))` [here](https://github.com/theislab/squidpy/blob/neighbors_diagonal/squidpy/gr/_build.py#L118) instead.

Re test: here's one (assuming setting diag to 1 for the above case):
```python
@pytest.mark.parametrize("type_rings", [("grid", 1), ("grid", 6), ("generic", 1)])
@pytest.mark.parametrize("set_diag", [False, True])
def test_set_diag(adata_squaregrid: AnnData, set_diag: bool, type_rings: Tuple[str, int]):
    typ, n_rings = type_rings
    spatial_neighbors(adata_squaregrid, coord_type=typ, set_diag=set_diag, n_rings=n_rings)
    G = adata_squaregrid.obsp["spatial_connectivities"]

    np.testing.assert_array_equal(G.diagonal(), float(set_diag))
```

michalk8:
@AnnaChristina do the changes I've mentioned in https://github.com/theislab/squidpy/pull/357#issuecomment-864075878 work for you? If yes, I'd say we can merge after the tests pass.
Also https://github.com/theislab/squidpy/pull/357/commits/85685bd3df6e8088171a88be4b621739c3499894 fixes that in some cases, diagonal of `distances` was not zero.

AnnaChristina:
> @AnnaChristina do the changes I've mentioned in [#357 (comment)](https://github.com/theislab/squidpy/pull/357#issuecomment-864075878) work for you? If yes, I'd say we can merge after the tests pass.
> Also [85685bd](https://github.com/theislab/squidpy/commit/85685bd3df6e8088171a88be4b621739c3499894) fixes that in some cases, diagonal of `distances` was not zero.

Yes this is perfect! Sorry haven't checked on updates for this recently. Thanks for implementing!

michalk8:
> Yes this is perfect! Sorry haven't checked on updates for this recently. Thanks for implementing!

Thanks a lot! Merging this (even though linting fails [locally is fine], will be dealt in a separate PR).

PR #360, opened 2021-06-07
Fix mibitof dataset, remove astropy

michalk8:
### Types of changes

### Description

Fix shape of `squidpy.datasets.mibitof`, remove `astropy` and `'all'` extras.

### How has this been tested?

N/A

### Closes

N/A


giovp:
No mobitof is wrong, should be ..09. i think you didn't download latest version but still loaded the one in cache

PR #381, opened 2021-07-13
Delaunay optimization

marcovarrone:
### Types of changes

### Description
In the computation of the spatial graph from the Delaunay triangulation, the code to obtain the list of rows and columns is unnecessary given that they are already computed by `scipy.spatial.Delaunay` and thus stored into `tri.vertex_neighbor_vertices`.
The connections are already in a format suitable for the csr_matrix.

Since the building of the csr_matrix between the two methods (delaunay vs. nearest neighbor) is now different I replaced the way of setting the diagonal elements in the case of `set_diag=True` into a way that is independent of how the matrix is built.

The modification results in a speed-up of the `gr.spatial_neighbors` function from:
`928 ms ± 9.53 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`

to
`46 ms ± 2.28 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)`

on the `sq.datasets.visium_fluo_adata` dataset with 2800 observations.

### How has this been tested?
The modifications replace an existing piece of code without adding functionalities and the previous version didn't actually contain a bug, so I only run the `test_spatial_connectivity.py` test.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.0.1
+current_version = 1.1.0
 commit = True
 tag = True
 files = setup.py squidpy/__init__.py
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,7 @@ except ImportError:
             "hannah.spitzer@helmholtz-muenchen.de",
         ]
     )
-    __version__ = "1.0.1"
+    __version__ = "1.1.0"
 
 setup(
     name="squidpy",
--- a/squidpy/__init__.py
+++ b/squidpy/__init__.py
@@ -5,7 +5,7 @@ __email__ = ", ".join(
         "hannah.spitzer@helmholtz-muenchen.de",
     ]
 )
-__version__ = "1.0.1"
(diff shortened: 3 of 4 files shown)


michalk8 (changes requested):
Hi @marcovarrone ,
thanks for the PR, looks great! Could you please fix linting - seems it's just a formatting [issue](https://github.com/theislab/squidpy/pull/381/checks?check_run_id=3059048460#step:6:61).
Once that's done, would be happy to merge.

giovp:
Great catch @marcovarrone ! thank you for the PR!
As mentioned by michal, only minor thing in pre-commit checks. I'd suggest you to `pre-commit install` and then commit again the file, the checks will run and fix the error. Let us know for anything else.

marcovarrone:
Hi @michalk8 and @giovp, thank you for the feedback.
I had some problems with running pre-commit (I fixed it and I will submit another PR for it soon).

I hope everything is fine now!

michalk8:
> I hope everything is fine now!

Looks great, merging this.

PR #393, opened 2021-08-03
Delaunay radius

### LINKED ISSUE #386: gr.spatial_neighbors fails if Delaunay returns isolated nodes
### Description
I notice that `scipy.spatial.Delaunay` may not assign certain nodes to any simplex.
With the current implementation the function `gr.spatial_neighbors` fails because it passes an empty list to the `euclidean_distances` function

### Minimal reproducible example
Unfortunately, I have not been able to reproduce the problem with a simple example.
I cannot understand in which conditions a node is not assigned in any simplex.
I tried to set up a dataset with a group of few nodes close to each other and one node very far from that group but the latter is still connected to some of the others.

### Traceback

<details>

```pytb
~/miniconda3/envs/squidpy/lib/python3.8/site-packages/squidpy/gr/_build.py in spatial_neighbors(adata, spatial_key, coord_type, n_rings, n_neigh, delaunay, radius, transform, key_added)
    115 
    116     elif coord_type == CoordType.GENERIC:
--> 117         Adj, Dst = _build_connectivity(coords, n_neigh, radius, delaunay=delaunay, return_distance=True)
    118     else:
    119         raise NotImplementedError(coord_type)

~/miniconda3/envs/squidpy/lib/python3.8/site-packages/squidpy/gr/_build.py in _build_connectivity(coords, n_neigh, radius, delaunay, neigh_correct, set_diag, return_distance)
    172 
    173                 #if len(vertex_neighs) > 0:
--> 174                 dists_neighs = euclidean_distances(
    175                     coords[vertex_neighs, :], coords[np.newaxis, i, :]
    176                 )

~/miniconda3/envs/squidpy/lib/python3.8/site-packages/sklearn/utils/validation.py in inner_f(*args, **kwargs)
     61             extra_args = len(args) - len(all_args)
     62             if extra_args <= 0:
---> 63                 return f(*args, **kwargs)
     64 
     65             # extra_args > 0

~/miniconda3/envs/squidpy/lib/python3.8/site-packages/sklearn/metrics/pairwise.py in euclidean_distances(X, Y, Y_norm_squared, squared, X_norm_squared)
    270            [1.41421356]])
    271     """
--> 272     X, Y = check_pairwise_arrays(X, Y)
    273 
    274     # If norms are passed as float32, they are unused. If arrays are passed as

~/miniconda3/envs/squidpy/lib/python3.8/site-packages/sklearn/utils/validation.py in inner_f(*args, **kwargs)
     61             extra_args = len(args) - len(all_args)
     62             if extra_args <= 0:
---> 63                 return f(*args, **kwargs)
     64 
     65             # extra_args > 0

~/miniconda3/envs/squidpy/lib/python3.8/site-packages/sklearn/metrics/pairwise.py in check_pairwise_arrays(X, Y, precomputed, dtype, accept_sparse, force_all_finite, copy)
    144                             estimator=estimator)
    145     else:
--> 146         X = check_array(X, accept_sparse=accept_sparse, dtype=dtype,
    147                         copy=copy, force_all_finite=force_all_finite,
    148                         estimator=estimator)

~/miniconda3/envs/squidpy/lib/python3.8/site-packages/sklearn/utils/validation.py in inner_f(*args, **kwargs)
     61             extra_args = len(args) - len(all_args)
     62             if extra_args <= 0:
---> 63                 return f(*args, **kwargs)
     64 
     65             # extra_args > 0

~/miniconda3/envs/squidpy/lib/python3.8/site-packages/sklearn/utils/validation.py in check_array(array, accept_sparse, accept_large_sparse, dtype, order, copy, force_all_finite, ensure_2d, allow_nd, ensure_min_samples, ensure_min_features, estimator)
    667         n_samples = _num_samples(array)
    668         if n_samples < ensure_min_samples:
--> 669             raise ValueError("Found array with %d sample(s) (shape=%s) while a"
    670                              " minimum of %d is required%s."
    671                              % (n_samples, array.shape, ensure_min_samples,

ValueError: Found array with 0 sample(s) (shape=(0, 3)) while a minimum of 1 is required by check_pairwise_arrays.
```

</details>

### Version
1.1.0

The fix is quite simple so I am already working on a PR :)

### LINKED ISSUE #387: Enable radius filter to Delaunay spatial_neighbors
### Type of the feature

- [x] Change an existing functionality, such as default behavior?
- [ ] Other?

### Description
The function `squidpy.gr.spatial_neighbors` already receives `radius` as a parameter for the k-nearest neighbor version, so that nodes farther than the radius value are not linked.

However, if `delaunay=True`, the `radius` parameter is ignored.
I was thinking that after computing the adjacency matrix using Delaunay triangulation we could remove the connections longer than `radius`, if a value for it is set.

### LINKED ISSUE #388: cylindrical shell option for sq.gr.spatial_neighbors
### Type of the feature

- [x] New function in `squidpy.gr`?
- [ ] Other?

### Description

Additional feature request für `sq.gr.spatial_neighbors` to allow a cylindrical shell as neighbourhood option, so all cells are considered as neighbors for min_radius < x < max_radius. This would be beneficial for ncem, see https://github.com/theislab/ncem/issues/22 

As ncem's backend is based on squidpy, I would like to add this.

This would include the following changes: rename `radius` to `max_radius` and add `min_radius` with default 0. I can create a PR for this.

michalk8:
### Types of changes

### Description

- `n_neighs` instead of `n_neigh`/`n_neigh_grid` (controls both grid/generic)
- `radius` can be specified as [min, max], keeps edges only in `[min, max]` interval (mb use `(min, max)`)
- `radius` in the above form can also affect graph from Delaunay triangulation
- fixed the rare bug in #386
- add `copy: bool = False`
- improve the docstrings

### How has this been tested?

Add 2 tests: 1 for copy, 1 parametrized for radius as a tuple.

### Closes

closes #388
closes #387
closes #386


michalk8:
Last set of tests seem to have [passed](https://github.com/theislab/squidpy/actions/runs/1095019859), ready for a review.

squidpy/gr/_build.py
```diff
@@ -46,43 +47,48 @@ def spatial_neighbors(
...
-    delaunay
-        Whether to compute the graph from Delaunay triangulation. The argument is used if `coord_type={c.GENERIC!r}`.
+            - `{c.GRID.s!r}` - number of neighboring tiles.
+            - `{c.GENERIC.s!r}` - number of neighborhoods for non-grid data. Only used when ``delaunay = False``.
     radius
-        Radius of neighbors for non-grid data. The argument is used if `coord_type={c.GENERIC!r}`.
+        Depending on the type:
+
+            - :class:`float` - neighborhood radius for non-grid data if ``coord_type = {c.GENERIC.s!r}``
+              and ``delaunay = False``.
```
giovp, line 63: current behaviour does not prune the graph if `delaunay=True`. Therefore, specifying `delaunay=False` in docs is confusing. I think the current behaviour is fine (just prune when Tuple is passed) but maybe should be clarified a bit?

Proposal:
```
- :class:`float` - compute the graph based on neighborhood radius. Only used when ``coord_type = {c.GENERIC.s!r}``.

- :class:`tuple` - prune the final graph to only contain edges in interval `[radius[0], radius[1]]`. Can be applied to any ``coord_type = {c.GENERIC.s!r}`` mode of graph construction.
```

the `Only used when ``coord_type = {c.GENERIC.s!r}``.` think is useful to make it consistent with delaunay

squidpy/gr/_build.py
```diff
@@ -151,75 +139,107 @@ def spatial_neighbors(
...
 
+    if copy:
+        return Adj, Dst
+
     _save_data(adata, attr="obsp", key=conns_key, data=Adj)
     _save_data(adata, attr="obsp", key=dists_key, data=Dst, prefix=False)
     _save_data(adata, attr="uns", key=neighs_key, data=neighbors_dict, prefix=False, time=start)
 
 
+def _build_grid(
```
giovp, line 154: right call to get this out of the main function

tests/graph/test_spatial_neighbors.py
```diff
@@ -0,0 +1,127 @@
+from typing import Tuple
```
giovp, line 1: awesome refactoring of spatial neighbors test, looks really great!

giovp (approved):
tested it on couple of use cases and seem to work really well, awesome job as always @michalk8 !

@AnnaChristina if you can give it a spin for your use case would be nice, otherwise LGTM.

squidpy/gr/_build.py
michalk8: > current behaviour does not prune the graph if delaunay=True.

I think I get it now - I just wanted to highlight that it's impossible to only specify 1 `float` when `delaunay = True`, which for KNN is the number of neighborhood radius (to threshold it, you need both min and max, as for all other cases).
I will remove the `delaunay` mention in this case.

Re the `tuple` change you've suggested: it's not correct, you can also threshold using `grid` to build the graph and I fail to see why it shouldn't (see the case of `n_rings > 1`).

giovp: >  it's not correct, you can also threshold using grid to build the graph and I fail to see why it shouldn't (see the case of n_rings > 1).

this I haven't tried, but if that's the case, I'd remove this behaviour. The idea of "grid" is that it completely disregards the "unit measure" and just build graph based on a regular grid (according to data). The `min_dist` between 2 nodes ina regular grid can be 0.1 or 1000 but it doesn't matter, as a grid graph will be built. Therefore, there is no meaning for a "radius" value in grid, because a radius is based on a "distance" in some "unit measure". In summary, imho there shouldn't be any method from "generic" that also work on "grid".

giovp: happy to discuss further if you disagree btw, but to the logics are very different to me, as well as the generated manifolds

michalk8: No I think it makes sense - guess I was confused why we then output both the similart and distances in this case as well.

giovp:
@michalk8 I think anna is on holidays, you can merge this

PR #396, opened 2021-08-10
Fix mypy error

michalk8:
### Types of changes

### Description

...

### How has this been tested?

...

### Closes

...


michalk8:
Hmm, I can't render the type annotation on RTD, since 3.9 is required. Will revert the `|` for union types...

michalk8:
Linting will most likely fail because `cellprofiler.org` is down (504).

PR #398, opened 2021-08-12
Make sure flake8-logging-format are used

shiplu:
By Default flake8 disables all extensions. Even though flake8-logging-format is installed it's not used actually. For that we need to add `--enable-extensions G` args to `flake8` command.

### Types of changes

### Description
Makes sure contributors can not push code that doesn't have good looking format.
...

### How has this been tested?
Yes
...

### Closes

...


michalk8:
Hi @shiplu ,
thanks for the PR, didn't know you need to do this to have it enabled. Upon further consideration, Ithink we can get rid of this extension in `flake8` altogether (the [optimazations](https://docs.python.org/3/howto/logging.html#optimization) in our case are not worth the effort imho).
I want to use `f-strings` to interpolate the when passing the data (), so `G004` would have to be ignored. The [linting](https://github.com/theislab/squidpy/pull/398/checks?check_run_id=3311196669#step:6:101) picked up in 2-3 places `G200`; problem is this can't be easily fixed, because, `scanpy`'s logging doesn't expose `exc_info`.
If you remove `flake8-logging-format` from `.pre-commit.yml`, than I will merge this PR.

michalk8:
closed via #412

PR #432, opened 2021-11-23
Add intervals argument to sq.gr.co_occurrence

hspitzer:

## Description

Allow to specify custom intervals for :func:`squidpy.gr.co_occurrence` and rename ``n_steps`` -> ``interval``. Useful e.g. when a log-space interval is more appropriate than the default linspace interval.

## How has this been tested?

1 new parametrized test

## Closes

N/A


hspitzer:
@michalk8 the plotting tests seem to fail. Any idea why? I did not change any plotting code and the default behaviour of the `interval` argument is as before.

michalk8:
> @michalk8 the plotting tests seem to fail. Any idea why? I did not change any plotting code and the default behaviour of the interval argument is as before.

No idea, must be because of a new napari version, would commit the regenerated files.

michalk8:
@hspitzer do you think this can be merged? I could quickly finalize it, just write 1 test.

hspitzer:
@michalk8 thanks for reminding me of this! We could join n_steps and interval arguments, but to stay consistent with the current version, it should be called n_steps, and just additionally accept an interval argument.

If its quick for you, I'm happy for you to add a test. Otherwise I'll put it on my list and get to it in the next week.

michalk8:
> If its quick for you, I'm happy for you to add a test. Otherwise I'll put it on my list and get to it in the next week.

@hspitzer I've added 1 new test (and cleaned the function a bit). I've also renamed ``n_steps`` to ``interval: int | array = 50`` since it feels more natural. If you approve, I will merge (linting is related to tutorials, they just need to be regenerated).
Note to self: check whether in tutorials/examples we explicitly use ``n_steps``

hspitzer (commented):
looks great, thanks a lot @michalk8 . I like the intervals argument better than n_steps, we should just check that we update tutorials if its used there.

Good to merge for me.

Issue #392, opened 2021-07-30
Issue with Squidpy's installation


tesalobo:
First of all I wanted to congratulate you and your team on the development of Squidpy and thank you for pouring so much work into building such a detailed documentation like Squidpy's.

The reason I am reaching to you is because I am trying to make use of your python package but it rises an error when I try to install it through the terminal like `pip install squidpy`.

The error pops up when trying to install tables:

```
Building wheels for collected packages: tables
Building wheel for tables (setup.py) ... error
ERROR: Command errored out with exit status 1:
```

Is there something I am doing wrong? Have you found this error too and are on the process of correcting it? If you could bring some light into this obstacle I have encounter it would be very appreciated. (At the end of the email you will find some details about the environment where I want to install Squidpy, in case it would be of any help)

Thank you very much in advance for your time and help,

--------------------------------------------------------------------------------------------------------------------------------------------

I have tried to install it on both, a windows and a Mac, and the same error keeps coming up. 

My python version in python 3.9 and these are the libraries inside the environment to which I would like to add Squidpy:

```
# packages in environment at /usr/local/Caskroom/miniconda/base/envs/GIST_env:
#
# Name                    Version                   Build  Channel
absl-py                   0.13.0                   pypi_0    pypi
alabaster                 0.7.12                   pypi_0    pypi
appdirs                   1.4.4                    pypi_0    pypi
appnope                   0.1.2                    pypi_0    pypi
argon2-cffi               20.1.0                   pypi_0    pypi
astunparse                1.6.3                    pypi_0    pypi
async-generator           1.10                     pypi_0    pypi
attrs                     21.2.0                   pypi_0    pypi
babel                     2.9.1                    pypi_0    pypi
backcall                  0.2.0                    pypi_0    pypi
blas                      1.0                         mkl  
bleach                    3.3.1                    pypi_0    pypi
ca-certificates           2021.7.5             hecd8cb5_1  
cached-property           1.5.2                      py_0  
cachetools                4.2.2                    pypi_0    pypi
cachey                    0.2.1                    pypi_0    pypi
certifi                   2021.5.30        py39hecd8cb5_0  
cffi                      1.14.6                   pypi_0    pypi
charset-normalizer        2.0.3                    pypi_0    pypi
cloudpickle               1.6.0                    pypi_0    pypi
cycler                    0.10.0                   pypi_0    pypi
dask                      2021.7.0                 pypi_0    pypi
debugpy                   1.4.0                    pypi_0    pypi
[...]

PR #436, opened 2021-12-02
Remote Zarr Image Container

ilan-gold:
## Description
This PR is somewhat short, but I learned a lot along the way!

This PR basically adds support for a zarr xarray dataset to be stored remotely, on a file server, and then read in by squidpy.

You can test this by doing something like:

1. Starting a http server:
```bash
brew install http-server
http-server -p 8080
```

2. And then:
```python
import squidpy as sq
img = sq.im.ImageContainer('http://127.0.0.1:8080/test_visium.zarr/')
crop = img.crop_corner(1000, 1000, size=1000)
sq.im.segment(img=crop, layer="image", method="watershed", channel_ids=0)
crop.show('segmented_watershed')
```

And you should see the image data flowing out of the server and into squidpy (`http-server` logs requests).

Part of the reason this took me so long is that I had to figure out how to get the data to load lazily (without confirming this, I can only be certain I've just written syntatic sugar for a download method  :smile:).  There were a few things I noticed here:

1.  I needed to add special arguments for lazy-loading.   I think that because dask is used in `squidpy`, I assumed this would be the default.  However without these arguments, the `segment` call prompted the actual computation of the segmentation which was unexpected.  Is this right for lazy-loading in terms of arguments? Too much?  I was a little confused as to `chunks` vs `lazy`.
```python
import squidpy as sq
# This doesn't seem to work without the `chunks` argument, which appears to be undocumented:
# https://squidpy.readthedocs.io/en/latest/classes/squidpy.im.ImageContainer.html
img = sq.im.ImageContainer('http://127.0.0.1:8080/test_visium.zarr/', chunks='auto')
crop = img.crop_corner(1000, 1000, size=1000)
sq.im.segment(img=crop, layer="image", method="watershed", channel_ids=0, chunks='auto', lazy=True)
crop.show('segmented_watershed')
```

2.   It took me a very long time to figure out that tests don't seem to work with python 3.9 on macos: #392 and https://github.com/theislab/squidpy/blob/c9317870fbac631de523596737dc837c8bdb66b5/.github/workflows/test.yml#L42-L54 or at least I couldn't get around a very starnge hdf5 issue that seems to have been documented in #392.  Is this still an issue?  The issue itself is closed but the CI doesn't run the tests on macos for python 3.9 it seems.

3.  Running the following computes the watershed over the whole image and then only shows the crop.  I guess this makes sense to some degree since you ask the computation to be done on the whole image and then only show a crop of it, especially given that `otsu` thresholding requires loading all the data, but it took me a bit to figure this out.  I thought the computation task would transfer to the crop independently.   Maybe some documentation about how lazy-loading in squidpy work?  Perhaps we could do this hand-in-hand with the first point?

```python
import squidpy as sq
img = sq.im.ImageContainer('http://127.0.0.1:8080/test_visium.zarr/', chunks='auto')
sq.im.segment(img=img, layer="image", method="watershed", channel_ids=0, chunks='auto', lazy=True)
crop = img.crop_corner(1000, 1000, size=1000)
crop.show('segmented_watershed')
```

The last thing is OME-NGFF.  I'm not sure what the best way forward is here.  One thing is that the data format is just completely different from `xarray`'s own data format, which seems to have some metadata and structure requirements:  https://xarray.pydata.org/en/stable/generated/xarray.open_dataset.html

Another thing is that OME-NGFF is mutliscale while `squidpy` and `xarray` are not from what I can tell.  So I'm not totally sure what to do about that either.  I'll maybe open an issue for this and continue the discussion there?

## How has this been tested?
I updated the io unit test to include remote datasets


giovp:
@michalk8 why did you close this? I think it'd be cool to have

michalk8:
> @michalk8 why did you close this? I think it'd be cool to have

Sorry, must've happened automatically when I removed `dev`. @ilan-gold would be great if you could reopen.

ilan-gold:
I cannot reopen.....

ilan-gold:
I can just open a new PR though unless one of you has the option to re-open

giovp:
I also can't reopen 😓 sorry ! better new PR

michalk8:
> I can just open a new PR though unless one of you has the option to re-open

Yeah, sorry for causing confusion, please open a new PR.

michalk8:
@ilan-gold could you please open a new PR?

ilan-gold:
Done @michalk8 but I think there are merge conficts

PR #438, opened 2021-12-30
Fix an import for python 3.10

dcjones:
This should let squidpy work with python 3.10, by importing `Iterable` from the right place.


michalk8:
> This should let squidpy work with python 3.10, by importing Iterable from the right place.

Thanks a lot, I just fixed the linter.
Can I make @giovp ``dev`` the default branch (for CI cron job + to be the default PR merge-branch)? Should've been done a long time ago imho.

PR #477, opened 2022-03-14
Fix grammar in docs

ivirshup:
The grammar for the napari bullet is a bit off. I think this will make more sense.


michalk8:
Could you please also fix also `README.rst` and `README_pypi.rst`? Same grammar is used there.

PR #493, opened 2022-03-22
Add ax keyword to nhood enrichment plot

### LINKED ISSUE #481: Retrieve matplotlib figure instance from plot function
Hi all,

I was wondering whether it could be possible to retrieve the matplotlib figure instance created within a nhood enrichment test, or - even better - pass a matplotlib axes object to `sq.gr.nhood_enrichment()` to allow some more customized plotting?

Thanks in advance!

jo-mueller:
## Description
An optional ``ax`` keyword can now be passed to :func:`squidpy.pl.nhood_enrichment` function. The keyword can be used to create a matplotlib figure outside of squidpy and then plot on any matplotlib subplot, which could give the user greater flexibility in working with nhood enrichment plots.

## How has this been tested?
Added a test in `test/plotting/test_graph.py` which creates a figure on a matplotlib axes object.

## Closes
Closes #481


squidpy/pl/_graph.py
```diff
@@ -13,6 +13,7 @@
 import pandas as pd
 
 import seaborn as sns
+import matplotlib as mpl
```
michalk8, line 16: Would rather use `from matplotlib.axes import Axes`

squidpy/pl/_graph.py
```diff
@@ -208,6 +209,7 @@ def nhood_enrichment(
     figsize: tuple[float, float] | None = None,
     dpi: int | None = None,
     save: str | Path | None = None,
+    ax: mpl.axes._subplots.Axes = None,
```
michalk8, line 212: Should be typed as `Axes | None`.

tests/plotting/test_graph.py
```diff
@@ -57,6 +59,9 @@ def test_plot_nhood_enrichment(self, adata: AnnData):
 
         pl.nhood_enrichment(adata, cluster_key=C_KEY)
 
+        fig, ax = plt.subplots()
```
michalk8, line 62: I believe the test fails because a) the figsize is set automatically [here](https://github.com/theislab/squidpy/blob/master/squidpy/pl/_graph.py#L250) and b) `contrained_layout=True` when creating the figure.

michalk8 (changes requested):
Hi @jo-mueller , thanks for the PR, looks good. Just 2 minor changes
and after the tests pass (see my comment, hope it will fix it), will be happy to merge!

jo-mueller:
Hey @michalk8 ,

thanks for the hints, I hope the above changes fix the problems!

giovp:
the dendrogram seesm to be failing: 
image

is that maybe cause it's not part of the `ax` ?

giovp (approved):

michalk8 (approved):

michalk8:
Think this can be merged. In the future, we can add ``ax`` to other functions which use ``_heatmap``.

giovp:
thanks again @jo-mueller ! LGTM!

Issue #507, opened 2022-04-06
Squidpy is not backwards compatible with ImageContainers that do not contain a z-dim


MxMstrmn:
## Description

Newer verions of squidpy (require z dim) do not load ImageContainer that were generated on another version of squidpy 

## Minimal reproducible example

```python
# requires ImageContainer that does not contain a z-dim
image = sq.im.ImageContainer().load(DATA_DIR/"ImageContainers"/"37_48_A1")
```
Happy to share `DATA_DIR` with you @giovp, @michalk8 if interested. 
## Traceback

<details>

```pytb
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-7-2f1703ad9406> in <module>
----> 1 image = sq.im.ImageContainer().load(DATA_DIR/"ImageContainers"/"37_48_A1", )

/mnt/home/icb/leon.hetzel/git/squidpy/squidpy/im/_container.py in load(cls, path, lazy, chunks)
    200         """
    201         res = cls()
--> 202         res.add_img(path, layer="image", chunks=chunks, lazy=True)
    203 
    204         return res if lazy else res.compute()

/mnt/home/icb/leon.hetzel/git/squidpy/squidpy/im/_container.py in add_img(self, img, layer, dims, library_id, lazy, chunks, copy, **kwargs)
    284             InferDimensions(dims) if isinstance(dims, str) else dims
    285         )
--> 286         res: xr.DataArray | None = self._load_img(img, chunks=chunks, layer=layer, copy=copy, dims=dims, **kwargs)
    287 
    288         if res is not None:

~/miniconda3/envs/py38/lib/python3.8/functools.py in _method(*args, **kwargs)
    910         def _method(*args, **kwargs):
    911             method = self.dispatcher.dispatch(args[0].__class__)
--> 912             return method.__get__(obj, cls)(*args, **kwargs)
    913 
    914         _method.__isabstractmethod__ = self.__isabstractmethod__

/mnt/home/icb/leon.hetzel/git/squidpy/squidpy/im/_container.py in _(self, img, chunks, dims, **_)
    364                 raise ValueError("Loading data from `Zarr` store is disallowed when the container is not empty.")
    365 
--> 366             self._data = transform_metadata(xr.open_zarr(str(img), chunks=chunks))
    367         elif suffix in (".nc", ".cdf"):
    368             if len(self._data):

[...]

PR #509, opened 2022-04-06
Fix missing Z-dim

michalk8:

## Description

Add singleton Z-dim when loading from *Zarr* and not present.

## How has this been tested?

1 new test.

## Closes

further addresses #507


MxMstrmn:
Nice man! 🙏 My handcrafted way looked something like this: 

```python
z_image = sq.im.ImageContainer()

new_img = np.zeros((21936,21936,1,4))
i = 0 
for layer in image: 
    if layer == 'bf': 
        continue
    new_img[...,i] = image[layer].values
    i += 1 
z_image.add_img(new_img, layer='37_48_A1')
```

Not efficient whatsoever, but does the job (if you have sufficient memory) :D

PR #572, opened 2022-06-17
fix the bug of relative coordinates

JiaShun-Xiao:
## Description

I have changed the relative coordinate to exact coordinate in the `feature_segmentation` function.
There are two reasons:
1. because of the normalization of "np.max(y) - np.min(y)" in denominator, current relative coordinate will return nan if there is only one cell in spot.
2. exact coordinate looks more reasonable than the relative coordinate when visualizing the segmentation result with spatial image as background.

## How has this been tested?

I have tested the revised code in two Visium data, it works very well.

### DIFF
--- a/squidpy/im/_feature_mixin.py
+++ b/squidpy/im/_feature_mixin.py
@@ -343,12 +343,12 @@ class FeatureMixin:
             y_slc, x_slc = coord.to_image_coordinates(padding).slice
 
             # relative coordinates
-            y = (y - np.min(y)) / (np.max(y) - np.min(y))
-            x = (x - np.min(x)) / (np.max(x) - np.min(x))
+            #y = (y - np.min(y)) / (np.max(y) - np.min(y))
+            #x = (x - np.min(x)) / (np.max(x) - np.min(x))
 
             # coordinates in the uncropped image
-            y = coord.slice[0].start + (y_slc.stop - y_slc.start) * y
-            x = coord.slice[1].start + (x_slc.stop - x_slc.start) * x
+            y = coord.slice[0].start + y#(y_slc.stop - y_slc.start) * y
+            x = coord.slice[1].start + x#(x_slc.stop - x_slc.start) * x
 
             return np.c_[x, y]  # type: ignore[no-any-return]
 


giovp:
hi @JiaShun-Xiao thanks for the cotnribution, can you elaborate a bit more and provide code snippets about this PR? I don't think I fully understand teh settigns in which this fails. Also, can you check out the [contributing guide](https://github.com/scverse/squidpy/blob/main/CONTRIBUTING.rst) for pre-commit checks (currently failing).

Thank you!

giovp:
hi @JiaShun-Xiao ,

could you please elaborate on the purpose of this PR? 

thank you

JiaShun-Xiao:
Hi, @giovp 

Thanks for your patience!
Current relative coordinates are fine in most cases. However, I think it could be improved by using exact coordinates as I mentioned in the first comment. 
The nuclei segmentation example (https://squidpy.readthedocs.io/en/latest/external_tutorials/tutorial_tangram.html) provided in squidpy document can demonstrate this.

Thank you!
Jiashun

giovp:
interesting ok, the tests seem to pass, could you add a small test with this specific edge case? Otherwise a minimal reproducible example here and I'd be happy to add it myself, thank you!

JiaShun-Xiao:
Sorry I am not familiar with tests in programming, I need some time to learn how to add tests for edge cases

giovp:
hi @JiaShun-Xiao , no problem. Could you please then post here few lines of code that explain what is the problem you are trying to solve? ideally something that can be run directly, with minimal dependency (e.g. squidpy, numpy etc)

JiaShun-Xiao:
Sure, the problem I am trying to solve is retriving exact cell coordinates/locations after nuclei segmentation with spatial histological images.
To demonstrate this, I add an example in the following Jupiter notebook to illustrate why we need exact nuclei coordinates instead of relative coordinates.

[squidpy-mouce-brain-MOp-demo.ipynb.zip](https://github.com/scverse/squidpy/files/9307070/squidpy-mouce-brain-MOp-demo.ipynb.zip)

[squidpy-mouce-brain-MOp-demo.pdf](https://github.com/scverse/squidpy/files/9307090/squidpy-mouce-brain-MOp-demo.pdf)

PR #592, opened 2022-08-18
remove sorted for axis in `gr.ligrec`

giovp:


giovp:
HI @wangjiawen2013 ,

apologies for the continuous back and forth. The solution of removing the sorting for ligrec was not optimal, and while playing around I realized that it is in fact possible to pass the order fo the target (or source group) directly at input of the plotting function, with something like this:

```python
import squidpy as sq
adata = sq.datasets.seqfish()
import pandas as pd

adata.obs["leiden"] = pd.Categorical(
    np.random.randint(0, 15, size=adata.obs.Area.shape)
)
res = sq.gr.ligrec(
    adata,
    n_perms=1,
    cluster_key="leiden",
    copy=True,
    use_raw=False,
    transmitter_params={"categories": "ligand"},
    receiver_params={"categories": "receptor"},
)
sq.pl.ligrec(
    res,
    source_groups="1",
    target_groups=adata.obs.leiden.cat.categories.astype(str),
    alpha=0.005,
)
```

image


you can see that the cluster order is now based on categories. I will therefore close this PR and the linked issue.

PR #609, opened 2022-10-08
fix ligrec from pandas update

giovp:

## Description

@michalk8 tests fail with pandas=1.5.0 in master (but don't with pandas 1.4.0). I tried to fix it with a hack and now some tests pass but I'd like you to take a look since I'm not sure it covers behaviour.

It might (but not sure) be related to this recent bugfix https://github.com/pandas-dev/pandas/issues/45715


michalk8:
> @michalk8 tests fail with pandas=1.5.0 in master (but don't with pandas 1.4.0). I tried to fix it with a hack and now some tests pass but I'd like you to take a look since I'm not sure it covers behaviour.

Thanks, will check!

PR #612, opened 2022-10-11
misc

giovp:

## Description
Change docs theme with Furo, update release note.


ivirshup:
@giovp, did you see @adamgayoso wants to switch the cookiecutter from furo to sphinx_book_theme? 

* scverse/cookiecutter-scverse#100

adamgayoso:
I think both are good, but the book theme requires is more deployable with zero changes. I'd be content if you picked whatever makes you happy :)

Both furo and book theme look better than the RTD theme IMO

giovp:
need to play around with css, will address with separate PR, thanks all for suggestions

PR #679, opened 2023-04-19
Convert cell_metadata index from integer to string for latest MERFISH public data

djlee1:
## Description

In the latest MERFISH public data, the cell_metadata index is recognized as an integer, which requires conversion to a string format.  This commit adds a function to convert the index to a string, ensuring compatibility with downstream applications.

## How has this been tested?

When attempting to load the latest public data from https://vizgen.com/data-release-program/, specifically the FFPE_Human_Immuno_oncology dataset, I have encountered an issue where the 'obs' of the Anndata object is filled with NaN values. This issue is caused by the presence of different index types between 'adata.obs' and 'coords', which leads to errors when merging them.

I have found that this issue can be resolved by setting the data type of 'coords' as 'str'. After making this adjustment, the 'obs' of the Anndata object is populated with the expected data.

### DIFF
--- a/squidpy/read/_read.py
+++ b/squidpy/read/_read.py
@@ -89,7 +89,7 @@ def visium(
     )
     coords.columns = ["in_tissue", "array_row", "array_col", "pxl_col_in_fullres", "pxl_row_in_fullres"]
     # https://github.com/scverse/squidpy/issues/657
-    coords.set_index(coords.index.astype(adata.obs.index.dtype), inplace=True)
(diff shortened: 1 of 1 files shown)


dfhannum:
I think this was addressed a previous [pull request](https://github.com/scverse/squidpy/pull/665)

dfhannum:
I've used the change in my local copy of squidpy successfully for the past weeks. When I downloaded the new version things stopped working today. 

I then realized that the change occurred in the wrong function. This change was in "def visium()" but it was meant for "def vizgen()"

PR #690, opened 2023-05-16
Add percentile threshold filtering to spatial_neighbors

LLehner:
Update of `_spatial_neighbor()` to allow for filtering the graph based on percentile threshold.


squidpy/gr/_build.py
```diff
@@ -201,6 +208,12 @@ def _spatial_neighbor(
         Dst.data[mask] = 0.0
         Adj.data[mask] = 0.0
         Adj.setdiag(a_diag)
+        radius_filtered = True
+
```
LLehner, line 212:
```suggestion
elif percentile is not None and coord_type == CoordType.GENERIC: #apply percentile filtering
```

squidpy/gr/_build.py
```diff
@@ -201,6 +208,12 @@ def _spatial_neighbor(
...
+    if apply_percentile and not radius_filtered:
+        threshold = np.percentile(np.array(Dst[Dst != 0]).squeeze(), percentile)
```
LLehner, line 214: return non-zero values with .data

squidpy/gr/_build.py
```diff
@@ -201,6 +208,12 @@ def _spatial_neighbor(
...
+        Adj[Dst > threshold] = 0.0
+        Dst[Dst > threshold] = 0.0
```
LLehner, line 217: add test

giovp (approved):

PR #701, opened 2023-06-06
Add spatialdata as possible input type to graph functions

LLehner:
Add spatialdata as possible input type to graph functions.

If a spatialdata object is provided instead of an anndata object, retrieve the table from spatialdata.


timtreis (approved):

timtreis:
@LLehner any idea why the tests on Ubuntu fail?

PR #738, opened 2023-08-03
Deprecate Napari

LLehner:
Add @deprecate decorator to various functions and classes.


giovp:
just seeing this, I think it's enough to just put it in the entry point so theinteractive class.

giovp:
@LLehner it seems the deprecated module is missing from dependency, can you add and re-test, thanks!

Issue #797, opened 2024-02-06
NaN value when importing Visium dataset


Lem-P:
Hi,
I am trying to import a dataset from 10X Visium H&E

I am importing the dataset with:
```python
adata = sq.read.visium('path')
adata.var_names_make_unique()
```
Then pre-processing:
```python
sc.pp.filter_cells(adata, min_counts = 1000)
sc.pp.filter_genes(adata, min_cells=5)
sc.pp.normalize_total(adata, inplace = True)
sc.pp.log1p(adata)
sc.pp.highly_variable_genes(adata, flavor='seurat', n_top_genes=4000, inplace=True)
sc.pp.pca(adata, n_comps=50, use_highly_variable=True, svd_solver='arpack')
sc.pp.neighbors(adata)
sc.tl.umap(adata)
sc.tl.louvain(adata, key_added='clusters')
```

Then to calculate the image_features, I create my ImageContainer (not clear to me if I can do it before filtering or not).
```python
library_id = "Mouse_2"
img = sq.im.ImageContainer(
    adata.uns["spatial"][library_id]["images"]["hires"],
    scale=adata.uns["spatial"][library_id]["scalefactors"]["tissue_hires_scalef"],
)
```
No problem until that point (except a lot of warnings about deprecated parameter in pandas)

I then do:

```python
for scale in [1.0, 2.0]:
    feature_name = f"features_summary_scale{scale}"
    sq.im.calculate_image_features(
        adata,
        img.compute(),
        features="summary",
        key_added=feature_name,
        n_jobs=4,
        scale=scale,
    )
```

but get this error:

## Traceback

<details>

```pytb
in ImageContainer.generate_spot_crops(self, adata, spatial_key, library_id, spot_diameter_key, spot_scale, obs_names, as_array, squeeze, return_obs, **kwargs)
    820 radius = int(round(diameter // 2 * spot_scale))
    822 # get coords in image pixel space from original space
--> 823 y = int(spatial[i][1] * scale)
    824 x = int(spatial[i][0] * scale)
    826 # if CropCoords exist, need to offset y and x

ValueError: cannot convert float NaN to integer
```

</details>

If trying to do neighborhood enrichment with
sq.gr.spatial_neighbors(adata)

I got this
ValueError: Input X contains NaN.

## Version

squidpy==1.4.1

giovp:
can you check whether you have nan in `adata.obsm["spatial"]`

Lem-P:
[...]

PR #803, opened 2024-02-17
fixed reading 10x formatted visium mtx files

### LINKED ISSUE #802: Issue reading visium data when the GEX file is a matrix
## Description

When using a mtx file to input the expression data, squidpy fails to match the `tissue_positions_list.csv` data to the file. I believe this is an issue due to how the file is read in, specifically since it utilizes `read_mtx` from the anndata package. Since the cell barcodes aren't contained in the mtx file, and instead in the `barcodes.tsv.gz` file, when the code tries to match the indices from `tissue_positions_list.csv`, none of them are contained in the anndata object. This leads to all the fields in the filled as `nan`. 

...

## Minimal reproducible example

```
adata = sq.read_visium("some/directory", counts_file="matrix.mtx", library_id='some_id')
```

</details>

## Version

1.4.1

...

LinearParadox:
solves #802

## Description

Replaced Anndata's read_mtx function with scanpys native read_10x_mtx function for reading mtx formatted visium experiments. This solves issues with matching the spot positions to barcodes.

## How has this been tested?

all tests ran locally 

## Closes

closes #802

### DIFF
--- a/src/squidpy/read/_utils.py
+++ b/src/squidpy/read/_utils.py
@@ -4,10 +4,10 @@ from pathlib import Path
 from typing import Any, Optional, Tuple
 
 import numpy as np
-from anndata import AnnData, read_mtx, read_text
+from anndata import AnnData, read_text
 from h5py import File
 from PIL import Image
-from scanpy import read_10x_h5
+from scanpy import read_10x_h5, read_10x_mtx
 
 from squidpy._constants._pkg_constants import Key
 from squidpy._utils import NDArrayA
@@ -49,7 +49,7 @@ def _read_counts(
     if count_file.endswith((".csv", ".txt")):
         adata = read_text(path / count_file, **kwargs)
     elif count_file.endswith(".mtx"):
-        adata = read_mtx(path / count_file, **kwargs)
+        adata = read_10x_mtx(path, **kwargs) 
     else:
         raise NotImplementedError("TODO")
 


giovp:
thank you @LinearParadox ! I wonder if it's related to #797 , @Lem-P could you try out this solution and see if it fixes the issue?

LinearParadox:
I think it's likely a different issue. When squidpy would read the mtx file, the object would have no gene / barcode names, and the failure would likely be noticed at qc (as you would have only a single peak in spot counts, since they'd be all NA). Do you know if the current test datasets in squidpy have the mtx files floating around somewhere? It would be nice to add some tests to make 100% sure the mtx reading is equivalent with reading h5 files.

giovp:
hi @LinearParadox , you are right, re-reading the issue I think it's due to the header being set wrong. 

regarding tests, unfortunately we don;'t have the mtx file in thedatasets, and we don't really have a test suite for readers in squidpy. For spatialdata, we have a robust and rather compute expensive pipeline that ensures that datasets can always be read by the readers and supported by spatialdata objects (set up by @LucaMarconato , cc). I think it's outside the scope of squidpy to set something up like that as we are hoping that users will transition to spatialdata eventually. You can find the (tested) datasets here: https://github.com/scverse/spatialdata-notebooks/tree/main/datasets . I will merge this once CI passes

giovp:
maybe @LinearParadox can you add a release note? you can add it in the `dev` release for now (in `docs/release_notes`) and I will move it to whatever version we do next. Thank you!

LinearParadox:
Should be done! Let me know if you need me to add more detail or if there's anything else I should modify/add!

docs/release/notes-dev.rst
```diff
@@ -1,2 +1,3 @@
 Squidpy dev (the-future)
 ========================
+Fixed the reading of 10x formatted mtx files
```
giovp, line 3:
```suggestion
- Fixed the reading of 10x formatted mtx files `@LinearParadox <https://github.com/LinearParadox>`__
  `#803 <https://github.com/scverse/squidpy/pull/803>`__
```

giovp (approved):

PR #807, opened 2024-03-04
Add method to calculate embeddings for variable by distance aggregation

LLehner:
## Description

Adds a method in ```tools``` to calculate embeddings of variables by their counts aggregated by distance.

## Example usage

`import squidpy as sq`

load example data set
`adata = sq.datasets.seqfish()`

Calculate distances of each observation to a specified anchor point (e.g. cell type or tissue location). Here we use cell type "Endothelium" in the annotation column "celltype_mapped_refined":
`sq.tl.var_by_distance(adata, groups="Endothelium", cluster_key="celltype_mapped_refined")`

The resulting distances are stored in `adata.obsm["design_matrix"]`. Now we can calculate the embeddings:
`sq.tl.var_embeddings(adata, cluster_key="Endothelium", design_matrix_key="design_matrix")`

Note that by default the bin of distance 0, meaning the counts that belong to the anchor point, are excluded. This can be changed by setting `include_anchor=False` in `sq.tl.var_embeddings()`.

By default 100 bins are used. The resulting embeddings are stored in `adata.uns["100_bins_distance_embeddings"]`.
We can plot the embedding (umap) as follows:

`embedding = adata.uns["100_bins_distance_embeddings"]`
`plt.scatter(embedding[0], embedding[1], c="grey")`
`plt.gca().set_aspect('equal', 'datalim')`
`plt.title('UMAP', fontsize=20)`

which results in:
image

## TODO

- [ ] Add a plotting function so this doesn't need to be done manually.
- [ ] Cluster the data

### DIFF
--- a/src/squidpy/tl/__init__.py
+++ b/src/squidpy/tl/__init__.py
@@ -3,3 +3,4 @@
 from __future__ import annotations
 
 from squidpy.tl._var_by_distance import var_by_distance
+from squidpy.tl._var_embeddings import var_embeddings
--- a/src/squidpy/tl/_var_embeddings.py
+++ b/src/squidpy/tl/_var_embeddings.py
@@ -1,0 +1,92 @@
+from __future__ import annotations
+
+
+import pandas as pd
+from anndata import AnnData
+from scanpy import logging as logg
+from sklearn.preprocessing import StandardScaler
+import umap
+
+from squidpy._docs import d
+
+__all__ = ["var_embeddings"]
+
+
+@d.dedent
+def var_embeddings(
+    adata: AnnData,
+    cluster_key: str,
+    design_matrix_key: str = "design_matrix",
+    n_bins: int = 100,
+    include_anchor: bool = False,
+) -> AnnData:
+    """
+    Cluster variables by previously calculated distance to an anchor point.
+
+    Parameters
+    ----------
+    %(adata)s
+    cluster_key
+        Annotation column in `.obs` that is used as anchor.
+    design_matrix_key
+        Name of the design matrix saved to `.obsm`.
+    n_bins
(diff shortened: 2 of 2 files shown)


giovp:
hi @LLehner , thank you for this, would you mind elaborating a bit when this would be used? also, what if the embedding are pre-calculated, or the user would like to use something other than the UMAP, should that be an option? finally, I think a test would be required before we get this in, thanks!

timtreis:
Hey @giovp, this feature was coming out of a discussion with @maiiashulman. We ran into a situation in which the "literature-curated" signature for hypoxia was either 20 or 4000 genes, the latter obviously being useless. So we wondered which other genes maybe show the same spatially variable pattern as a function of distance to a certain cell-type (e.g. epithelial). This is essentially a graphical method to see if a given set of genes (f.e. the 20 gene signature) even varies in a similar pattern. 

But I agree with your points; if we see that it's actually doing something useful, we should make it a bit more flexible.

LLehner:
@timtreis this function now returns an `anndata` object, which is i think simplifies further processing, compared to storing the new count matrix somewhere in `.varm` or `.uns`. Because if we want to make us of already implemented dimreduction and clustering methods, then the count matrix needs to be in `.X` and for visualization we need the variable names stored as categories in `.obs`. Doing all of this in the same `anndata` will just make things cluttered.

Additionally the question is whether a `spatialdata` object should be required as input instead of an `anndata`one, because then a new table could be added directly instead of having multiple disconnected tables.

The function call would change from:
`adata_new = sq.tl.var_embeddings(adata, group="Endothelium", design_matrix_key="design_matrix")`
to
`sq.tl.var_embeddings(sdata, group="Endothelium", design_matrix_key="design_matrix")`

PR #831, opened 2024-05-27
Add function for calculating niches

LLehner:
## Description

Adds a function that calculates niches using different strategies. The initial function calculates niches based on neighborhood profiles similar to [here](https://github.com/immunitastx/monkeybread/blob/main/src/monkeybread/calc/_neighborhood_profile.py).

This PR will get updated with methods discussed in https://github.com/scverse/squidpy/issues/789.


giovp:
hi @LLehner I was just wondering what's the status of this? I see a lot of conflicts and tests still missing. No rush, just curious if there is a timeline

LLehner:
@giovp I want to wrap it up this month. Tests will be added and if possible 1-2 more methods and the previous ones compared to their original implementations.

timtreis:
Hey @LLehner 

> How should multiple slides be dealt with? All these methods should work with multiple slides, but i'm still not sure how the adjacency matrix changes when you run sq.gr.spatial_neighbors() on data with multiple slides. Is it a block diagonal matrix where each block on the diagonal is the adjacency matrix for a single slide? If yes, does it suffice to calculate the neighborhood graph once for all data or should it be done on individual slides? I noticed this can change niche results.

You mean a scenario where the user is just storing multiple slides in the same sdata object, like the `point8`, `point16`, ´point24` dataset we have? These should be fully independent since they have no biological connection.

> I think some parts could be sped up by parallelization (e.g. clustering with multiple resolutions), what would you recommend there?

Do you have some rough numbers? We could give the user the option to define `n_cores` or sth and then use the parallel processing we already have, but we should definitely keep the option to run single-threaded so that it can be used inside tools like snakemake that take care of job distribution across cores. Otherwise, that'd cause errors the user cannot circumvent.

>  If you run sq.calculate_niche() more than one time with flavor="neighborhood" i noticed that the first function call takes as much time as you would expect but for subsequent runs it's much (~10x) faster (e.g. first you run the method with one cluster resolution then call the method again with some other cluster resolutions). Its almost like the following function runs don't do neighborhood calculations (referring to sc.pp.neighbors here) anymore, but that shouldn't be the case, as nothing is cached and calculations happen on a new AnnData object for every function call. Also the "issue" persists even if you change the count matrix shape (by e.g. masking data).

Could it be that the OS has some of the compiled bytecode or data in cache? 🤔 

>  Would you include some form of logging or verbosity such that user sees what the method is currently doing? Depending on the data and settings, it the function can take a while.

What runtime are we speaking here about? I tend to be a fan of some intermediate output (if one can also shut it up, f.e. with some verbosity level)

src/squidpy/gr/_niche.py
```diff
@@ -0,0 +1,432 @@
...
+    abs_nhood: bool = False,
+    adj_subsets: int | list[int] | None = None,
+    aggregation: str = "mean",
+    n_components: int = 3,
+    random_state: int = 42,
+    spatial_connectivities_key: str = "spatial_connectivities",
+) -> AnnData | pd.DataFrame:
+    """Calculate niches (spatial clusters) based on a user-defined method in 'flavor'.
+    The resulting niche labels with be stored in 'adata.obs'. If flavor = 'all' then all available methods
+    will be applied and additionally compared using cluster validation scores.
```
giovp, line 44: space

src/squidpy/gr/_niche.py
```diff
@@ -0,0 +1,432 @@
...
+    Parameters
+    ----------
+    %(adata)s
+    flavor
+        Method to use for niche calculation. Available options are:
+            - `{c.NEIGHBORHOOD.s!r}` - cluster the neighborhood profile.
```
giovp, line 50: these should be defined in constants and added with inject docs, see other docs on how to do it

src/squidpy/gr/_niche.py
```diff
@@ -0,0 +1,432 @@
...
+            - `{c.UTAG.s!r}` - use utag algorithm (matrix multiplication).
+            - `{c.CELLCHARTER.s!r}` - cluster adjacency matrix with Gaussian Mixture Model (GMM) using CellCharter's approach.
+            - `{c.SPOT.s!r}` - calculate niches using optimal transport. (coming soon)
+            - `{c.BANKSY.s!r}`- use Banksy algorithm. (coming soon)
+    %(library_key)s
```
giovp, line 55: this won' twork without injecting docs, see the read the docs build to see how it renders

giovp: but the function is not added in the docstrings so it is not rendering it at allk

src/squidpy/gr/_niche.py
```diff
@@ -0,0 +1,432 @@
...
+    subset
+        Restrict niche calculation to a subset of the data.
+    table_key
+        Key in `spatialdata.tables` to specify an 'anndata' table. Only necessary if 'sdata' is passed.
+    mask
```
giovp, line 60: what's the use case for this?

src/squidpy/gr/_niche.py
```diff
@@ -0,0 +1,432 @@
...
+        Boolean array to filter cells which won't get assigned to a niche.
+        Note that if you want to exclude these cells during neighborhood calculation already, you should subset your AnnData table before running 'sq.gr.spatial_neigbors'.
+    groups
+        Groups based on which to calculate neighborhood profile (E.g. columns of cell type annotations in adata.obs).
+        Required if flavor == 'neighborhood'.
+    n_neighbors
+        Number of neighbors to use for 'scanpy.pp.neighbors' before clustering using leiden algorithm.
+        Required if flavor == 'neighborhood' or flavor == 'UTAG'.
```
giovp, line 68: you can use the constants also here

src/squidpy/gr/_niche.py
```diff
@@ -0,0 +1,432 @@
...
+    resolutions
+        List of resolutions to use for leiden clustering.
+        Required if flavor == 'neighborhood' or flavor == 'UTAG'.
+    subset_groups
+        Groups (e.g. cell type categories) to ignore when calculating the neighborhood profile.
+        Optional if flavor == 'neighborhood'.
+    min_niche_size
+        Minimum required size of a niche. Niches with fewer cells will be labeled as 'not_a_niche'.
```
giovp, line 76: mmh don't like this deafult, what about NA or None?

src/squidpy/gr/_niche.py
```diff
@@ -0,0 +1,432 @@
...
+            adata = adata.tables[table_key].copy()
+        else:
+            raise ValueError("Please specify which table to use with `table_key`.")
+    else:
+        adata = adata
+
+    if flavor == "neighborhood":
+        """adapted from https://github.com/immunitastx/monkeybread/blob/main/src/monkeybread/calc/_neighborhood_profile.py"""
+
+        # calculate the neighborhood profile for each cell (relative and absolute proportion of e.g. each cell type in the neighborhood)
```
giovp, line 112: please refactor with separate functions

src/squidpy/gr/_niche.py
```diff
@@ -0,0 +1,432 @@
...
+            adata_neighborhood = adata_neighborhood[mask]
+
+        # required for leiden clustering (note: no dim reduction performed in original implementation)
+        sc.pp.neighbors(adata_neighborhood, n_neighbors=n_neighbors, use_rep="X")
+
+        if resolutions is not None:
+            if not isinstance(resolutions, list):
+                resolutions = [resolutions]
+        else:
+            raise ValueError("Please provide resolutions for leiden clustering.")
```
giovp, line 138: this could be `resolutions = [resolutions] if not isinstance(resolutions, list) else resolutions` but then you'd also have to validate the args for this specific niche flavour. Using spearate functions, these checks could be done before

src/squidpy/gr/_niche.py
```diff
@@ -0,0 +1,432 @@
...
+    np.add.at(abs_freq, (np.arange(m)[:, None], cat_values), 1)
+
+    # normalize by n_neighbors to get relative frequency of each category
+    rel_freq = abs_freq / k
+
+    return pd.DataFrame(rel_freq, index=adata.obs.index), pd.DataFrame(abs_freq, index=adata.obs.index)
+
+
+def _utag(adata: AnnData, normalize_adj: bool, spatial_connectivity_key: str) -> AnnData:
+    """Performs inner product of adjacency matrix and feature matrix,
```
giovp, line 266: docs formatting, please do
```
"""
...
"""
```

src/squidpy/gr/_niche.py
```diff
@@ -0,0 +1,432 @@
...
+    ----------
+        distribution: An array of probabilities (should sum to one)
+
+    Returns
+        The Shannon entropy
+    """
+    return float(-(distribution * np.log(distribution + 1e-8)).sum())
+
+
+def _iter_uid(
```
giovp, line 416: please add docstring to this function

src/squidpy/gr/_niche.py
```diff
@@ -0,0 +1,432 @@
...
+    """Shannon entropy
+
+    Parameters
+    ----------
+        distribution: An array of probabilities (should sum to one)
+
+    Returns
+        The Shannon entropy
+    """
+    return float(-(distribution * np.log(distribution + 1e-8)).sum())
```
giovp, line 413: I think both numpy and scipy implement entropy, why is this required?

src/squidpy/gr/_niche.py
```diff
@@ -0,0 +1,432 @@
...
+
+    if n_classes is None:
+        return float(f1_scores.mean())
+
+    assert n_classes >= len(f1_scores), f"Expected {n_classes:=}, but found {len(f1_scores)}, which is greater"
+
+    return float(np.pad(f1_scores, (0, n_classes - len(f1_scores))).mean())
+
+
+def jensen_shannon_divergence(adatas: AnnData | list[AnnData], library_key: str, slide_key: str | None = None) -> float:
```
giovp, line 376: I think scipy has implemention for this as well.
https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.jensenshannon.html

giovp (changes requested):
that looks great @LLehner , I think it would benefit from a refactoring to take out the single implementation, validate arguments, and then run the functions. We discussed about potentially even doing separate modules, e.g.
```
squidpy.gr.niches.utag
...
```
but I'm not sure it is necessary, what do you and @timtreis think?
Keeping it like this would be also fine, but then the arguments should be handled a bit better (e.g. check whether the passed arguments are fit for the single implementation/flavour required, and otherwise fail right away). Also how each argument maps to each implementation could be documented in the docs.

Also some work to be done in the docs more generally.

giovp:
also on this question

> How should multiple slides be dealt with? All these methods should work with multiple slides, but i'm still not sure how the adjacency matrix changes when you run sq.gr.spatial_neighbors() on data with multiple slides. Is it a block diagonal matrix where each block on the diagonal is the adjacency matrix for a single slide? If yes, does it suffice to calculate the neighborhood graph once for all data or should it be done on individual slides? I noticed this can change niche results.

yes, and yes. If neighbor calculation is run with multiple slides, then a block diagonal matrix of the spatial neighbor is returned, in fact treating the slides to be independent.

> >I think some parts could be sped up by parallelization (e.g. clustering with multiple resolutions), what would you recommend there?

> Do you have some rough numbers? We could give the user the option to define n_cores or sth and then use the parallel processing we already have, but we should definitely keep the option to run single-threaded so that it can be used inside tools like snakemake that take care of job distribution across cores. Otherwise, that'd cause errors the user cannot circumvent.

yes agree! You can take a look at the `Parallel` module we have now and just reuse it

> Would you include some form of logging or verbosity such that user sees what the method is currently doing? Depending on the data and settings, it the function can take a while.

absolutely please log everything you see fit!

src/squidpy/gr/_niche.py
LLehner: If you want to make use of all existing connectivities but still want to ignore some cells during niche assignment. It's based on the 'mask' parameter here: https://monkeybread.readthedocs.io/en/latest/generated/monkeybread.calc.cellular_niches.html#monkeybread.calc.cellular_niches. The difference to subsetting is: if you subset you also cut the connectivities. With a mask you keep the connectivities but avoid labeling the masked cells.

src/squidpy/gr/_niche.py
```diff
@@ -137,17 +126,199 @@ def calculate_niche(
         aggregation,
         n_components,
         random_state,
-        spatial_connectivities_key,
+        inplace,
     )
 
+    if resolutions is None:
```
LLehner, line 133: Why not set this as default right away?

src/squidpy/gr/_niche.py
```diff
@@ -137,17 +126,199 @@ def calculate_niche(
...
+        resolutions = [0.5]
+
+    if distance is None:
```
LLehner, line 136: this too

src/squidpy/gr/_niche.py
```diff
@@ -137,17 +126,199 @@ def calculate_niche(
...
+    library_str = f"_{library_key}" if library_key is not None else ""
+
+    if flavor == "cellcharter":
+        base_column = "cellcharter_niche"
+        if library_key is None:
+            return [base_column]
+        elif libraries is not None and len(libraries) > 0:
+            return [f"{base_column}_{lib}" for lib in libraries]
+
+    # For neighborhood and utag, we need to handle resolutions
```
LLehner, line 268: Would do this with if-else block maybe

src/squidpy/gr/_niche.py
```diff
@@ -137,17 +126,199 @@ def calculate_niche(
...
 
 def _get_nhood_profile_niches(
     adata: AnnData,
     mask: pd.core.series.Series | None,
-    groups: str,
-    n_neighbors: int,
+    groups: str | None,
+    n_neighbors: int | None,
     resolutions: float | list[float],
-    subset_groups: list[str] | None,
```
LLehner: why remove this completely? It serves a different purpose than masking. This was also described in the original tutorial. E.g. a mask allows to restrict niche calculation to specific cell types but they can have neighbors outside the masked region. Subsetting means only observations within that subset are considered as neighbors.

src/squidpy/gr/_niche.py
```diff
@@ -209,35 +372,37 @@ def _get_nhood_profile_niches(
...
+            adata.obs[f"nhood_niche_res={res}"] = adata.obs.index.map(
+                adata_masked.obs[f"nhood_niche_res={res}"]
             ).fillna("not_a_niche")
 
     return
 
 
 def _get_utag_niches(
     adata: AnnData,
-    subset_groups: list[str] | None,
```
LLehner: Here subsetting serves the same purpose as before.

src/squidpy/gr/_niche.py
```diff
@@ -461,73 +627,206 @@ def _jensen_shannon_divergence(adata: AnnData, niche_key: str, library_key: str)
...
+                raise ValueError(f"'library_key' must be a column in 'adata.obs', got {library_key}")
+        elif isinstance(data, SpatialData):
+            if table_key is None:
+                raise ValueError("'table_key' is required when 'data' is a SpatialData object")
+            if table_key not in data.tables:
+                raise ValueError(f"'table_key' must be a valid table key in 'data', got {table_key}")
+            if library_key not in data.tables[table_key].obs.columns:
+                raise ValueError(f"'library_key' must be a column in 'adata.obs', got {library_key}")
+
+    if n_neighbors is not None and not isinstance(n_neighbors, int):
```
LLehner, line 717: Why check these separately?

src/squidpy/gr/_niche.py
```diff
@@ -461,73 +627,206 @@ def _jensen_shannon_divergence(adata: AnnData, niche_key: str, library_key: str)
...
+        Dictionary of parameter names to their values
+    param_specs
+        Dictionary with 'required', 'optional', and 'unused' parameter lists for the flavor
+    """
+    unnecessary_args = []
+
+    for param_name in param_specs["unused"]:
+        param_value = param_dict.get(param_name)
+
+        # Special handling for boolean parameters with default values
```
LLehner, line 857: is this necessary?

PR #859, opened 2024-07-25
Add function to assign observations to a sliding window

### LINKED ISSUE #829: Sliding window across spatial neighbourhood
## Description

Implementation of sliding window (inspired by [Kasumi](https://www.biorxiv.org/content/10.1101/2024.03.06.583691v1)) over a spatial neighbourhood or view. 

Input:
- window size
- overlap

Output:
- assignment of cells to sliding window

FrancescaDr:
## Description

This PR adds a new function, sliding_window, to the codebase. The sliding_window function divides a tissue slice into regularly shaped, spatially contiguous regions (windows). It allows for the analysis of spatial data by segmenting the tissue into smaller, manageable windows with optional overlap.

## How has this been tested?

PyTest on `sq.datasets.mibitof()`

## Closes

Closes #829


src/squidpy/tl/_sliding_window.py
```diff
@@ -0,0 +1,129 @@
...
+from squidpy._docs import d
+from squidpy._utils import NDArrayA
+from squidpy.gr._utils import _save_data
+
+__all__ = ["sliding_window"]
+
+
+@d.dedent
+def sliding_window(
+    adata: AnnData,
```
timtreis, line 21: we're moving `squidpy` to be fully `spatialdata` compatible, so please add that here as well. Super simple, check: https://github.com/scverse/squidpy/blob/e6b5b0ccd2739aa50b55c7628e685a6750249d8f/src/squidpy/gr/_nhood.py#L159

src/squidpy/tl/_sliding_window.py
```diff
@@ -0,0 +1,129 @@
...
+    sliding_window_key: str = "sliding_window",
+    spatial_key: str = "spatial",
+    copy: bool = False,
+) -> AnnData:
+    """
+    Divide a tissue slice into regulary shaped spatially contiguous regions (windows).
+
+    Parameters
+        ----------
+        adata: AnnData
```
timtreis, line 35: Please adjust doc style to rest of package

src/squidpy/tl/_sliding_window.py
```diff
@@ -0,0 +1,129 @@
...
+def sliding_window(
+    adata: AnnData,
+    library_key: str | None = None,
+    window_size: int = 2500,
+    overlap: int = 0,
+    coord_columns: tuple[str, str] = ("globalX", "globalY"),
+    sliding_window_key: str = "sliding_window",
+    spatial_key: str = "spatial",
+    copy: bool = False,
+) -> AnnData:
```
timtreis, line 29: You annotate that you return an AnnData, but you either modify in place (None) or return a pandas df

src/squidpy/tl/_sliding_window.py
```diff
@@ -0,0 +1,129 @@
...
+            Base name for sliding window columns.
+        overlap: int
+            Overlap size between consecutive windows.
+        %(spatial_key)s
+        copy: bool
+            Whether to return a copy of the AnnData object.
+
+        Returns
+        -------
+        If ``copy = True``, returns the design_matrix
```
timtreis, line 53: which `design_matrix`?

src/squidpy/tl/_sliding_window.py
```diff
@@ -0,0 +1,129 @@
...
+    copy: bool = False,
+) -> AnnData:
+    """
+    Divide a tissue slice into regulary shaped spatially contiguous regions (windows).
+
+    Parameters
+        ----------
+        adata: AnnData
+            Annotated data matrix.
+        window_size: int
```
timtreis, line 37: What if `window_size` is 0 or negative? Add error handling

src/squidpy/tl/_sliding_window.py
```diff
@@ -0,0 +1,129 @@
...
+            Size of the sliding window.
+        library_key: str
+            Only cells with the same library_key in `adata.obs` are assigned to the same window
+        coord_columns: Tuple[str, str]
+            Tuple of column names in `adata.obs` that specify the coordinates (x, y), e.i. ('globalX', 'globalY')
+        sliding_window_key: str
+            Base name for sliding window columns.
+        overlap: int
```
timtreis, line 45: What if `overlap` is negative? Is this intended to have spaced out sections? Also errorhandling

src/squidpy/tl/_sliding_window.py
```diff
@@ -0,0 +1,129 @@
...
+            )
+        else:
+            for i in range(num_grid_systems):
+                grid_col_name = f"{sliding_window_key}_{i+1}"
+                row, col = (i // grid_len), (i % grid_len)
+                x_grid_start = min_x + (overlap * col)
+                y_grid_start = min_y + (overlap * row)
+                xgrid, ygrid = (lib_data[x] - x_grid_start) // window_size, (lib_data[y] - y_grid_start) // window_size
+                xrows = xgrid.max()
+                sliding_window_df.loc[lib_mask, grid_col_name] = (
```
timtreis, line 106: This can be super inefficient for large data, f.e. Xenium. Can you vectorise?

src/squidpy/tl/_sliding_window.py
```diff
@@ -0,0 +1,129 @@
...
+                    str(lib) + "_" + f"{1+i}_" + ((ygrid * (xrows + 1)) + xgrid).astype(int).astype(str)
+                ).astype("category")
+
+    if copy:
+        logg.info("Finish", time=start)
+        return sliding_window_df
+    else:
```
timtreis, line 113: `else` not needed here. Your `if` condiiton returns early, so your `else` case is everything afterwards

src/squidpy/tl/_sliding_window.py
```diff
@@ -0,0 +1,129 @@
...
+        num_grid_systems = grid_len**2
+        for i in range(num_grid_systems):
+            sliding_window_df[f"{sliding_window_key}_{i+1}"] = np.nan
+
+    for lib in libraries:
+        if library_key is not None:
+            lib_mask = adata.obs[library_key] == lib
+            lib_data = adata.obs[lib_mask]
+
+        min_x, min_y = lib_data[x].min(), lib_data[y].min()
```
timtreis, line 90: What if the user accidentally converted their coords to str? Error handling, maybe warn and convert

src/squidpy/tl/_sliding_window.py
```diff
@@ -0,0 +1,129 @@
...
+                x_grid_start = min_x + (overlap * col)
+                y_grid_start = min_y + (overlap * row)
+                xgrid, ygrid = (lib_data[x] - x_grid_start) // window_size, (lib_data[y] - y_grid_start) // window_size
+                xrows = xgrid.max()
+                sliding_window_df.loc[lib_mask, grid_col_name] = (
+                    str(lib) + "_" + f"{1+i}_" + ((ygrid * (xrows + 1)) + xgrid).astype(int).astype(str)
+                ).astype("category")
+
+    if copy:
+        logg.info("Finish", time=start)
```
timtreis, line 111: Did you mean to log some end time here?

tests/tools/test_sliding_window.py
```diff
@@ -0,0 +1,51 @@
...
+    @pytest.mark.parametrize("overlap", [0, 2])
+    def test_design_matrix_several_slices(
+        self,
+        adata_mibitof: AnnData,
+        library_key: str,
+        window_size: int,
+        overlap: int,
+    ):
+        df = sliding_window(
+            adata_mibitof,
```
timtreis, line 23: Would be nicer to evaluate on toy anndata f.e. with a 4x4 point grid, then you can explicitly evaluate on that. That actually also makes me think of the edge case when a cell is right on the border - does your function handle that?

timtreis (changes requested):

src/squidpy/tl/_sliding_window.py
FrancescaDr: I saw this was also used in the _var_by_distance.py tool. [https://github.com/scverse/squidpy/blob/main/src/squidpy/tl/_var_by_distance.py
```suggestion
        logg.info("Finish", time=start)
```](https://github.com/scverse/squidpy/blob/e6b5b0ccd2739aa50b55c7628e685a6750249d8f/src/squidpy/tl/_var_by_distance.py#L171)

File: src/squidpy/tl/_sliding_window.py (hunk 6 above)
Comment (FrancescaDr) [reply src/squidpy/tl/_sliding_window.py:45]: What do you mean by spaced out sections? Sections that are not completely equally sized? Then yes

File: src/squidpy/tl/_sliding_window.py (hunk 10 above)
Comment (timtreis) [reply src/squidpy/tl/_sliding_window.py:111]: @LLehner is this a typo?

Comment (LLehner) [reply src/squidpy/tl/_sliding_window.py:111]: Probably a leftover from an early draft of the method. Can be removed.

Comment (timtreis) [reply src/squidpy/tl/_sliding_window.py:111]: The entire logging or just that you log the start time as the end time?

Comment (LLehner) [reply src/squidpy/tl/_sliding_window.py:111]: In _var_by_distance.py both logging statements can be removed.

Comment (FrancescaDr) [reply src/squidpy/tl/_sliding_window.py:111]: I will remove it then

File: tests/tools/test_sliding_window.py [hunk 12]
```diff
@@ -0,0 +1,51 @@
...
+            grid_len = window_size // overlap
+            num_grid_systems = grid_len**2
+            for i in range(num_grid_systems):
+                assert f"sliding_window_{i+1}" in df.columns  # correct number of columns; multiple sliding windows
+
+            for col in df.columns:
+                lib_windows = []
+                for lib in adata_mibitof.obs[library_key].unique():
+                    lib_windows.append(set(df[adata_mibitof.obs[library_key] == lib][col]))
+                    print(lib_windows)
```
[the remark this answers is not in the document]
FrancescaDr: I removed the print statement, the rest is needed to check that the sliding windows are created per library key and that there is no overlap in nodes that are included across the sliding windows

src/squidpy/tl/_sliding_window.py
```diff
@@ -0,0 +1,134 @@
...
+    if overlap < 0:
+        raise ValueError("Overlap must be non-negative.")
+
+    if isinstance(adata, SpatialData):
+        adata = adata.table
+
+    x, y = coord_columns
+
+    if x not in adata.obs or y not in adata.obs:
+        adata.obs[x] = adata.obsm[spatial_key][:, 0]
```
LLehner, line 66: Is it really necessary to copy coordinates as new columns to .obs? Should be possible to work directly with .obsm

LLehner:
Hi @FrancescaDr, i just tested your code on data with a single slide (so I didn't provide a library_key) and it seems like the sliding windows are only created when you provide a library key.

E.g. the following code does run but doesn't return multiple sliding windows:

`adata = sq.datasets.seqfish()`
`sq.tl.sliding_window(adata, window_size = x) # here x can be any integer, doesn't change the result`

src/squidpy/tl/_sliding_window.py
timtreis: Sorry, should've been clearer. If I were to set overlap to -10, I'd expect my windows to be 10 units apart each, instead of overlapping, right? Do we allow for this?

FrancescaDr:
> Hi @FrancescaDr, i just tested your code on data with a single slide (so I didn't provide a library_key) and it seems like the sliding windows are only created when you provide a library key.
> 
> E.g. the following code does run but doesn't return multiple sliding windows:
> 
> `adata = sq.datasets.seqfish()` `sq.tl.sliding_window(adata, window_size = x) # here x can be any integer, doesn't change the result`

If I choose x = 1,2,3 I get a adata.obs['sliding_window'] with 'None_Z' where Z is the number of the sliding window.

src/squidpy/tl/_sliding_window.py
FrancescaDr: I thought this would be the easiest way to ensure that the user could indicate different x or y variables and after running the sliding window have the x and y coordinates that were used available as reference.

LLehner:
> If I choose x = 1,2,3 I get a adata.obs['sliding_window'] with 'None_Z' where Z is the number of the sliding window.

Ah right, for this dataset the coordinates are on a small scale, i initially tried it with window size 10, 100, 1000...ok i see that it works now!

FrancescaDr:
> Ah right, for this dataset the coordinates are on a small scale, i initially tried it with window size 10, 100, 1000...ok i see that it works now!

Yes the scale sliding window size variable always depends on the density and size of the FOV. It might be valuable to add a function that suggest appropriate sliding window sizes depending on e.i. number of average cells in a window.

timtreis (approved):
Made a few things more robust. Code shown in tutorial notebook works as expected

PR #869, opened 2024-08-08
Fix links to references

LLehner:
Fix broken links.


LLehner:
No longer necessary since lint passes again.

PR #877, opened 2024-08-23
remove napari from install

giovp:
closes https://github.com/scverse/squidpy/issues/860 

todo:
- [x] add warning in changelog and readme
- [x] add warning in notebook https://github.com/scverse/squidpy_notebooks/pull/130


giovp:
This is what's in the landing page
image

similar thing is in the notebook but unfort it doesn't seem like it's rendering here, will first merge to main and then check from the build there. After that, will cut new release

LLehner (approved):
Looks good to me!

PR #949, opened 2025-01-30
Added function to download Visium as SpatialData

timtreis:
The `squidpy.datasets` function were downloading AnnDatas or ImageContainers. For a lot of work, I'd love to have an easy SpatialData download. So here it is. Data was created as described here: https://github.com/scverse/spatialdata-notebooks/pull/131

Edit: I had to write some quite hacky code around the download function to also have it be automatically downloaded during CI/CD. Theoretically, that entire part should be refactored because it's (unnecessarily complicated)**3, but time yada yada

### DIFF
--- a/src/squidpy/datasets/_10x_datasets.py
+++ b/src/squidpy/datasets/_10x_datasets.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import os
+import shutil
 import tarfile
 from pathlib import Path
 from typing import (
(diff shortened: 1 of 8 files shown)


timtreis:
converting to draft until issues with https://github.com/scverse/spatialdata-notebooks/pull/131 fixed

src/squidpy/datasets/_10x_datasets.py
```diff
@@ -134,3 +142,54 @@ def visium(
         )
 
     return read_visium(base_dir / sample_id)
+
+
+def visium_hne_sdata(folderpath: Path | str) -> sd.SpatialData:
```
LucaMarconato, line 147: Good PR! The only comment I have is here: this function is not really Visium specific and replicating this for other techs would lead to code redundancy. I would turn this function into a more general one and pass the Visium-specific strings as argument.

LucaMarconato (approved):

src/squidpy/datasets/_10x_datasets.py
timtreis: Refactored it 👍

PR #967, opened 2025-03-04
Add filter function that subsets the entire object

timtreis:
Users can easily break `SpatialData` objects by filtering the contained `AnnData` tables using scanpy - this might remove observations that are still present in other elements, such as `Shapes`. This function properly filters the entire object, enabled things such as shown here:

 
image


LucaMarconato:
Hi thanks. Can you show an example of error that occurs after the legacy filtering?

PR #975, opened 2025-03-19
Re-implementating co_occurrence()

### LINKED ISSUE #755: sq.gr.co_occurrence gives different results for n_splits=1 and n_splits>1
## Description
Calculating co-occurrences I noticed some inf and nan values in the outcome and while looking into it I saw Issue #689 and tried to reduce n_splits to 1. Not only did the inf and nan values go away, but all the other values changed significantly. 

I would very much appreciate a fix because the function n_splits=1 produces an error for very large samples.
Thank you :)

## Version
1.3.0

wenjie1991:

## Description

Togeter with [@MDLDan](https://github.com/MDLDan) we reimplement the `squidpy.gr.co_occurrence()` function using Numba. 
The new algorithm removes the need for a pre-calculated pairwise distance matrix, enabling it to handle large datasets without splitting. Parallel processing is enabled by default, increasing the runtime speed by 40 times.
image

We also implemented it in Rust using PyO3 and achieved similar performance. We chose to push the Numba implementation.

Following issues are related:
https://github.com/scverse/squidpy/issues/229
https://github.com/scverse/squidpy/issues/755
https://github.com/scverse/squidpy/issues/223
https://github.com/scverse/squidpy/issues/582

## How has this been tested?

- All `squidpy.gr.co_occurrence()` related have passed in `squidpy` package test. 
- We also compared the new and old implementations output:
   - Until the number of cells do not require squidpy to split the differences are in the 1e-08 range.
   - When the number of cells requires squidpy to spit differences are in the order of 1e-02 see ([
](https://github.com/scverse/squidpy/issues/755).

## Closes

closes [#755](https://github.com/scverse/squidpy/issues/755)


Intron7:
@wenjie1991 This already looks very good and promising. But I believe you can squeeze out even more performance. You can start by adjusting the memory access pattern to be efficent. You can also numba_njit the outer function and parallelize it. Also i would cache the kernel that makes it even more efficent.

wenjie1991:
@MDLDan

Intron7:
@wenjie1991 I'll take a look later next week. Thank you for optimising this a bit more.

selmanozleyen:
I am writing here to document the current performances here is the results with the current branch. Peakmem and time respectively.
```
========= ====== ==========
--              layer      
--------- -----------------
dataset   None   off-axis 
========= ====== ==========
  imc     628M     649M   
========= ====== ==========


========= ========= ==========
--               layer        
--------- --------------------
dataset     None    off-axis 
========= ========= ==========
  imc     305±0ms   306±0ms  
========= ========= ==========
```

And with main:
```
========= ======= ==========
 --              layer       
 --------- ------------------
  dataset    None   off-axis 
 ========= ======= ==========
    imc     1.65G    1.46G   
 ========= ======= ==========

========= ========= ==========
--               layer        
--------- --------------------
dataset     None    off-axis 
========= ========= ==========
  imc     12.2±0s   12.2±0s  
========= ========= ==========
```

src/squidpy/gr/_ppatterns.py
```diff
@@ -266,85 +267,120 @@ def _score_helper(
...
-                    probs_con[c, i] = 0
-                else:
-                    probs_con[c, i] = probs_conditional[i] / probs[i]
+    # Allocate a 2D array to store a flat local result per point.
+    k2 = k * k
+    local_results = np.zeros((n, l_val * k2), dtype=np.int32)
 
-        out[:, :, idx] = probs_con
+    for i in prange(n):
+        local_counts: NDArrayA = np.zeros(l_val * k2, dtype=np.int32)
```
Intron7: I might be blanking here but do we need this can't we write into `local_results` directly?

selmanozleyen: we can. I also noticed this

Intron7: I would still change this to be more memory efficient

wenjie1991: I agree, the original purpose of using `local_counts` was to prevent "write-collision," but this issue shouldn't occur. I believe @selmanozleyen's commit has addressed it.

selmanozleyen: @wenjie1991 I can't push to this branch so I'd appreciate if you could apply that commit yourself please

wenjie1991: @selmanozleyen Your commit has been merged.

Intron7 (approved):

selmanozleyen (approved):

HXD3D0323:
Dear  Authors @wenjie1991 @codecov-commenter @Intron7 @selmanozleyen ,
    Thank you for developing and sharing this method. While reading the implementation of the co-occurrence calculation, I am trying to better understand the normalization strategy , particularly why the denominator is defined as row_sums[c, r] rather than the total number of neighbors surrounding center cells of type c used in the following code in src/squidpy/gr/_ppatterns.py :
`occ_prob = np.zeros((k, k, l_val), dtype=np.float64)

row_sums = counts.sum(axis=0)
totals = row_sums.sum(axis=0)

for r in prange(l_val):
    probs = row_sums[:, r] / totals[r]

    for c in range(k):
        for i in range(k):
            if probs[i] != 0.0 and row_sums[c, r] != 0.0:
                occ_prob[i, c, r] = (
                    counts[c, i, r] / row_sums[c, r]
                ) / probs[i]`
From the code, counts[c, i, r] appears to represent the number of neighboring pairs within radius r, where the center cell is of type c and the neighboring cell is of type i.

The normalization term is defined as`row_sums = counts.sum(axis=0)`
which yields$$
\mathrm{row\_sums}[i,r]
=
\sum_c \mathrm{counts}[c,i,r].
$$

Therefore, `row_sums[c, r]` corresponds to

$$
\sum_{c'} \mathrm{counts}[c',c,r],
$$

which seems to represent the total number of times cell type $c$ appears as a neighboring cell across all center-cell types at radius $r$.

My question concerns the denominator

$$
\frac{\mathrm{counts}[c,i,r]}
     {\mathrm{row\_sums}[c,r]}.
$$

If the goal is to estimate a conditional probability such as

$$
P(\mathrm{neighbor}=i \mid \mathrm{center}=c),
$$

I would have expected the denominator to be

$$
\sum_i \mathrm{counts}[c,i,r],
$$

i.e., the total number of neighbors observed around center cells of type $c$, since this corresponds directly to conditioning on the center-cell type.

In contrast, the current implementation uses

$$
\sum_{c'} \mathrm{counts}[c',c,r],
$$

which appears to normalize by the frequency with which cell type $c$ occurs as a neighboring cell rather than as a center cell.

I would greatly appreciate any explanation of the statistical reasoning behind this normalization strategy.

Thank you very much for your time and for making the implementation publicly available.

PR #978, opened 2025-03-24
Change niche flavor to cellcharter_simple and default distance = 3

### LINKED ISSUE #977: Change CellCharter name and default values
Thank you very much for implementing CellCharter!
As I discussed in the past, @LLehner had to implement a simplified version of CellCharter to avoid adding new dependences.
The most important one is replacing scVI's dimensionality reduction with PCA.

In my experience developing CellCharter I noticed that using PCA can considerably decrease the quality of the niches, especially in spatial transcriptomics datasets, and is not as scalable for large datasets.
For this reason I would like to propose to change the name of the `flavor` parameter in the `calculate_niche` function from `"cellcharter"` to `"cellcharter_simple"`.

**This is to discourage people from using the squidpy's implementation for benchmarking CellCharter against other methods.** 

In addition, assuming that people using squidpy's implementation will have less familiarity with CellCharter, I would propose using a more _good-enough_ `distance=3` as a default value instead of the current `distance=1` (the docs currently say that it's required but from the code it looks like 1 is set by default).

A PR is on the way!

Let me know what you think :)

marcovarrone:
## Description
- Change `flavor` option in `calculate_niche to `cellcharter_simple` for CellCharter
- Default `distance = 3` for `cellcharter_simple`. 

## How has this been tested?
`pytest` runs successfully. 
However, I see there are no unit tests for `calculate_niche` with `cellcharter` (now `cellcharter_simple`).

## Closes
Closes #977

### DIFF
--- a/src/squidpy/_constants/_constants.py
+++ b/src/squidpy/_constants/_constants.py
@@ -129,6 +129,6 @@ class TenxVersions(str, ModeEnum):
 class NicheDefinitions(ModeEnum):
     NEIGHBORHOOD = "neighborhood"
     UTAG = "utag"
-    CELLCHARTER = "cellcharter"
+    CELLCHARTER = "cellcharter_simple"
     SPOT = "spot"
     BANKSY = "banksy"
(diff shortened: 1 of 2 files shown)


timtreis:
Hey @marcovarrone, sorry for having let this gone stale. I'm not the biggest fan of changing the name but very much see your point - I've added the `use_rep` parameter and a warning in case the user now goes with the PCA approach - would this work for you?

PR #984, opened 2025-04-07
spatial_autocorr limit the numba threads as n_jobs temporarily

selmanozleyen:
## Description

I am not sure if this is a bug. But it makes sense for the user to expect numba to be using at most `n_jobs` on their cores. I made one solution like this one but I think any code that uses numba will have to be modified this way if we see this as a bug right @ilan-gold ? or am I missing something?

## Closes

https://github.com/scverse/squidpy/issues/957#event-17114743180

### DIFF
--- a/src/squidpy/gr/_ppatterns.py
+++ b/src/squidpy/gr/_ppatterns.py
@@ -10,7 +10,7 @@ import numba.types as nt
 import numpy as np
 import pandas as pd
 from anndata import AnnData
-from numba import njit
+from numba import get_num_threads, njit, set_num_threads
 from numpy.random import default_rng
 from scanpy import logging as logg
 from scanpy.get import _get_obs_rep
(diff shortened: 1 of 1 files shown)


selmanozleyen:
I assumed it is numba related it uses score_helper which uses moran's I which is implemented with numba in scanpy. https://github.com/scverse/squidpy/blob/afcb8d0e81085a1af03ae5a9f299c5df00e95d61/src/squidpy/gr/_ppatterns.py#L254
moran helper in scanpy:
https://github.com/scverse/scanpy/blob/15c5434ad0382614a16df612745c183807675d04/src/scanpy/metrics/_morans_i.py#L131

I checked locally with htop and this runs on all the cores without the changes I made
```python
import numpy as np
import pandas as pd

import anndata as ad
import scanpy as sc
import squidpy as sq



# load the pre-processed dataset

sq.gr.spatial_autocorr(adata, n_jobs=1, n_perms=10000000, mode="moran")
```

ilan-gold:
Right @selmanozleyen yes I got lost in the sauce.   I understand now better, I think.  So:

1. The `n_jobs` parameter is meant for `parallelize`, not `numba`
2. Separately `numba` has its own setting [the environment variable NUMBA_NUM_THREADS](https://numba.pydata.org/numba-doc/dev/user/threading-layer.html)
3. Setting the former does not interact with the later, so limiting `n_jobs` means `numba` may still max out your CPU (or similar behavior)

If so, then I think this issue is one of documentation, you're right.

timtreis:
Based on the original comment it seems that numba just grabs cores irregardless though. I think we should limit numba here to only act "within core" since as Selman said the user expects only `n_jobs` to be busy.

selmanozleyen:
I will close this as we decided this is a documentation issue

PR #987, opened 2025-04-09
Docs: Edit the argument documentation of `n_jobs` of the `parallelize` function

selmanozleyen:
Related: https://github.com/scverse/squidpy/issues/957

## Description

Following the discussion in https://github.com/scverse/squidpy/pull/984#issuecomment-2783410367 I modified the docs. I also wrote that this argument will be removed in the future based on our conversations but I am not sure about that.

If that is correct I will create an issue titled: `Reconsidering usage of paralellize` or something like that

### DIFF
--- a/src/squidpy/_docs.py
+++ b/src/squidpy/_docs.py
@@ -107,7 +107,8 @@ Nothing, just plots the figure and optionally saves the plot.
 """
 _parallelize = """\
 n_jobs
-    Number of parallel jobs.
+    Number of parallel jobs to use. If the function uses numba compiled functions, numba may
+    use cores depending on the number of threads set in the environment regardless of this argument.
 backend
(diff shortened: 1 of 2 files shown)


selmanozleyen:
@ilan-gold  I could send a future warning but are we going to deprecate something else if we are going to replace parallelize? I feel like we should throw warnings if we will surely replace it because many functions seem to use parallelize and we might be sending annoying warnings for a long time. Maybe I can also just remove the sentence where I say n_jobs will be removed in the future and just mention that it us to numba ultimately how many cores will be utilized.

selmanozleyen:
@ilan-gold I also don't understand, I think @Intron7  had an opinion on the `parallelize` utility function. If it's worth refactoring maybe we should consider but maybe we can merge this if thats still fine

PR #1002, opened 2025-05-21
Numba warning fix

selmanozleyen:
Fixes: https://github.com/scverse/squidpy/issues/994

Without the fix this is the warning I get in a notebook:

```
[/Users/selman.ozleyen/mambaforge/envs/squidpy/lib/python3.11/site-packages/numba/parfors/parfor.py:2395](https://file+.vscode-resource.vscode-cdn.net/Users/selman.ozleyen/mambaforge/envs/squidpy/lib/python3.11/site-packages/numba/parfors/parfor.py:2395): NumbaPerformanceWarning: 
prange or pndindex loop will not be executed in parallel due to there being more than one entry to or exit from the loop (e.g., an assertion).

File "/Users/selman.ozleyen/projects/squidpy/docs/notebooks/tutorials", line 31:
<source missing, REPL/exec in use?>
```

I don't think it is possible to add a test case because in theory that part of the code is unreachable

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -229,5 +229,5 @@ ban-relative-imports = "all"
 
 [tool.pytest.ini_options]
 filterwarnings = [
-    "error::NumbaPerformanceWarning"
+    "error::numba.NumbaPerformanceWarning"
 ]


ilan-gold (approved):

ilan-gold:
To catch this sort of thing, the project should use the `filterwarnings` option and tests run so that warnings are treated as errors as is done in scanpy/anndata.

ilan-gold:
So in this PR, you could also add `filterwarnings = [ "error:NumbaPerformanceWarning" ]` or do it locally and go one-by-one and catch any others that exist

ilan-gold:
I think it might need to be a double colon: https://docs.pytest.org/en/stable/how-to/capture-warnings.html#controlling-warnings sorry for the mistype

ilan-gold (approved):

PR #1006, opened 2025-05-26
Fix outline color behavior

### LINKED ISSUE #1005: Wrong outline colors
## Description

When I try to render with outlines, using `sq.pl.spatial_scatter()`, I get yellow and purple outlines instead of white and black ones. This is due to the fact that in `_spatial_utils.py` in `_shaped_scatter()`, the given rgba array is interpreted as "values that are mapped to colors" using the colormap and norm. Maybe this behavior was introduced in #903 so I'd suggest changing the color argument back to `c=outline_params.bg_color` instead of `c=colors.to_rgba_array(outline_params.bg_color)` which gives the desired behavior for me. Same for the `gap_color`.

**Note**: this is also visible in current test images! (e.g. `SpatialStatic_spatial_scatter_group_outline.png`). Still the tests pass, so I think the threshold for testing might not be strict enough here (as discussed @timtreis). Example:
Reference image
Image

Current output

Image

## Minimal reproducible example

```python
sq.pl.spatial_scatter(adata, shape="circle", color="cluster", outline=True, crop_coord=(1500, 1500, 3000, 3000))
```
Gives us 

Image

Even though the default outline colors are white and black.

## Version

1.6.5

Sonja-Stockhaus:
## Description

`bg_color` and `gap_color` of the outlines are not converted to RGBA arrays anymore before being passed to `_shaped_scatter()` since this would be interpreted as values to color the shapes by and lead to incorrect rendering.

## How has this been tested?
```python
sq.pl.spatial_scatter(adata, shape="circle", color="cluster", outline=True, crop_coord=(1500, 1500, 3000, 3000), groups="Cortex_1")
``` 
Before:
grafik
After:
grafik

## Closes

closes #1005


Sonja-Stockhaus:
This shows that the visual testing might have to be stricter

PR #1008, opened 2025-05-28
`parallelize` sets numba cores to 1

selmanozleyen:
Fixes https://github.com/scverse/squidpy/issues/957.

I test this with this script: https://gist.github.com/selmanozleyen/d130bca5bd1d549e96e6664d93c70fb6

I will create mock functions that utilize parallelize with different parameters (like `use_runner` arg of `parallelize`)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -232,3 +232,4 @@ ban-relative-imports = "all"
 filterwarnings = [
     "error::numba.NumbaPerformanceWarning"
 ]
+addopts = "-n auto"


tests/utils/test_parallelize.py
```diff
@@ -0,0 +1,104 @@
...
+    expected = [func(arr1[i], arr2) for i in range(len(arr1))]
+    # this will be set to something other than 1,2,8
+    # we want to check if setting the threads works
+    # then after the function is run if the numba cores are set back to 1
+    old_num_threads = 3
+    numba.set_num_threads(old_num_threads)
+    # Get initial state
+    initial_process = psutil.Process()
+    initial_children = {p.pid for p in initial_process.children(recursive=True)}
+    initial_children = {psutil.Process(pid) for pid in initial_children}
```
ilan-gold, line 81:
```python
initial_process = psutil.Process()
initial_children = {p.pid for p in initial_process.children(recursive=True)}
```
why would you want to make these `Process` objects, especially if they are set-diffed below with `pid` (which is a number)?

tests/utils/test_parallelize.py
```diff
@@ -0,0 +1,104 @@
...
+def test_parallelize_loky(func, n_jobs):
+    start_time = time.time()
+    seed = 42
+    rng = np.random.RandomState(seed)
+    n = 8
+    arr1 = [rng.randint(0, 100, n) for _ in range(n)]
+    arr2 = np.arange(n)
+    runner = partial(mock_runner, func=func)
+    # this is the expected result of the function
+    expected = [func(arr1[i], arr2) for i in range(len(arr1))]
```
ilan-gold, line 72:
```python
expected = [func(arr1[i], arr2) for i in arr1]
```

tests/utils/test_parallelize.py
```diff
@@ -0,0 +1,104 @@
...
+    final_children = {p.pid for p in initial_process.children(recursive=True)}
+    final_numba_threads = numba.get_num_threads()
+
+    assert init_numba_threads == old_num_threads, "Numba threads should not change"
+    assert final_numba_threads == 1, "Numba threads should be 1"
+    assert len(result) == len(expected), f"Expected: {expected} but got {result}. Length mismatch"
+    for i in range(len(arr1)):
+        assert np.all(result[i] == expected[i]), f"Expected {expected[i]} but got {result[i]}"
+
+    processes = final_children - initial_children
```
ilan-gold, line 96: Hmmm, but isn't `initial_children` full of `Process` objects and `final_children` just hte ids? How does this diff work?

tests/utils/test_parallelize.py
ilan-gold: Relatedly, why would the initial process have any children in any of the cases? Maybe add a comment? I can see it appears non-deterministic, but that should at the minimum be commented (if not the cause idscovered)

src/squidpy/_utils.py
```diff
@@ -55,6 +56,11 @@ def _unique_order_preserving(
     return [i for i in iterable if not (i in seen or seen_add(i))], seen
 
 
+def _callback_wrapper(chosen_runner: Callable[..., Any], *args: Any, **kwargs: Any) -> Any:
+    numba.set_num_threads(1)
```
ilan-gold, line 60: Wouldn't we want the number set back to the original number? I don't think we want to make this change globally.

pyproject.toml
```diff
@@ -231,3 +232,4 @@ ban-relative-imports = "all"
 filterwarnings = [
     "error::numba.NumbaPerformanceWarning"
 ]
+addopts = "-n auto"
```
ilan-gold, line 235: I don't think this is a good idea.  You should add this sort of thing to the CI instead, otherwise people will always run things in parallel even if it's just one test

src/squidpy/_utils.py
selmanozleyen: then we can set it again at the end of the wrapper function. But I thought we wanted to set it globally for good because I am not sure what the behavior would be when it is set to something >1 again (especially if another parallelize is running independently meanwhile)

tests/utils/test_parallelize.py
ilan-gold: I have a suspicion this is an `xdist` issue - I've run an individual test now 12 times and counting without failure where I added
```python
    assert len(initial_children) == 0
```
early in the test.

So I think we should look into ensuring all of these tests run in a single group (see slack), which could help with determinism.  And if we still can't get it working, it would _definitely_ be worth commenting (or investigating hooking into `xdist` and checking process numbers from its workers).  I think there is a way to check this via an api.  It's possible this issue of children processes only occurs on the master process

selmanozleyen: my bad I wanted to print them with detail before thats why it stayed that way. From what I observed the processes loky forks are always from the initial process. I will add more details to the tests then

ilan-gold: No I can reproduce without `xdist`, but there is always one child process for me, and also all of the tests have only one excpet `numba_parallel-1`, which passes.  I think there is an explanation here?

ilan-gold: And when I turn on `xdist`, it is also consistently `vanilla-{1,2,8}` that fail with the same number of subprocesses, as well as `dask-8`

src/squidpy/_utils.py
ilan-gold: > (especially if another parallelize is running independently meanwhile)

Is this possible?

> But I thought we wanted to set it globally for good because I am not sure what the behavior would be when it is set to something >1 again

Hmm that is a good point. Why could something bad happen?

tests/utils/test_parallelize.py
selmanozleyen: thats weird... I should work on making the tests more tight and deterministic then

ilan-gold: Prinitng out the process ids, I can see that when not run in parallel, the first test has no children i.e., `numba_parallel-1` and then all subsequent runs have a single child process which indicates to me that there is something actually going on here with resource clean up.

You can see this by changin `n_jobs` to `[2,1,8]` - the first test has two children (makes sense), but then subsequent tests all have the same 4 children, two of which are the first two children! I think something is not being cleaned up.

selmanozleyen: It can be a cleanup issue but maybe joblib is using created processes as a pool? to run other workers in case to reuse the allocated processes?

selmanozleyen: > "Note that the 'loky' backend now used by default for process-based parallelism automatically tries to maintain and reuse a pool of workers by it-self even for calls without the context manager."

https://joblib.readthedocs.io/en/latest/parallel.html

ilan-gold: Can we check this without loky then?

selmanozleyen: then it fails with multiprocessing. I am looking into it

src/squidpy/_utils.py
selmanozleyen: > Is this possible?

I imagined `numba.set_num_threads` to be changing the ENV variable. I checked the docs and it wasn't clear if the change was temporary or not. Finally I found in the implementations when one uses `numba.set_num_threads` it sets a mask specific to that thread like in this: https://github.com/numba/numba/blob/50abd703c88cda775e6a68cd7be442004ddf1fbf/numba/np/ufunc/omppool.cpp#L57-L63

Otherwise I wouldn't be sure if two scripts using parallelize could change each others thread count or not. So I think setting it to the previous value seems safe enough.

selmanozleyen:
To check if the new processes inherited the `set_numba_cores` I also wanted to understand how this function creates the processes.

ilan-gold:
> To check if the new processes inherited the set_numba_cores I also wanted to understand how this function creates the processes.

But the test was only checking for processes being created/cleaned up, so I think we would instead somehow want to check the environment variable on each process (somehow), no? Not sure how to go about this given the current API TBH but possible you could create a dummy function that gets passed into `_parallelize` and just calls https://numba.pydata.org/numba-doc/dev/user/threading-layer.html#numba.get_num_threads

selmanozleyen:
I agree on this, I just got lost a bit with the curiosity.

Issue #957, opened 2025-02-18
spatial_autocorr uses all the cores


GloriaLiu28:
Hi,
when I run squidpy.gr.spatial_autocorr to calculate moran, it used all the 112 cores of my server, even the n_jobs were set to n_jobs=8, how to limit the number of cores it uses ? I find the old issues about occurence also talked about this. Could you please also  set numba_parallell = False to squidpy.gr.spatial_autocorr  function?

maltekuehl:
For a quick fix, you should be able to set [the environment variable NUMBA_NUM_THREADS](https://numba.pydata.org/numba-doc/dev/user/threading-layer.html).

selmanozleyen:
hi @GloriaLiu28 can you check if this still exists with the most recent version of scanpy?

 @flying-sheep I noticed we use gearys stats in this `spatial_autocorr` and some of the gearys related metrics use a decorator from scanpy which might set the `parallel=True` https://github.com/scverse/scanpy/blob/b058b1792b91b1baaf2e5caf395960005098c2ac/src/scanpy/metrics/_gearys_c.py#L233-L234. Here is the decorator https://github.com/scverse/scanpy/blob/b058b1792b91b1baaf2e5caf395960005098c2ac/src/scanpy/_compat.py#L116. The function is compiled parallel if `not _is_in_unsafe_thread_pool()` but I am not sure if this means `spatial_autocorr` will run a parallel numba kernel and I think we need to make sure it doesn't.

flying-sheep:
It will not be parallelized when the current thread’s name starts with `ThreadPoolExecutor`, which is the case in dask.

joblib seems to use this: https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.ThreadPool

so scanpy’s workaround will probably not work there. we should augment it so it does.

I also found the way joblib adapts to manually reduced thread numbers, which seems like a nice starting point for us: https://github.com/joblib/joblib/blob/ed0806a497268005ad7dad30f79e1d563927d7c6/joblib/_parallel_backends.py#L65

PR #1009, opened 2025-06-11
Fix the problems with the latest merge

selmanozleyen:
This implements the fixes I mentioned here: https://github.com/scverse/squidpy/pull/1008#issuecomment-2962018891

### DIFF
--- a/src/squidpy/_utils.py
+++ b/src/squidpy/_utils.py
@@ -57,8 +57,12 @@ def _unique_order_preserving(
 
 
 def _callback_wrapper(chosen_runner: Callable[..., Any], *args: Any, **kwargs: Any) -> Any:
+    original_num_threads = numba.get_num_threads()
     numba.set_num_threads(1)
-    return chosen_runner(*args, **kwargs)
+    try:
+        return chosen_runner(*args, **kwargs)
+    finally:
+        numba.set_num_threads(original_num_threads)
 
 
 class Signal(Enum):
@@ -172,29 +176,38 @@ def parallelize(
     chosen_runner = runner if use_runner else callback
 
     def wrapper(*args: Any, **kwargs: Any) -> Any:
+        original_num_threads = numba.get_num_threads()
+        # setting it also here but will
+        # also set inside the callback wrapper because
+        # it's not entirely clear in numba documentation
+        # whether this thread number mask will be applied
+        # to other processes spawned by joblib.
         numba.set_num_threads(1)
-        if pass_queue and show_progress_bar:
-            pbar = None if tqdm is None else tqdm(total=col_len, unit=unit)
-            queue = Manager().Queue()
-            thread = Thread(target=update, args=(pbar, queue, len(collections)), name="ParallelizeUpdateThread")
-            thread.start()
-        else:
-            pbar, queue, thread = None, None, None
+        try:
+            if pass_queue and show_progress_bar:
+                pbar = None if tqdm is None else tqdm(total=col_len, unit=unit)
+                queue = Manager().Queue()
+                thread = Thread(target=update, args=(pbar, queue, len(collections)), name="ParallelizeUpdateThread")
+                thread.start()
+            else:
+                pbar, queue, thread = None, None, None
 
-        res = jl.Parallel(n_jobs=n_jobs, backend=backend)(
-            jl.delayed(_callback_wrapper)(
-                *((chosen_runner, i, cs) if use_ixs else (chosen_runner, cs)),
-                *args,
-                **kwargs,
-                queue=queue,
+            res = jl.Parallel(n_jobs=n_jobs, backend=backend)(
+                jl.delayed(_callback_wrapper)(
+                    *((chosen_runner, i, cs) if use_ixs else (chosen_runner, cs)),
+                    *args,
+                    **kwargs,
+                    queue=queue,
+                )
+                for i, cs in enumerate(collections)
             )
-            for i, cs in enumerate(collections)
-        )
 
-        if thread is not None:
-            thread.join()
+            if thread is not None:
+                thread.join()
 
-        return res if extractor is None else extractor(res)
+            return res if extractor is None else extractor(res)
+        finally:
+            numba.set_num_threads(original_num_threads)
 
     if n_jobs is None:
         n_jobs = 1
--- a/tests/utils/test_parallelize.py
+++ b/tests/utils/test_parallelize.py
@@ -2,19 +2,23 @@
 
 from __future__ import annotations
 
-import time
 from collections.abc import Callable
 from functools import partial
 
 import dask.array as da
 import numba
 import numpy as np
-import psutil
 import pytest  # type: ignore[import]
 
 from squidpy._utils import Signal, parallelize
 
 # Functions to be parallelized
+
+
+def wrap_numba_check(x, y, inner_function, check_threads=True):
+    if check_threads:
+        assert numba.get_num_threads() == 1
+    return inner_function(x, y)
 
 
 @numba.njit(parallel=True)
@@ -38,9 +42,9 @@ def vanilla_func(x, y) -> np.ndarray:
 # Mock runner function
 
 
-def mock_runner(x, y, queue, func):
+def mock_runner(x, y, queue, function):
     for i in range(len(x)):
-        x[i] = func(x[i], y)
+        x[i] = function(x[i], y, check_threads=True)
         if queue is not None:
             queue.put(Signal.UPDATE)
     if queue is not None:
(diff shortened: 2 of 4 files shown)


src/squidpy/_utils.py
```diff
@@ -172,29 +172,38 @@ def update(pbar: tqdm.std.tqdm, queue: SigQueue, n_total: int) -> None:
     chosen_runner = runner if use_runner else callback
 
     def wrapper(*args: Any, **kwargs: Any) -> Any:
+        original_num_threads = numba.get_num_threads()
+        # setting it also here but will
+        # also set inside the callback wrapper because
+        # it's not entirely clear in numba documentation
+        # whether this thread number mask will be applied
+        # to other processes spawned by joblib.
```
ilan-gold: I think then maybe we want to set the environment variable within each launched job as well as `set_num_threads`, no? https://numba.pydata.org/numba-doc/dev/user/threading-layer.html#setting-the-number-of-threads

I see https://joblib.readthedocs.io/en/latest/developing.html#release-0-14-1 explicitly mentioning this use-case

ilan-gold: This might help keep things locally restricted as well - no risk of overriding the user's setting if we set the variable in the process

selmanozleyen: Thanks for pointing this out. When I follow the PR's, this is set with 
> Since joblib 0.14, it is also possible to programmatically override the default number of threads using the inner_max_num_threads argument of the [parallel_config()](https://joblib.readthedocs.io/en/latest/generated/joblib.parallel_config.html#joblib.parallel_config) function as follows:

```python
from joblib import Parallel, delayed, parallel_config

with parallel_config(backend="loky", inner_max_num_threads=2):
    results = Parallel(n_jobs=4)(delayed(func)(x, y) for x, y in data)
```
https://joblib.readthedocs.io/en/latest/parallel.html

However, when I comment out `set_numba_threads` it doesn't pass the test in 
```
def wrap_numba_check(x, y, inner_function, check_threads=True):
    if check_threads:
        assert numba.get_num_threads() == 1
    return inner_function(x, y)
```
But I will still add it.

Also when I try to set the `os.environ["NUMBA_NUM_THREADS"] = "1"` before the jobs spawn I get this error
```
            from numba.np.ufunc import parallel
            if parallel._is_initialized:
>               raise RuntimeError("Cannot set NUMBA_NUM_THREADS to a "
                                   "different value once the threads have been "
                                   "launched (currently have %s, "
                                   "trying to set %s)" %
                                   (_NUMBA_NUM_THREADS,
                                    globals()['NUMBA_NUM_THREADS']))
E               RuntimeError: Cannot set NUMBA_NUM_THREADS to a different value once the threads have been launched (currently have 1, trying to set 8)
```
I tried this in case the backend wasn't loky but I don't think this will work well.

ilan-gold: Hmm, ok @selmanozleyen I also see that the joblib doesn't allow setting environment variables for workers in loky ahead of startup.  Maybe then we should just simplify things, perhaps we just do `set_num_threads` before launching and then reset after, and only test that the number is right in each job via `get_num_threads`.

I believe, at the risk of numba launching more threads than we need, that this would be a solution to the underlying issue of oversubscription: #957

> Also when I try to set the os.environ["NUMBA_NUM_THREADS"] = "1" before the jobs spawn I get this error

We can then make clear to users that if they want to mitigate the extra threads issue, they can set the numba env variable themselves to 1.

And what happens if you set this variable within each process?

selmanozleyen: > And what happens if you set this variable within each process?

I get the same error as above for `numba_serial-x` only which is strange.

I also noticed pids printed for 
```python
def wrap_numba_check(x, y, inner_function, check_threads=True):
    if check_threads:
        # print pid
        print(f"PID: {os.getpid()}")
        assert numba.get_num_threads() == 1
    return inner_function(x, y)
```
are the same. So this means I am using parallelize wrongly for my case.

ilan-gold: > are the same. So this means I am using parallelize wrongly for my case.

I don't follow.  What is wrong and what is "my case?"

selmanozleyen: For the test case I wrote I mean. I expected each function call to have a seperate pid but it doesn't

tests/utils/test_parallelize.py
```diff
@@ -51,54 +55,36 @@ def mock_runner(x, y, queue, func):
...
     seed = 42
     rng = np.random.RandomState(seed)
     n = 8
     arr1 = [rng.randint(0, 100, n) for _ in range(n)]
     arr2 = np.arange(n)
-    runner = partial(mock_runner, func=func)
+    runner = partial(mock_runner, function=func)
     # this is the expected result of the function
-    expected = [func(arr1[i], arr2) for i in range(len(arr1))]
+    expected = [func(a1, arr2, check_threads=False) for a1 in arr1]
```
ilan-gold: I'm confused, maybe lost the plot on the conversation, but if we set the number of threads, under what circumstances would we expect the assertion for checking threads to fail?

selmanozleyen: these are not spawned jobs so the number of threads here would be 3 for this particulat function for example. The number of threads are set inside the parallelize function itself. That's why I only check later with `p_func = parallelize(runner, arr1, n_jobs=n_jobs, backend="loky", use_ixs=False, n_split=1)`

ilan-gold: Ah, yes, duh!

selmanozleyen:
@ilan-gold I can confirm with this script and checking from htop https://gist.github.com/selmanozleyen/0af34e05d7704371fa609a1011bb422c this program uses `njobs` cores

tests/utils/test_parallelize.py
```diff
@@ -51,54 +55,36 @@ def mock_runner(x, y, queue, func):
...
-        "vanilla": vanilla_func,
+        "numba_parallel": partial(wrap_numba_check, inner_function=numba_parallel_func),
+        "numba_serial": partial(wrap_numba_check, inner_function=numba_serial_func),
+        "dask": partial(wrap_numba_check, inner_function=dask_func),
+        "vanilla": partial(wrap_numba_check, inner_function=vanilla_func),
     }[request.param]
 
 
-@pytest.mark.timeout(60)
+@pytest.mark.timeout(200)
```
ilan-gold: Why does this need so long?

selmanozleyen: I have no idea. Before it passed because it didn't actually create `n_jobs` processes. My guess is the overhead for creating a process. For example when I set `n_jobs=1` the script I shared ran way faster.

src/squidpy/_utils.py
flying-sheep: if you use a process-based backend, what should happen is:
1. it starts a new python process
2. it sets all the env variables in the correct way (based on `inner_max_num_threads`)
3. it imports all the stuff (e.g. numba) for the first time in that process, therefore numba will not cry about *its* parallel backend already being initialized

if that's not what happens, something is wrong and you're not actually using process-based parallelization I think.

selmanozleyen: sorry, I guess I did a mistake when trying out setting `inner_max_num_threads` now it passes the tests without ever using `numba.set_num_threads` but it won't work with `multiprocessing` backend in which there are tests cases in squidpy. We can ignore `backend=multiprocessing` and as it is said to be the legacy option in joblib docs anyway

selmanozleyen: so in the docs I mentioned `inner_max_num_threads=1` is set only for `loky` backend therefore over-subscription is only handled in that case similar to `joblib`. I think this is the best possible solution and it doesn't use `numba.set_num_threads`. Do you think its good enough to merge?

tests/utils/test_parallelize.py
```diff
@@ -51,54 +55,26 @@ def mock_runner(x, y, queue, func):
...
-        "dask": dask_func,
-        "vanilla": vanilla_func,
+        "numba_parallel": partial(wrap_numba_check, inner_function=numba_parallel_func),
+        "numba_serial": partial(wrap_numba_check, inner_function=numba_serial_func),
+        "dask": partial(wrap_numba_check, inner_function=dask_func),
+        "vanilla": partial(wrap_numba_check, inner_function=vanilla_func),
     }[request.param]
 
 
 @pytest.mark.timeout(60)
```
ilan-gold: Why do we still need this?

src/squidpy/_utils.py
```diff
@@ -172,29 +172,29 @@ def update(pbar: tqdm.std.tqdm, queue: SigQueue, n_total: int) -> None:
...
         else:
             pbar, queue, thread = None, None, None
-
-        res = jl.Parallel(n_jobs=n_jobs, backend=backend)(
-            jl.delayed(_callback_wrapper)(
-                *((chosen_runner, i, cs) if use_ixs else (chosen_runner, cs)),
-                *args,
-                **kwargs,
-                queue=queue,
+        jl_kwargs = {"inner_max_num_threads": 1} if backend == "loky" else {}
```
ilan-gold, line 182: If the intention is not to support any other backends in squidpy, we should deprecate them on our side.

If the intention is continue supporting other backends, we should make sure this feature works with those.

@timtreis What do you think here?  I would guess the threading backend is useful for testing but not sure.  We are currently testing different backends in the codebase at least once from what I can see.

tests/utils/test_parallelize.py
selmanozleyen: I thought it would be a sanity check with different functions that support numba and dask. For example if there was a function with jax it would throw a warning saying jax doesn't work well with `os.fork` or something like that. If you think its not worth it I can remove it

ilan-gold: Hmm, ok yes, I suppose it couldn't hurt. But I would maybe make the number lower, as tight a bound as is reasonable.

selmanozleyen: you mean the timeout number right?

ilan-gold: Yup

src/squidpy/_utils.py
ilan-gold: I think it would probably be easier to just special case the `set_num_threads` to only be used for non-loky backends and not worry about what squidpy supports or doesn't @selmanozleyen

selmanozleyen: okay

tests/utils/test_parallelize.py
selmanozleyen: Ok I set it to 30 seconds. And other backends are also supported. I can add `multiprocessing` to the tests to check for that case but I didn't. Other than that everything should be ready

tests/utils/test_parallelize.py
```diff
@@ -62,8 +62,12 @@ def func(request) -> Callable:
     }[request.param]
 
 
-@pytest.mark.timeout(60)
-def test_parallelize_loky(func):
+# Timeouts are also useful because some processes don't return in
+# in case of failure.
```
ilan-gold, line 67: Are failures expected though?

tests/utils/test_parallelize.py
```diff
@@ -62,8 +62,12 @@ def func(request) -> Callable:
...
+
+@pytest.mark.parametrize("backend", ["loky", "multiprocessing"])
```
ilan-gold: Why no `threading`?

tests/utils/test_parallelize.py
ilan-gold: Sorry what happened here? I saw a commit with `multiprocessing` but no `threading` (?) but now I don't see it in the diff.

EDIT: ok https://github.com/scverse/squidpy/pull/1009/commits/b850762329fd336f615dcbd09755e8ff8b2e09d8 so you can check if something is being run in CI via the [eponymous environment variable](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables) and skip only there

tests/utils/test_parallelize.py
selmanozleyen: no but that was why the tests were 1hours before. Just to shorten that

tests/utils/test_parallelize.py
selmanozleyen: damn thanks thats cool! But a test still requires more than 30 seconds and I am not sure if I should just loosen the timeout. I guess I can also skip the slow ones.

tests/utils/test_parallelize.py
ilan-gold: ah ok, i misunderstood

tests/utils/test_parallelize.py
```diff
@@ -51,54 +56,45 @@ def mock_runner(x, y, queue, func):
...
+
+
+@pytest.mark.timeout(30)
+@pytest.mark.skipif(os.environ.get("CI") == "true", reason="Only testing 'loky' backend in CI environment")
+@pytest.mark.parametrize("backend", ["threading", "multiprocessing"])
+def test_parallelize_others(func, backend):
+    _test_parallelize(func, backend)
+
+
+def _test_parallelize(func, backend):
```
ilan-gold:
```suggestion
@pytest.mark.timeout(30)
@pytest.mark.parametrize("backend", [pytest.param("threading", marks=pytest.mark.skipif(os.environ.get("CI") == "true", reason="Only testing 'loky' backend in CI environment")), pytest.param("multiprocessing", marks=pytest.mark.skipif(os.environ.get("CI") == "true", reason="Only testing 'loky' backend in CI environment")), "loky"])
def test_parallelize(func, backend):
```

Or something to this effect.

ilan-gold (approved):

PR #1013, opened 2025-06-30
Run notebooks in CI

selmanozleyen:
Closes https://github.com/scverse/squidpy/issues/993

### DIFF
--- a/.run_notebooks.sh
+++ b/.run_notebooks.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+
+# Check if the base directory is provided as an argument
+if [ "$#" -ne 1 ]; then
+    echo "Usage: $0 <base_notebook_directory>"
+    exit 1
+fi
+
+# Base directory for notebooks
+base_dir=$1
+
+# Define notebook directories or patterns
+declare -a notebooks=(
+    "$base_dir/examples/tools/*.ipynb"
+    "$base_dir/examples/plotting/*.ipynb"
+)
+
+# Initialize an array to hold valid notebook paths
+declare -a valid_notebooks
+
+# Gather all valid notebook files from the patterns
+echo "Gathering notebooks..."
+for pattern in "${notebooks[@]}"; do
+    for nb in $pattern; do
+        if [[ -f "$nb" ]]; then  # Check if the file exists
+            valid_notebooks+=("$nb")  # Add to the list of valid notebooks
+        fi
+    done
+done
+
+# Check if we have any notebooks to run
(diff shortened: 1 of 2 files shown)


flying-sheep:
@LucaMarconato wants to look into how to do this with caching, see also https://github.com/scverse/cookiecutter-scverse/issues/40

.github/workflows/test-notebooks.yml
```diff
@@ -0,0 +1,41 @@
...
+          submodules: recursive
+      - name: Set up Python ${{ matrix.python }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python }}
+
+      - name: Install pip dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install tox
```
ilan-gold: Use `uv` for a nice speedup!

.run_notebooks.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
```
ilan-gold: 100% agree with Phil, let's use python here

.github/workflows/test-notebooks.yml
```diff
@@ -0,0 +1,41 @@
+name: Test notebooks
```
ilan-gold, line 1: For reproducibility, should we have a check for whether the notebooks change? This would basically turn the notebooks into tests-for-stability, but could be a nice check.  @timtreis or @LucaMarconato have you guys ever received feedback about stability?

.github/workflows/test-notebooks.yml
selmanozleyen: done! should I change to uv + hatch for the rest as well in a different PR?

.github/workflows/test-notebooks.yml
selmanozleyen: we could have expected and output folders to compare but we'd have to update expected when we expect/want the output to change. Similar to plot unit tests. I think this is doable but would fit better as a separate issue imo

timtreis: In what sense "If the notebook changes"? Aren't jupyter notebooks famously annoying to compare? 

I've seen people convert them into .py scripts automatically and then compare these but I'm not sure of that adresses our issue here?

selmanozleyen: I think it's not related to these tests particularly. I am mainly talking about comparing the plot results and maybe the html results. Not the whole notebooks file themselves.

PR #1019, opened 2025-07-28
Add SpatialLeiden as clustering method

### LINKED ISSUE #879: Add SpatialLeiden as clustering method
Basically already (mostly) uses `scanpy` / `squidpy` for preproc steps, could wrap remaining python code and provide as method: 
image

https://www.biorxiv.org/content/10.1101/2024.08.23.609349v1

melocery:
Close #879 

Added support for `spatialleiden` as a flavor in `sq.gr.calculate_niche`, following up on the idea mentioned in #879. The method is now integrated via the `flavor="spatialleiden"` option, wrapping the `spatialleiden.spatialleiden()` function.


timtreis:
Hey @melocery, thank you for PR! However, there's a licensing issue with it, so we can't accept it in this state.

Your PR adds a hard dependency on `SpatialLeiden`. While it mostly mirrors our dependencies, it has `leidenalg` as a hard dependency, which is licensed under GPL3. GPL3 is a copy-left license, and therefore your package is actually also GPL3 (not MIT), and ours would be too. That'd prevent most commercial entities from using Squidpy, which we don't want to prevent.

For us to depend on `SpatialLeiden`, you'd have to find a way how people can install and run your algorithm without the `leidenalg` package.

pyproject.toml
```diff
@@ -99,6 +99,10 @@ optional-dependencies.docs = [
   "sphinxcontrib-bibtex>=2.3",
   "sphinxcontrib-spelling>=7.6.2",
 ]
+optional-dependencies.leiden = [
+  "leidenalg",
+  "spatialleiden>=0.3",
```
timtreis:
```suggestion
  "spatialleiden>=0.4.0",
```

timtreis:
merged! Thank you for the contribution @niklasmueboe !

PR #1023, opened 2025-08-12
Add COZI to nhood_enrichment()

chiarasch:
## Description

- Adds `normalization` parameter for conditional and total normalization of neighbor counts to `nhood_enrichment()` as described in [*Schiller et al bioRxiv*](https://doi.org/10.1101/2025.03.31.646289).  
- Adds conditional cell ratio (CCR) output when `normalization`= `conditional`. CCR is the proportion of cells of an index type that neighbors cells of the neighbor type. 
- Adds `nhood_enrichment_dotplot()`to plot zscore and CCR at once.
- Adds `min_cell_count` parameter to `nhood_enrichment()` to filter out clusters of cells with counts below the set threshold. Filtered out interaction z-scores are replaced with NA. 

## How has this been tested?

Added additional test functions for nhood_enrichment with conditional normalization and CCR. Test images for dotplot are missing, all other tests run successfully.

## TODO

- [x] Add dotplot test images for test


chiarasch:
I get a resolution-too-deep error for dependencies for ubuntu related to tox and the error for macos also seems to be connected to tox. I only added an import warnings in the code, so I did not change any dependencies. Can someone help?

selmanozleyen (changes requested):
The unit tests here aren't enough. Only isnan checks. Isn't there a way to ensure we have sensible results?

Also the doc formatting needs to be fixed so the CI's don't fail.

Meanwhile I also want to do performance improvement in this function and this PR. We can keep that separate though

src/squidpy/pl/_graph.py
```diff
@@ -237,6 +239,160 @@ def nhood_enrichment(
...
+        - Color for z-score of enrichment
+        - Dot size for conditional cell ratio (CCR), scaled continuously
+
+    Parameters
+    ----------
+    adata : AnnData
+        Annotated data matrix.
+    cluster_key : str
+        Key in `adata.obs` where the cluster (cell type) annotation is stored.
+    zscore_key : str, optional
```
selmanozleyen: zscore_key isn't being used here I think

src/squidpy/pl/_graph.py
```diff
@@ -237,6 +239,160 @@ def nhood_enrichment(
...
 
+@d.dedent
+def nhood_enrichment_dotplot(
+    adata: AnnData,
+    cluster_key: str,
+    zscore_key: str = "ct_nhood_enrichment",
+    annotate: bool = False,
+    title: str | None = None,
+    cmap: str = "coolwarm",
+    palette: Palette_t = None,
```
selmanozleyen: also palette

selmanozleyen (approved):
I applied the changes I mentioned. Thanks a lot. I also numbafied nhood_enrichment here

selmanozleyen:
Hi @chiarasch ! I changed `normalization="total"` to divide by the number of index cells instead of the row sum of edges, since that's what you do in [scimap_COZI](https://github.com/SchapiroLabor/scimap_COZI/blob/191a1b1501ac463fc619f521462aba264cb34470/scimap/tools/spatial_interaction.py#L315-L316) — so the row sum was probably just a bug.

Also fixed stored zeros being counted as edges (pruning with `adj.data[mask] = 0` kept every pruned edge). COZI itself and the CCR match the paper.

PR #1028, opened 2025-08-24
Fix for spatial_neighbors breaks when transform='spectral'

selmanozleyen:
Fixes https://github.com/scverse/squidpy/issues/1014

### DIFF
--- a/src/squidpy/gr/_build.py
+++ b/src/squidpy/gr/_build.py
@@ -13,9 +13,8 @@ import numpy as np
 import pandas as pd
 from anndata import AnnData
 from anndata.utils import make_index_unique
-from fast_array_utils import njit
 from fast_array_utils import stats as fau_stats
-from numba import prange
+from numba import njit, prange
 from scanpy import logging as logg
 from scipy.sparse import (
     SparseEfficiencyWarning,


src/squidpy/gr/_build.py
```diff
@@ -450,17 +469,60 @@ def outer(indices: NDArrayA, indptr: NDArrayA, degrees: NDArrayA) -> NDArrayA:
...
+
+    Parameters
+    ----------
+    A : scipy.sparse.csr_matrix
+
+    Returns
+    -------
+    scipy.sparse.csr_matrix
+    """
+    degrees = np.squeeze(np.array(np.sqrt(1.0 / A.sum(axis=0))))
```
Intron7: I think we should use fast array utils for this

src/squidpy/gr/_build.py
```diff
@@ -439,7 +439,26 @@ def _build_connectivity(
...
+        Diagonal scaling vector.
+
+    Returns
+    -------
+    array of float
+        Length equals len(indices). Entry-wise factors d_i * d_j.
+    """
+
     res = np.empty_like(indices, dtype=np.float64)
     start = 0
```
Intron7: What does start do please also make the loop parallel

src/squidpy/gr/_build.py
```diff
@@ -450,17 +469,60 @@ def outer(indices: NDArrayA, indptr: NDArrayA, degrees: NDArrayA) -> NDArrayA:
...
+    Returns
+    -------
+    scipy.sparse.csr_matrix
+        Matrix with same sparsity pattern as A.
+    """
+    if A.shape[0] != len(degrees):
+        raise ValueError("len(d) must equal number of rows of A")
+    factors = _didj_factors_csr(A.indices, A.indptr, degrees)
+    F = csr_matrix((factors, A.indices, A.indptr), shape=A.shape)
+    B = A.multiply(F)
```
Intron7: Since this is an elementwise multiplication can we make this into a numba kernel

src/squidpy/gr/_build.py
flying-sheep: Also for any function containing a parameter called `indices` and/or `indptr`, it would be better to use `fau` and simply use one parameter.

src/squidpy/gr/_build.py
flying-sheep: am I misunderstanding something or is this just

```py
b = csr_matrix((a.data * factors, a.indices, a.indptr), shape=a.shape)
```

Also please follow [naming rules](https://docs.astral.sh/ruff/rules/#pep8-naming-n):
- please don’t use single letter variables unless it’s a loop index or something like that
- use `UPPERCASE_WITH_UNDERSCORES` for constants

selmanozleyen: oh I thought multiply was matrix multp.🤦 my bad

src/squidpy/gr/_build.py
```diff
@@ -453,29 +453,67 @@ def _build_connectivity(
     return Adj
 
 
-@njit
-def outer(indices: NDArrayA, indptr: NDArrayA, degrees: NDArrayA) -> NDArrayA:
-    res = np.empty_like(indices, dtype=np.float64)
-    start = 0
-    for i in range(len(indptr) - 1):
+@njit(parallel=True)
```
Intron7: you should also cache this kernel

src/squidpy/gr/_build.py
```diff
@@ -453,29 +453,67 @@ def _build_connectivity(
...
+    degrees : array of float, shape (n,)
+        Diagonal scaling vector.
+
+    Returns
+    -------
+    array of float
+        Length equals len(data). Entry-wise factors d_i * d_j * data[k]
+    """
+
+    res = np.empty_like(data, dtype=np.float64)
```
Intron7: why np.float64?

selmanozleyen: because original function used np.float64. Didn't want to change old behaviour

pyproject.toml
```diff
@@ -76,6 +76,7 @@ dependencies = [
    "xarray>=2024.10.0",
    "imagecodecs>=2025.8.2,<2026",
    "zarr>=2.6.1",
+   "fast_array_utils",
```
flying-sheep:
```suggestion
   "fast-array-utils",
```

src/squidpy/gr/_build.py
```diff
@@ -453,29 +453,67 @@ def _build_connectivity(
...
 
 
-@njit
-def outer(indices: NDArrayA, indptr: NDArrayA, degrees: NDArrayA) -> NDArrayA:
-    res = np.empty_like(indices, dtype=np.float64)
-    start = 0
-    for i in range(len(indptr) - 1):
+@njit(parallel=True, cache=True)
+def _csr_bilateral_diag_scale_helper(
+    data: NDArrayA, indices: NDArrayA, indptr: NDArrayA, degrees: NDArrayA
```
flying-sheep:
```suggestion
    mat: csr_matrix | csr_array, degrees: NDArrayA
```

`fast-array-utils` allows this

flying-sheep (changes requested):

src/squidpy/gr/_build.py
selmanozleyen: thanks, didn't know that

src/squidpy/gr/_build.py
Intron7: can we check if this is needed

selmanozleyen: well I can't tell tbh because there was no tests for the spectral transform lol. We can just use float32 I guess. But the better way would be to ask ths users of `spatial_neighbours`

selmanozleyen: @Intron7 ok the results seem similar when I checked them. I also added some numerical tests to ensure stability. Also when I recalled that this was just a normalization of weights in a graph so it should not be very sensitive to precision in theory.

selmanozleyen:
@timtreis could you merge this please? @Intron7  said he can't review it and @flying-sheep already had a look on this

PR #1029, opened 2025-08-24
Moving to uv + hatch

selmanozleyen:
Fixes https://github.com/scverse/squidpy/issues/1027

I mainly moved from `tox+pip` to `hatch+uv`.
The CI's should be equivalent but I changed:
- Changed matrix to be more of an explicit list
- Used `github.workspace` instead of hardcoded path for linux runners

### DIFF
--- a/tests/datasets/test_dataset.py
+++ b/tests/datasets/test_dataset.py
@@ -23,6 +23,7 @@ class TestDatasetsImports:
 # TODO(michalk8): parse the code and xfail iff server issue
 class TestDatasetsDownload:
     @pytest.mark.timeout(120)
+    @pytest.mark.internet()
     def test_download_imc(self, tmp_path: Path):
         with warnings.catch_warnings():
             warnings.simplefilter("ignore", category=OldFormatWarning)
@@ -35,6 +36,7 @@ class TestDatasetsDownload:
                 pytest.xfail(str(e))
 
     @pytest.mark.timeout(120)
+    @pytest.mark.internet()
     def test_download_visium_hne_image_crop(self, tmp_path: Path):
         with warnings.catch_warnings():
             warnings.simplefilter("ignore", category=OldFormatWarning)


.github/workflows/test.yml
```diff
@@ -1,104 +1,108 @@
...
+  PYTEST_ADDOPTS: "-v --color=yes -n auto"
+  FORCE_COLOR: "1"
+  MPLBACKEND: agg
+  # It's impossible to ignore SyntaxWarnings for a single module,
+  # so because leidenalg 0.10.0 has them, we pre-compile things: https://github.com/vtraag/leidenalg/issues/173
+  UV_COMPILE_BYTECODE: "1"
+
+defaults:
+  run:
+    shell: bash -e {0} # -e to fail on error
```
flying-sheep: This isn’t actually necessary: https://github.com/actions/runner-images/issues/4459#issuecomment-965290856

```suggestion
```

.github/workflows/test.yml
```diff
@@ -1,104 +1,108 @@
...
-            pre-release: "allow"
-            name: "Python 3.12 (pre-release)"
-
-    env:
-      OS: ${{ matrix.os }}
-      PYTHON: ${{ matrix.python }}
-      UV_PRERELEASE: ${{ matrix.pre-release || 'disallow' }}
-
+            python: "3.12"
+            prerelease: "allow"
```
flying-sheep: derive this from `hatch show --json`. For this to work, you’ll need to have all jobs runnable locally via `hatch`. Which isn’t hard. E.g. the cookiecutter’s configuration shows how to do that: https://github.com/scverse/cookiecutter-scverse-instance/blob/0bdc68ecaaa73fc44adcddb7f80420fbb46ad588/pyproject.toml#L69-L77

scanpy’s workflow has an example for how to derive `test-type`: https://github.com/scverse/scanpy/blob/6cf550fa5045c9ee0bc0d2b8476fea9bc2bb2af4/.github/workflows/ci.yml#L38

but tbh you could also just run coverage for all test jobs. It’s a bit slower but not that much I think.

.github/workflows/test.yml
```diff
@@ -1,104 +1,108 @@
...
+    env: # environment variable for use in codecov's env_vars tagging
+      ENV_NAME: ${{ matrix.name }}
+      UV_PRERELEASE: ${{ matrix.prerelease || 'disallow' }}
```
flying-sheep:
```suggestion
```

should be done in hatch, see above

.github/workflows/test.yml
```diff
@@ -1,104 +1,108 @@
...
+          path: .pytest_cache/d/squidpy-data
+          key: pytest
 
-      - name: Download datasets
-        if: steps.data-cache.outputs.cache-hit != 'true'
+      - name: System dependencies (Linux)
+        if: matrix.os == 'ubuntu-latest'
         run: |
-          tox -e download-data
+          set -u # fail on unset variables
```
flying-sheep: Which variables? I don’t see any being used here.

hatch.toml
```diff
@@ -0,0 +1,51 @@
...
+
+
+
+[[envs.hatch-test.matrix]]
+python = ["3.10", "3.11", "3.12"]
+deps = ["stable"]
+
+[envs.hatch-test.scripts]
+run = "python -m pytest --color=yes --cov --cov-append --cov-report=xml --cov-config=pyproject.toml --ignore docs/ {args:-vv}"
+run-cov = "python -m pytest --color=yes --cov --cov-append --cov-report=xml --cov-config=pyproject.toml --ignore docs/ {args:-vv}"
```
flying-sheep: why are these defined at all?

don’t do anything with `--cov*` in `run`. `run` is “run without coverage”.

regarding `run-cov`, keep in mind that it’s used for local and CI usage, so the default should work fine as nothing that’s done here should be here:

- `--color=yes` isn’t a good idea. We know it works in CI, but the default detection should happen locally
- `--cov` isn’t necessary, the default uses `coverage run` instead.
- `--cov-append` does nothing I think
- `--cov-report=xml` should definitely not be run by default, so create the XML report in the GH workflow yaml
- `--cov-config=pyproject.toml` also does nothing I think
- `--ignore=docs` is configuration you always want, so it should happen in the `[tool.pytest.ini_options]` table in `pyproject.toml`.

flying-sheep (changes requested):
You’re removing the regular cronjob run, should you?

Looks generally really good! There are some minor redundancies, and some confusion around what needs to be done to collect coverage.

There are two ways:
- `coverage run -m pytest ...` and then `coverage xml`
- `pytest --cov --cov-report=xml`

since hatch’s default `run-cov` uses `coverage run -m pytest`, you could use that default in two ways to collect coverage and then create the XML report:

- ```yaml
  run: uvx hatch run ${{ matrix.name }}:run-cov --cov --cov-report=xml
  ```
- ```yaml
  run: |
    uvx hatch run ${{ matrix.name }}:run-cov
    uvx hatch run ${{ matrix.name }}:coverage combine
    uvx hatch run ${{ matrix.name }}:coverage xml
  ```

we had issues with the second one which is why we went for the first one, even if that’s a bit weird, as we’re esssentially have two instances of `coverage.py`: one from `coverage run` and one from `pytest-cov`’s `--cov`

.github/workflows/test.yml
selmanozleyen: I have two reasons to do it this way:
- I couldn't derive the OS in `hatch show --json` and wanted an equivalent to old code.
- I think I could do it with hatch show --json but I liked that it was explicit and easy to understand

If you think that's better practice I will do that. I will try to use different os'es for some of the configurations but I am worried that it might look too complicated

selmanozleyen: anyway I think having all runs for ubuntu and macos is ok I guess. so I adopted the fromJson approach

flying-sheep: Congrats on doing it!

The advantage of having it like this is usually
1. the hatch matrix is kept in sync with the CI matrix (no difference between testing locally and on CI)
2. you don’t have to touch CI when you touch the matrix

It’s pretty complex now, the first advantage is gone, but the second advantage is still there!

flying-sheep (approved):
looks great! sorry for making you do all the `jq` wizardry, I didn’t realize you’d go for preserving the matrix exactly.

selmanozleyen:
thanks @flying-sheep but I still prefer the explicit way for my use case because it might be very confusing to someone looking at this when they want to change the python versions. It's simpler in your case but `test.yml` would also be involved again with the versions.

about the numerical errors I created https://github.com/scverse/squidpy/issues/1030. I'd like to have your opinion on how I should approach it

PR #1036, opened 2025-09-14
Functions to QC histopathology images

### LINKED ISSUE #1034: Function to QC H&E images

timtreis:


src/squidpy/experimental/im/_detect_tissue.py
```diff
@@ -17,9 +17,9 @@
...
 from spatialdata.transformations import get_transformation
 
-from squidpy._utils import _get_scale_factors, _yx_from_shape
+from squidpy._utils import _ensure_dim_order, _get_scale_factors, _yx_from_shape
 
-from ._utils import _flatten_channels, _get_image_data
+from ._utils import _flatten_channels, _get_element_data
 
 
 class DETECT_TISSUE_METHOD(enum.Enum):
```
selmanozleyen: I just noticed this Enum. I think the Enum classes themselves are supposed to be in UpperCamelCase.

https://docs.python.org/3/howto/enum.html
> Because Enums are used to represent constants, and to help avoid issues with name clashes between mixin-class methods/attributes and enum names, we strongly recommend using UPPER_CASE names for members, and will be using that style in our examples.

selmanozleyen:
```suggestion
class DetectTissueMethod(enum.Enum):
```

selmanozleyen (approved):

src/squidpy/gr/_utils.py
```diff
@@ -51,6 +51,85 @@ def _check_tuple_needles(
     return filtered
 
 
+# modified from pandas' source code
```
selmanozleyen: Sorry why are we reintroducing this function again?

src/squidpy/gr/_ligrec.py
```diff
@@ -425,16 +426,18 @@ def test(
...
+                fill_value=0,
             ),
-            "pvalues": pd.DataFrame(
-                {c: pd.arrays.SparseArray(res.pvalues[:, i], fill_value=np.nan) for i, c in enumerate(columns)},
-                index=index,
+            "pvalues": _create_sparse_df(
+                res.pvalues,
+                index=pd.MultiIndex.from_frame(interactions, names=[SOURCE, TARGET]),
+                columns=pd.MultiIndex.from_tuples(clusters, names=["cluster_1", "cluster_2"]),
+                fill_value=np.nan,
```
selmanozleyen: Please undo the changes here. I think this is a result of merge confict resolution

src/squidpy/experimental/im/_utils.py
```diff
@@ -1,9 +1,107 @@
...
+        self.offset_x = offset_x
+        # Calculate number of tiles needed to cover entire image, accounting for offset
+        grid_start_y = min(0, self.offset_y)
+        grid_start_x = min(0, self.offset_x)
+        total_h_needed = self.H - grid_start_y
+        total_w_needed = self.W - grid_start_x
+        self.tiles_y = (total_h_needed + self.ty - 1) // self.ty
+        self.tiles_x = (total_w_needed + self.tx - 1) // self.tx
+        # Cache immutable derived values
+        self._indices = np.array(
```
selmanozleyen: This can be vectorized with numpy I think. If you want me to do it lmk.

src/squidpy/experimental/im/_utils.py
```diff
@@ -1,9 +1,107 @@
...
+            x0 = max(0, min(x0, self.W))
+            y1 = max(0, min(y1, self.H))
+            x1 = max(0, min(x1, self.W))
+            b.append([y0, x0, y1, x1])
+        return np.array(b, dtype=int)
+
+    def centroids_and_polygons(self) -> tuple[np.ndarray, list[Polygon]]:
+        return self._centroids_polys
+
+    def _compute_centroids_and_polygons(self) -> tuple[np.ndarray, list[Polygon]]:
```
selmanozleyen: cents can be easily vectorizable and we can use shapely API for doing this instead of for loops

src/squidpy/experimental/im/_utils.py
```diff
@@ -1,9 +1,107 @@
...
+    def indices(self) -> np.ndarray:
+        return self._indices
+
+    def names(self) -> list[str]:
+        return self._names
+
+    def bounds(self) -> np.ndarray:
+        return self._bounds
+
+    def _compute_bounds(self) -> np.ndarray:
```
selmanozleyen: this should also be vectorizable

src/squidpy/experimental/im/_sharpness_metrics.py
```diff
@@ -0,0 +1,151 @@
...
+                    jj = _clamp(j + dj, 0, w - 1)
+                    y += lk[di + 1, dj + 1] * block[ii, jj]
+            s += y
+            s2 += y * y
+    mean = s / n
+    # var = E[y^2] - (E[y])^2
+    var = (s2 / n) - (mean * mean)
+    var_val = var if var > 0.0 else 0.0
+    return np.array([[var_val]], dtype=np.float32)
+
```
selmanozleyen: these are just convolutions (from CV not signal processing) why hard-code them like this?

src/squidpy/experimental/im/_sharpness_metrics.py
```diff
@@ -0,0 +1,151 @@
...
+@njit(cache=True, fastmath=True)
+def _clamp(v: int, lo: int, hi: int) -> int:
+    if v < lo:
+        return lo
+    if v > hi:
+        return hi
+    return v
+
+
+@njit(cache=True, fastmath=True)
```
selmanozleyen:
```python
from skimage.filters import sobel_h, sobel_v

def _tenengrad_mean(block: np.ndarray) -> np.ndarray:
    energy = sobel_h(b) ** 2 + sobel_v(b) ** 2
    return np.array([[float(energy.mean())]], dtype=np.float32)
```

Maybe this works instead? or maybe `scipy.ndimage.convolve(..., mode="nearest")` or something like this?

src/squidpy/experimental/im/_qc_metrics.py
```diff
@@ -0,0 +1,213 @@
...
+def _tissue_fraction(block: np.ndarray) -> np.ndarray:
+    """Fraction of pixels that are tissue (nonzero) in a binary mask tile."""
+    return np.array([[float(block.mean())]], dtype=np.float32)
+
+
+# --- Registry ---
+
+_METRIC_REGISTRY: dict[QCMetric, tuple[InputKind, MetricFn]] = {
+    # Sharpness (grayscale)
+    QCMetric.TENENGRAD: (InputKind.GRAYSCALE, lambda a: _tenengrad_mean(_ensure_f32_2d(a))),
```
selmanozleyen: here you already ensure f32_2d. why have it in the function also again? Also I don't like functions being called ensure and that changes the input. It gives the idea that it just makes checks and no modification. I would prefer having a more extended InputKind that already implies that its f32 and 2d

src/squidpy/experimental/im/_qc_metrics.py
```diff
@@ -0,0 +1,213 @@
...
+# --- Tissue coverage (mask input) ---
+
+
+def _tissue_fraction(block: np.ndarray) -> np.ndarray:
+    """Fraction of pixels that are tissue (nonzero) in a binary mask tile."""
+    return np.array([[float(block.mean())]], dtype=np.float32)
+
+
+# --- Registry ---
+
```
selmanozleyen, line 71: like the idea with the metric registry puts structure into complicated input types. Should consider adopting something like this more generally. Also do you know how well this works with spatialdata images and if they use something similar?

src/squidpy/experimental/im/_sharpness_metrics.py
```diff
@@ -0,0 +1,151 @@
...
+                    gx += gxk[di + 1, dj + 1] * v
+                    gy += gyk[di + 1, dj + 1] * v
+            s += gx * gx + gy * gy
+    mean_val = s / (h * w)
+    return np.array([[mean_val]], dtype=np.float32)
+
+
+@njit(cache=True, fastmath=True)
+def _laplacian_variance(block: np.ndarray) -> np.ndarray:
+    """Population variance of Laplacian response."""
```
selmanozleyen, line 22: this is just `skimage.filters.laplace` right? with slight modifications? there is surely a way to do this in less code with scikit image.

src/squidpy/experimental/im/_sharpness_metrics.py
```diff
@@ -0,0 +1,151 @@
...
+    s2 = 0.0
+    for i in range(h):
+        for j in range(w):
+            v = block[i, j]
+            s += v
+            s2 += v * v
+    mean = s / n
+    var = (s2 / n) - (mean * mean)
+    var_val = var if var > 0.0 else 0.0
+    return np.array([[var_val]], dtype=np.float32)
```
selmanozleyen: probably np.var is enough right?

hatch.toml
```diff
@@ -39,14 +39,17 @@ download = "python ./.scripts/ci/download_data.py {args}"
 deps = ["stable"]
 python = ["3.11", "3.12", "3.13"]
 
-[[envs.hatch-test.matrix]]
```
selmanozleyen, line 42: also undo these please

tests/conftest.py
```diff
@@ -162,8 +162,7 @@ def adata_squaregrid() -> AnnData:
 def paul15() -> AnnData:
     # session because we don't modify this dataset
     adata = sc.datasets.paul15()
-    sc.pp.filter_cells(adata, min_counts=1)
```
selmanozleyen, line 165: also undo this

selmanozleyen (changes requested):
Things to do:
- Sync with main (some files are old)
- Vectorize for loops I mentioned
- Use scikit-image when we can

src/squidpy/experimental/im/_qc_metrics.py
timtreis: > like the idea with the metric registry puts structure into complicated input types. 

agree!

> Also do you know how well this works with spatialdata images and if they use something similar?

Wdym?

selmanozleyen: > Wdym?

like in general I wonder if there is something already similar to this in spatialdata codebase. Not the registry itself but the InputKind.

timtreis: Ahh, not that I'm aware of 🤔 We're channel agnostic. In sdata-plot I try to infer this from the channel names but formally it doesn't exist

src/squidpy/experimental/im/_qc_metrics.py
```diff
@@ -0,0 +1,212 @@
...
+    QCMetric.HEMATOXYLIN_STD,
+    QCMetric.EOSIN_MEAN,
+    QCMetric.EOSIN_STD,
+    QCMetric.HE_RATIO,
+    QCMetric.FOLD_FRACTION,
+}
+
+
+# --- Intensity metrics (grayscale input) ---
+
```
selmanozleyen: Maybe you aren't done yet but while I have it in mind I want to say I'd put intensity metrics in another file like done in sharpness metrics and have the registry in shorter file.

Also functions called outside a module should not start with `_`. `squidpy.experimental.im._sharpness_metrics`
is already private so `_tenengrad_mean` can be `tenengrad_mean`. Thats why in your IDE it might show `_tenengrad_mean` as unused function because the assumption is `_tenengrad_mean` is only meant to be used in the function it is defined.

src/squidpy/experimental/im/_qc_metrics.py
```diff
@@ -0,0 +1,98 @@
...
+from collections.abc import Callable
+from enum import StrEnum
+
+import numpy as np
+
+from squidpy.experimental.im._intensity_metrics import (
+    _brightness_mean,
+    _brightness_std,
+    _entropy,
```
selmanozleyen: Again, if `_entropy` is needed to import then no need to make it file/module private. `_intensity_metrics` is already private

src/squidpy/experimental/im/_sharpness_metrics.py
```diff
@@ -0,0 +1,97 @@
...
+    return np.array([[max(var_val, 0.0)]], dtype=np.float32)
+
+
+def _pop_variance(block: np.ndarray) -> np.ndarray:
+    """Population variance of pixel intensities."""
+    b = _to_f32_2d(block)
+    return np.array([[float(np.var(b))]], dtype=np.float32)
+
+
+@lru_cache(maxsize=4)
```
selmanozleyen: is there a reason why this is 4? at least can we have a comment here to justify the heuristic?

timtreis: Was an attempt at precaching frequently occuring masks to speed things up, but didn't really make a noteworthy difference

src/squidpy/experimental/im/_qc_metrics.py
timtreis: Changed 👍

src/squidpy/experimental/im/_intensity_metrics.py
```diff
@@ -0,0 +1,133 @@
...
+# Thread-local 1-entry cache: avoids re-computing HED deconvolution when
+# multiple HED metrics are called sequentially on the same dask block.
+# Uses threading.local so each dask worker thread gets its own cache,
+# and ctypes.data (the buffer's memory address) as the key so it remains
+# valid and unique while the array is alive.
+_hed_local = threading.local()
+
+
+def _rgb_to_hed_cached(block_rgb: np.ndarray) -> np.ndarray:
+    """Cached wrapper around ``_rgb_to_hed``."""
```
selmanozleyen:
```suggestion
    """Cached wrapper around ``rgb_to_hed``."""
```

src/squidpy/experimental/im/_intensity_metrics.py
```diff
@@ -0,0 +1,133 @@
...
+    rgb_clipped = np.clip(block_rgb, 0.0, 1.0)
+    return rgb2hed(rgb_clipped)
+
+
+# Thread-local 1-entry cache: avoids re-computing HED deconvolution when
+# multiple HED metrics are called sequentially on the same dask block.
+# Uses threading.local so each dask worker thread gets its own cache,
+# and ctypes.data (the buffer's memory address) as the key so it remains
+# valid and unique while the array is alive.
+_hed_local = threading.local()
```
selmanozleyen: Since these helpers aren't directly exposed I dont understand why we would need caching like this. This is a bit hacky. I made a suggestion as a PR. https://github.com/scverse/squidpy/pull/1152

selmanozleyen (approved):
all looks good except the points I made. We can put them as a TODO and move on if we want to

PR #1043, opened 2025-09-30
Replacing fixed size tuples as return types

selmanozleyen:
Fixes https://github.com/scverse/squidpy/issues/1038

### DIFF
--- a/src/squidpy/gr/_nhood.py
+++ b/src/squidpy/gr/_nhood.py
@@ -33,7 +33,7 @@ __all__ = ["nhood_enrichment", "centrality_scores", "interaction_matrix"]
 
 class NhoodEnrichmentResult(NamedTuple):
     """Result of nhood_enrichment function.
-    
+
     Attributes
     ----------
     zscore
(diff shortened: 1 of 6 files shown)


src/squidpy/gr/_nhood.py
```diff
@@ -30,7 +30,24 @@
...
+    Attributes
+    ----------
+    zscore : NDArrayA
+        Z-score values of enrichment statistic.
+    count : NDArrayA
+        Enrichment count.
+    """
+
+    zscore: np.ndarray[Any, np.dtype[Any]]
+    counts: np.ndarray[Any, np.dtype[Any]]  # NamedTuple inherits from tuple so cannot use 'count' as attribute name
```
flying-sheep: should use `NDArray[np.number]` (from `numpy.typing`) or so, the full type API isn’t public. Also what does `NDArrayA` in the docs mean?

selmanozleyen: It means `NDArray[Any]`. Fixed

PR #1044, opened 2025-09-30
Method to detect specimens in H&E images

timtreis:
For other downstream functions, such as https://github.com/scverse/squidpy/pull/1036, one needs a function to robustly detect where in the image the tissue is and how many of those there are.

This function
a) implements two algorithms for identifying the tissue (otsu & felzenszwalb)
b) deals with arbitrary channel input
c) heuristically tries to identify what is a sample and what is either just random stuff (dirt, Visium frame etc). As a fallback, one can pass in the number of samples expected which should be more robust
d) adds the mask back to the sdata object with the same structure and transformations as the original image had
e) does everything in dask so it's quite fast

```
sdata = sq.datasets.visium_hne_sdata()

sq.exp.im.detect_tissue(
    sdata,
    image_key="hne",
)

sdata
```
```
SpatialData object, with associated Zarr store: [/Users/tim.treis/.cache/squidpy/visium_hne_sdata.zarr](https://file+.vscode-resource.vscode-cdn.net/Users/tim.treis/.cache/squidpy/visium_hne_sdata.zarr)
├── Images
│     └── 'hne': DataTree[cyx] (3, 11757, 11291), (3, 5878, 5645), (3, 2939, 2822), (3, 1469, 1411)
├── Labels
│     └── 'hne_tissue': DataTree[yx] (11757, 11291), (5878, 5645), (2939, 2822), (1469, 1411)
├── Shapes
│     └── 'spots': GeoDataFrame shape: (2688, 2) (2D shapes)
└── Tables
      └── 'adata': AnnData (2688, 18078)
with coordinate systems:
    ▸ 'global', with elements:
        hne (Images), hne_tissue (Labels), spots (Shapes)
with the following elements not in the Zarr store:
    ▸ hne_tissue (Labels)
```

```
(
    sdata
    .pl.render_images("hne")
    .pl.render_labels("hne_tissue", fill_alpha=0, contour_px=10, outline_alpha=1)
    .pl.show()
)
```
image

### DIFF
--- a/src/squidpy/exp/im/_detect_tissue.py
+++ b/src/squidpy/exp/im/_detect_tissue.py
@@ -1,0 +1,546 @@
+from __future__ import annotations
+
+import enum
+from dataclasses import dataclass
+from typing import Any
+
+import numpy as np
+import spatialdata as sd
+import xarray as xr
+from dask_image.ndinterp import affine_transform as da_affine
+from scipy import ndimage
+from skimage import filters, measure
+from skimage.filters import gaussian, threshold_otsu
+from skimage.morphology import binary_closing, disk
+from skimage.segmentation import felzenszwalb
+from skimage.util import img_as_float
+from spatialdata._logging import logger as logg
+from spatialdata.models import Labels2DModel
+from spatialdata.transformations import get_transformation
+
+from squidpy._utils import _get_scale_factors, _yx_from_shape
+
+from ._utils import _flatten_channels, _get_image_data
+
+
+class DETECT_TISSUE_METHOD(enum.Enum):
+    OTSU = enum.auto()
+    FELZENSZWALB = enum.auto()
+
+
+@dataclass(slots=True)
+class BackgroundDetectionParams:
+    """
+    Which corners are background, and how large the corner boxes should be.
+    If no corners are flagged True, the orientation falls back to 'bright background'.
+    """
+
+    ymin_xmin_is_bg: bool = True
+    ymax_xmin_is_bg: bool = True
+    ymin_xmax_is_bg: bool = True
+    ymax_xmax_is_bg: bool = True
+    corner_size_pct: float = 0.01  # size of each corner box as fraction of img height/width
+
+    @property
+    def any_corner(self) -> bool:
+        return any((self.ymin_xmin_is_bg, self.ymax_xmin_is_bg, self.ymin_xmax_is_bg, self.ymax_xmax_is_bg))
+
+
+@dataclass(slots=True)
+class FelzenszwalbParams:
+    """
+    Size-aware superpixel defaults for felzenszwalb segmentation.
+    """
+
+    grid_rows: int = 100  # ~desired grid, only for scale heuristics
+    grid_cols: int = 100
+    sigma_frac: float = 0.008  # blur = this * short side (clipped to [1,5] px)
+    scale_coef: float = 0.25  # scale = coef * target_area
+    min_size_coef: float = 0.20  # min_size = coef * target_area
+
+
+def detect_tissue(
+    sdata: sd.SpatialData,
+    image_key: str,
+    scale: str = "auto",
+    method: DETECT_TISSUE_METHOD | str = DETECT_TISSUE_METHOD.FELZENSZWALB,
+    corners_are_background: bool = True,
+    mask_smoothing_cycles: int = 0,
+    new_labels_key: str | None = None,
+    inplace: bool = True,
+    **kwargs: Any,
+) -> xr.DataArray:
+    """
+    Detect tissue and return a boolean mask (y,x) where True = specimen.
+
+    Parameters
+    ----------
+    sdata : sd.SpatialData
+        SpatialData object containing the image
+    image_key : str
+        Key of the image in sdata.images
+    scale : str, default "auto"
+        - If a specific scale key is provided (e.g., "scale0", "scale1", ...),
+          that image scale is used verbatim.
+        - If "auto": pick the smallest available scale. If that smallest scale
+          exceeds the pixel threshold, it is further downscaled to be under the
+          threshold for calculations.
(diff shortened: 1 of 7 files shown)


src/squidpy/experimental/__init__.py
selmanozleyen, line 1: I'd actually prefer calling this module `experimental` instead of `exp` (also similar to `{anndata,scanpy}.experimental` module). Because I was wondering what it was until I saw this.

src/squidpy/exp/im/_detect_tissue.py
```diff
@@ -0,0 +1,562 @@
...
+    mask_smoothing_cycles : int, default 0
+        Number of cycles of (2D) morphological closing to apply to the mask
+    new_labels_key : str | None, default None
+        Key to store the new labels in the SpatialData object
+    inplace: bool, default True
+        Whether to store the new labels in the SpatialData object or return the mask.
+        If the mask is saved to the SpatialData object, it will inherit the scale_factors
+        of the image, if present.
+    **kwargs
+        Optional keyword arguments:
```
selmanozleyen: I am strongly against using kwargs. I can easily imagine the user mistyping `felzenszwalb_params` and wondering why it doesn't change the results :D . We need errors for cases like these.

selmanozleyen (commented):
hi these are some initial feedbacks I will get into more details tomorrow

src/squidpy/exp/im/_detect_tissue.py
flying-sheep: Agreed, makes no sense to arbitrarily nest them in the docs and collect them into a dict in the code.

I think having advanced parameters tucked away in the data class signatures the way it is is ideal already.

“Optional keyword arguments” are keyword arguments that have a default, so often all of them.

src/squidpy/exp/im/_detect_tissue.py
```diff
@@ -0,0 +1,562 @@
...
+    grid_cols: int = 100
+    sigma_frac: float = 0.008  # blur = this * short side (clipped to [1,5] px)
+    scale_coef: float = 0.25  # scale = coef * target_area
+    min_size_coef: float = 0.20  # min_size = coef * target_area
+
+
+def detect_tissue(
+    sdata: sd.SpatialData,
+    image_key: str,
+    scale: str = "auto",
```
flying-sheep: Always use a `*` if a function signature that has more than like 2 or 3 arguments total. Usually you can just put it after the parameters that have no default, in some cases it makes sense that the first optional parameter can be specified by keyword or position.

```suggestion
    sdata: sd.SpatialData,
    image_key: str,
    *,
    scale: str = "auto",
```

I’d personally do this to make `sdata` a mandatory positional parameter, but people can still do `image_key="..."` if they want:

```suggestion
    sdata: sd.SpatialData,
    /,
    image_key: str,
    *,
    scale: str = "auto",
```

hatch.toml
```diff
@@ -40,7 +40,8 @@ extra-dependencies = [
     "pytest-timeout",
 ]
 
-
+[envs.hatch-test.scripts]
+download = "python ./.scripts/ci/download_data.py {args}"
```
timtreis, line 35: @selmanozleyen not sure why this wasn't failing before when you merged the hatch PR, I assume you were hitting a cache that expired since then

hatch.toml
```diff
@@ -11,6 +11,7 @@ extra-dependencies = [
 [envs.coverage.scripts]
```
flying-sheep: why does this exist? you should just use `hatch test --cov` (i.e. use the `run-cov` script from hatch), not a dedicated environment

hatch.toml
```diff
@@ -39,7 +39,14 @@ extra-dependencies = [
...
 ]
-
+[envs.hatch-test.scripts]
+run = "pytest {args}"
+run-cov = "pytest --cov --cov-report=term {args}"
+cov-erase = "coverage erase"
+cov-report = "coverage report --omit='tox/*'"
+cov-combine = "coverage combine"
+cov-xml = "coverage xml --omit='tox/*' -o coverage.xml"
+download = "python ./.scripts/ci/download_data.py {args}"
```
flying-sheep, line 35: don’t define this table if possible. [the defaults](https://hatch.pypa.io/latest/config/internal/testing/#scripts) are there for a reason (e.g. `HATCH_TEST_ARGS` is there for a reason), and there’s no reason to have the [`omit` here instead of in `[tool.coverage]`](https://coverage.readthedocs.io/en/7.11.0/config.html#run-omit), right?

flying-sheep (commented):
there were no commands missing, hatch pre-defines them: https://hatch.pypa.io/latest/config/internal/testing/#scripts

unfortunately adding the `diff-cover` command means you have to re-define everything, since you can’t just override individual scripts in an environment (known hatch issue).

To explain how things should work:

- all test/coverage dependencies should be in one spot (here the `test` extra)
- the `hatch-test` env should define the commands as explained in the link above:
  - `run` should run the tests using `pytest ...`
  - `run-cov` should run the tests using `coverage run -m pytest ...`
  - `cov-combine` should combine the `.coverage-xyz` files written from the subprocesses spawned by `pytest-xdist`
  - `cov-report` should do all the reporting
- locally, you should use `hatch test [-c] [-p] ...` which will use these commands behind the scenes
  - VS Code only understands `pytest-cov`, which is therefore *only* needed for VS Code’s test running GUI.
- in CI, we sadly need to use `matrix_name`, so we can’t use the nice `hatch test` command and need to manually use the `run*` and `cov*` commands

hatch.toml
selmanozleyen: wdym? when I run `hatch run data:download --dry-run` locally it works

selmanozleyen:
True, there wasn't a CODECOV token because we now take it env variables from the gh repo settings and I just used mine. Thought it would use the orgs since I was part of it. Nice catch

PR #1045, opened 2025-10-01
Fix: sepal numba compilation options changes the results significantly

selmanozleyen:
Fixes: https://github.com/scverse/squidpy/issues/1035

### DIFF
--- a/src/squidpy/gr/_sepal.py
+++ b/src/squidpy/gr/_sepal.py
@@ -290,8 +290,11 @@ def _entropy(
     """Get entropy of an array."""
     xnz = xx[xx > 0]
     xs: np.float64 = np.sum(xnz)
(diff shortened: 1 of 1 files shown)


src/squidpy/gr/_sepal.py
```diff
@@ -290,8 +281,11 @@ def _entropy(
     """Get entropy of an array."""
```
flying-sheep: This is what that does right?

```suggestion
    """Compute Shannon entropy of an array of probability values (in nats)."""
```

Nats instead of Shannons, because you use the natural logarithm instead of log2.

Maybe also quote why the result is 0 if the sum is 0? https://stats.stackexchange.com/a/433096

flying-sheep (changes requested):

PR #1053, opened 2025-10-29
Add helper functions to tile image for inference

timtreis:
```
import squidpy as sq

sdata = sq.datasets.visium_hne_sdata()
sq.experimental.im.make_tiles(sdata, image_key="hne", preview=True)
```
image

```
import squidpy as sq

sdata = sq.datasets.visium_hne_sdata()
sq.experimental.im.make_tiles_from_spots(
    sdata,
    spots_key="spots",
    image_key="hne",
    tissue_mask_key="hne_tissue",
	preview=True,
)
```
image

### DIFF
--- a/src/squidpy/experimental/im/_detect_tissue.py
+++ b/src/squidpy/experimental/im/_detect_tissue.py
@@ -24,7 +24,7 @@ from squidpy._utils import _ensure_dim_order, _get_scale_factors, _yx_from_shape
 from ._utils import _flatten_channels, _get_element_data
 
 
-class DETECT_TISSUE_METHOD(enum.Enum):
+class DetectTissueMethod(enum.Enum):
     OTSU = enum.auto()
     FELZENSZWALB = enum.auto()
 
@@ -72,7 +72,7 @@ def detect_tissue(
     image_key: str,
     *,
     scale: str = "auto",
-    method: DETECT_TISSUE_METHOD | str = DETECT_TISSUE_METHOD.OTSU,
+    method: DetectTissueMethod | str = DetectTissueMethod.OTSU,
     channel_format: Literal["infer", "rgb", "rgba", "multichannel"] = "infer",
     background_detection_params: BackgroundDetectionParams | None = None,
     corners_are_background: bool = True,
@@ -100,8 +100,8 @@ def detect_tissue(
     method
         Tissue detection method. Valid options are:
 
-            - `DETECT_TISSUE_METHOD.OTSU` or `"otsu"` - Otsu thresholding with background detection.
-            - `DETECT_TISSUE_METHOD.FELZENSZWALB` or `"felzenszwalb"` - Felzenszwalb superpixel segmentation.
+            - `DetectTissueMethod.OTSU` or `"otsu"` - Otsu thresholding with background detection.
+            - `DetectTissueMethod.FELZENSZWALB` or `"felzenszwalb"` - Felzenszwalb superpixel segmentation.
 
     channel_format
         Expected format of image channels. Valid options are:
@@ -157,7 +157,7 @@ def detect_tissue(
     # Normalize method
     if isinstance(method, str):
         try:
-            method = DETECT_TISSUE_METHOD[method.upper()]
+            method = DetectTissueMethod[method.upper()]
         except KeyError as e:
             raise ValueError('method must be "otsu" or "felzenszwalb"') from e
 
@@ -190,7 +190,7 @@ def detect_tissue(
         img_grey = img_grey_da.values  # may compute
 
     # First-pass foreground
-    if method == DETECT_TISSUE_METHOD.OTSU:
+    if method == DetectTissueMethod.OTSU:
         img_fg_mask_bool = _segment_otsu(img_grey=img_grey, params=bgp)
     else:
         p = felzenszwalb_params or FelzenszwalbParams()


src/squidpy/experimental/im/_make_tiles.py
```diff
@@ -0,0 +1,887 @@
...
+from spatialdata.transformations import get_transformation, set_transformation
+
+from squidpy._utils import _yx_from_shape
+
+from ._utils import _get_element_data
+
+__all__ = ["TileGrid", "make_tiles", "make_tiles_from_spots"]
+
+
+class TileGrid:
```
Zethson: If this class is userfacing it should probably have a docstring

src/squidpy/experimental/im/_make_tiles.py
```diff
@@ -0,0 +1,887 @@
...
+        # We need tiles from min(0, offset_y) to at least H
+        # So total coverage needed is from min(0, offset_y) to H
+        grid_start_y = min(0, self.offset_y)
+        grid_start_x = min(0, self.offset_x)
+        total_h_needed = self.H - grid_start_y
+        total_w_needed = self.W - grid_start_x
+        self.tiles_y = (total_h_needed + self.ty - 1) // self.ty
+        self.tiles_x = (total_w_needed + self.tx - 1) // self.tx
+
+    def indices(self) -> np.ndarray:
```
Zethson, line 65: should this or names be a property? Nit: should you use a cache for these attributes to not always reculate/create a new array?

src/squidpy/experimental/im/_make_tiles.py
```diff
@@ -0,0 +1,887 @@
...
+        return np.array([[iy, ix] for iy in range(self.tiles_y) for ix in range(self.tiles_x)], dtype=int)
+
+    def names(self) -> list[str]:
+        return [f"tile_x{ix}_y{iy}" for iy in range(self.tiles_y) for ix in range(self.tiles_x)]
+
+    def bounds(self) -> np.ndarray:
```
Zethson, line 71: See comment above but I think here it might be more useful. If this doesn't change a lot and users access these regularly, you should consider using a cache for this and/or storing it as a property.

Same for centroids below

src/squidpy/experimental/im/_make_tiles.py
```diff
@@ -0,0 +1,887 @@
...
+            x1 = ((ix + 1) * self.tx + self.offset_x) if ix < self.tiles_x - 1 else self.W
+            # Clamp bounds to image dimensions
+            y0 = max(0, min(y0, self.H))
+            x0 = max(0, min(x0, self.W))
+            y1 = max(0, min(y1, self.H))
+            x1 = max(0, min(x1, self.W))
+            b.append([y0, x0, y1, x1])
+        return np.array(b, dtype=int)
+
+    def centroids_and_polygons(self) -> tuple[np.ndarray, list[Polygon]]:
```
Zethson, line 89: It might be better to have this not return a tuple but a dataclass or NamedTuple

src/squidpy/experimental/im/_make_tiles.py
```diff
@@ -0,0 +1,887 @@
...
+        pad_x = self.tiles_x * self.tx - int(arr_yx.shape[1])
+        a = arr_yx.rechunk((self.ty, self.tx))
+        return da.pad(a, ((0, pad_y), (0, pad_x)), mode="edge") if (pad_y > 0 or pad_x > 0) else a
+
+    def coarsen(self, arr_yx: da.Array, reduce: Literal["mean", "sum"] = "mean") -> da.Array:
+        reducer = np.mean if reduce == "mean" else np.sum
+        return da.coarsen(reducer, arr_yx, {0: self.ty, 1: self.tx}, trim_excess=False)
+
+
+class _SpotTileGrid:
```
Zethson, line 115: My comments from above also apply to this class and any additional class

src/squidpy/experimental/im/_make_tiles.py
```diff
@@ -0,0 +1,887 @@
...
+        cents: list[list[float]] = []
+        for y0, x0, y1, x1 in bounds:
+            cy = (y0 + y1) / 2.0
+            cx = (x0 + x1) / 2.0
+            cents.append([cy, cx])
+            polys.append(Polygon([(x0, y0), (x1, y0), (x1, y1), (x0, y1), (x0, y0)]))
+        return np.asarray(cents, dtype=float), polys
+
+
+def _get_largest_scale_dimensions(
```
Zethson, line 168: I don't understand why this function got a docstring but the others didn't

src/squidpy/experimental/im/_make_tiles.py
```diff
@@ -0,0 +1,887 @@
...
+    if "y" in img_da.sizes and "x" in img_da.sizes:
+        return int(img_da.sizes["y"]), int(img_da.sizes["x"])
+    else:
+        # Fallback: assume last two dimensions are spatial
+        return int(img_da.shape[-2]), int(img_da.shape[-1])
+
+
+def _choose_label_scale_for_image(label_node: Labels2DModel, target_hw: tuple[int, int]) -> str:
+    if not hasattr(label_node, "keys"):
+        return "scale0"
```
Zethson: Not quite clear to me why this is the default value. A single line comment might help

src/squidpy/experimental/im/_make_tiles.py
```diff
@@ -0,0 +1,887 @@
...
+    min_tissue_fraction: float = 1.0,
+    new_shapes_key: str | None = None,
+    preview: bool = True,
+) -> None:
+    """
+    Create and filter tiles suitable for inference.
+
+    This function combines tile grid generation and filtering based on tissue coverage.
+    It creates a tile grid, filters tiles based on tissue criteria, and optionally
+    generates a preview plot.
```
Zethson: put this into the line above so that it's more easier readable. It's not too wide

src/squidpy/experimental/im/_make_tiles.py
```diff
@@ -0,0 +1,887 @@
...
+    scale
+        Scale level to use for mask processing. The tile grid is always generated
+        based on the largest (finest) scale of the image.
+    min_tissue_fraction
+        Minimum fraction of tile that must be in tissue. Default is 1.0 (exclusively in tissue).
+    new_shapes_key
+        Key to save the tile grid in ``sdata.shapes``. If ``None``, defaults to ``f"{image_key}_tiles"``.
+    preview
+        If ``True``, generate a preview plot showing the tile grid colored by classification.
+        Default is ``True``.
```
Zethson:
```suggestion
```

src/squidpy/experimental/im/_make_tiles.py
```diff
@@ -0,0 +1,887 @@
...
+        return None
+    if not transformations:
+        return None
+    # Prefer 'global' if present
+    if "global" in transformations:
+        return "global"
+    return next(iter(transformations.keys()))
+
+
+def _derive_tile_size_from_spots(coords: np.ndarray) -> tuple[int, int]:
```
Zethson, line 854: There's a few error messages here that don't suggest a solution. Always suggest one if possible

Zethson (commented):
A few random comments as requested

src/squidpy/experimental/im/_make_tiles.py
timtreis: Isn't user facing, just a helper

src/squidpy/experimental/im/_make_tiles.py
```diff
@@ -0,0 +1,902 @@
...
+    spots_key: str,
+    image_key: str | None = None,
+    tissue_mask_key: str | None = None,
+    scale: str = "auto",
+    min_tissue_fraction: float = 1.0,
+    new_shapes_key: str | None = None,
+    preview: bool = True,
+) -> None:
+    """
+    Create tiles centered on Visium spots, classify them by tissue coverage, and optionally render a preview.
```
LLehner, line 468: Is it necessary to restrict this to Visium spots? This could also work for arbitrary segmentations as long as cell coordinates + e.g radius are given. It definitely works on custom cell circles.

src/squidpy/experimental/im/_make_tiles.py
```diff
@@ -0,0 +1,902 @@
...
+    # Generate preview plot if requested
+    if preview and shapes_key in sdata.shapes:
+        try:
+            (
+                sdata.pl.render_images(image_key)
+                .pl.render_shapes(shapes_key, color="tile_classification", fill_alpha=0.5)
+                .pl.show()
+            )
+        except (AttributeError, KeyError, ValueError) as e:
+            logger.warning(f"Could not generate preview plot: {e}")
```
LLehner, line 453: For some reason this throws a warning: ```WARNING  Could not generate preview plot: 'SpatialData' object has no attribute 'pl'``` when testing with some random data not from ```sq.datasets```

Not sure if the problem is on my side since packages should be up to date:
spatialdata=0.6.1
spatialdata-plot=0.2.13

LLehner (commented):
Looks good! Just wondering about the plot warning, but maybe thats on my side.

src/squidpy/experimental/im/_make_tiles.py
selmanozleyen: I think you need to import spatialdata-plot to fix that

src/squidpy/experimental/im/_make_tiles.py
```diff
@@ -0,0 +1,902 @@
...
+    # Convert to numpy array if needed
+    if hasattr(mask_da, "compute"):
+        mask = np.asarray(mask_da.compute())
+    elif hasattr(mask_da, "values"):
+        mask = np.asarray(mask_da.values)
+    else:
+        mask = np.asarray(mask_da)
+
+    # Ensure 2D (y, x) shape
+    if mask.ndim > 2:
```
selmanozleyen: this is in case it is like 1,x,y or x,y,1 right? I would just throw an error for this case. Giving a specific error message to suggest calling squeeze is more valuable imo. Otherwise the output might confuse users that gave something like x,y,2.

The error would also confuse people who gave 4d input and the code says you have shape of 3d because you squezed earlier

src/squidpy/experimental/im/_make_tiles.py
```diff
@@ -0,0 +1,902 @@
...
+
+def _get_mask_from_labels(sdata: sd.SpatialData, mask_key: str, scale: str) -> np.ndarray:
+    """Extract a 2D mask array from ``sdata.labels`` at the requested scale."""
+    if mask_key not in sdata.labels:
+        raise KeyError(f"Mask key '{mask_key}' not found in sdata.labels")
+
+    label_node = sdata.labels[mask_key]
+    mask_da = _get_element_data(label_node, scale, "label", mask_key)
+
+    if hasattr(mask_da, "compute"):
```
selmanozleyen: Isaac would give me advice to not use duck-typing often back in my HiWi days.

I am not sure if you expect xarray or dask-array but this should work for both

```python
from dask.base import is_dask_collection

if is_dask_collection(mask_da):
    # Handles dask arrays, dask dataframes, and xarray objects wrapping dask
    mask_da.compute()
```

I also think you should do the same with `elif hasattr(mask_da, "values"):`. Because it makes it very hard to read the code also.

src/squidpy/experimental/im/_utils.py
```diff
@@ -1,67 +1,63 @@
...
 
-from typing import Literal
+from typing import Any, Literal
 
-import spatialdata as sd
 import xarray as xr
-from spatialdata._logging import logger as logg
+from spatialdata._logging import logger
 
+__all__ = ["_get_element_data", "_flatten_channels"]
```
selmanozleyen: I think this is a repo-wide thing I'd like to discuss but why make the functions private if you are going to expose them? Private `_get_element_data` usually suggests it should only be used in that file. Some IDE's even warn interpret those functions as unused even if they are used outside but not within the files they are defined in.

From pep08:
```
_single_leading_underscore: weak “internal use” indicator. E.g. from M import * does not import objects whose names start with an underscore.
```

`_utils` is already private and that is enough to mark those functions for internal use only.

selmanozleyen (commented):
Besides my comments I checked the code coverage and it suggested that some cases weren't tested.
-      if H_mask != H or W_mask != W: # (line 803)
I am not sure if its accurate but overall I'd check the coverage report to ensure robustness for some cases.

src/squidpy/experimental/im/_make_tiles.py
LLehner: > I think you need to import spatialdata-plot to fix that

true, now it works. Thanks! @selmanozleyen

src/squidpy/experimental/im/_utils.py
timtreis: Agree

src/squidpy/experimental/im/_make_tiles.py
timtreis: > Is it necessary to restrict this to Visium spots? This could also work for arbitrary segmentations 

I think yes, because it generates non-overlapping tiles. Yes, you can generate them centered around cell segmentation masks but then the tiles would overlap which would probably create interpretability issues. So I'd semantically restrict it to the clean usecase - what users then actually do with it is their issue imho.

> I think you need to import spatialdata-plot to fix that

wdym @selmanozleyen ?

src/squidpy/experimental/im/_make_tiles.py
timtreis: > this is in case it is like 1,x,y or x,y,1 right? 

It'd always be (1, y, x) since SpatialData canonically has `cyx` as a format. 

> I would just throw an error for this case. 

Disagree, here we have the chance to, without guessing, correct a potentially weird user input. I'll add a warn statement informing the user that it's canonically wrong though and that we squeezed away the dim

> The error would also confuse people who gave 4d input and the code says you have shape of 3d because you squezed earlier

We don't really support 4D upstream yet so I'm not yet super worried.

src/squidpy/experimental/im/_make_tiles.py
timtreis: Good point, changed it

PR #1065, opened 2025-11-20
update var by dist normalization

LLehner:
Distances to anchor point are now actually normalized for each slide respectively.

### DIFF
--- a/src/squidpy/tl/_var_by_distance.py
+++ b/src/squidpy/tl/_var_by_distance.py
@@ -146,12 +146,11 @@ def var_by_distance(
         else:
             df["obs"] = adata.obs_names
 
-        # store dataframes by (slide, anchor) combination and also the corresponding maximum distance for normalization
+        # store dataframes by (slide, anchor) combination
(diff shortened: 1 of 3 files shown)


src/squidpy/tl/_var_by_distance.py
```diff
@@ -146,12 +144,11 @@ def var_by_distance(
         else:
             df["obs"] = adata.obs_names
 
-        # store dataframes by (slide, anchor) combination and also the corresponding maximum distance for normalization
+        # store dataframes by (slide, anchor) combinationq
```
timtreis: typo

LLehner:
> Can you describe the issue you're solving here better and then also translate this into a test?

Previous normalization was useless as it just changed the the distance values to be within [0,1] without actually changing the scale. Now the distances from categories with a lower maximum are stretched to be comparable. This doesn't affect the raw distances which are still provided when running this function. I will add a test.

timtreis (approved):

PR #1074, opened 2025-12-03
Added more complex tissue detection method for complexer tissues

timtreis:
Segmenting this image is hard:

image

Otsu f.e. yields
image
since we have very dark regions in the image with background-esque areas inbetween.

The new method, inspired by [this](https://scikit-image.org/docs/0.25.x/auto_examples/segmentation/plot_trainable_segmentation.html) is based on the assumption that the corners are background, then uses randomforest to classify the tissue into bg and fg and finally refines the mask.

```
sq.experimental.im.detect_tissue(
    sdata,
    image_key="hne",
    method="weka",
    n_samples=2,
)
```
image

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -382,7 +382,8 @@ def sdata_mask_graph():
             "table": sd.models.TableModel().parse(adata),
         }
     )
-    
+
+
 @pytest.fixture()
 def sdata_hne():
     return sq.datasets.visium_hne_sdata()


src/squidpy/experimental/im/_detect_tissue.py
```diff
@@ -67,23 +71,158 @@ class FelzenszwalbParams:
...
+        mask[:, -r:] = False
+    return mask
+
+
+def _apply_border_margin(mask: np.ndarray, margins: tuple[int, int, int, int]) -> np.ndarray:
+    """
+    Zero-out a border defined by (top, bottom, left, right) around a boolean mask.
+    """
+    t, b, l, r = margins
+    if t == b == l == r == 0:
```
selmanozleyen: I checked and this works as expected in python. I didn't personally like it because in theory if r was undefined and t was 5 that section of code wouldn't error because it would short-circuit, which is fine in many cases. But I personally would just prefer margins == (0,0,0,0) for clearer code. Because in other languages `t == b == l == r == 0` could mean `(((t == b) == l) == r) == 0` or something like that

src/squidpy/experimental/im/_detect_tissue.py
```diff
@@ -67,23 +71,158 @@ class FelzenszwalbParams:
...
+    out = np.array(mask, copy=True)
+    if t > 0:
+        out[:t, :] = False
+    if b > 0:
+        out[-b:, :] = False
+    if l > 0:
+        out[:, :l] = False
+    if r > 0:
+        out[:, -r:] = False
+    return out
```
selmanozleyen, line 174: are all these cases reasonable? if so, are all these cases tested?

timtreis: This is really only the user manually overwriting part of the segmentation mask starting at the edges of the img. Not tested right now since it's fairly trivial but I can add it

PR #1076, opened 2025-12-06
Adding a registry to have the hashes of datasets (restructured for aws s3)

selmanozleyen:
again a continuation of https://github.com/scverse/squidpy/pull/1072 to add hashes and use the uploaded datasets

changes made

- Now we don't download in unit tests since it's costly for s3 and we know they should be active links since we host it
- Some datasets were using scanpy defaults and others where using DEFAULT_CACHE_DIR = '~/.cache/squidpy' now we do DEFAULT_CACHE_DIR because it is a global path
- updated .scripts/ci/download_data.py to download all datasets
- One unified DatasetDownloader class instead of multiple ways to download datasets in squidpy 
- Everything hard-coded like links and dataset names,links, hashes are in only datasets.yaml registry and when interfacing from registries e.g. `visium_hne_image = _make_image_loader("visium_hne_image")`
- Now we don't need to redownload everything when the script is updated. It only triggers the run of the download file but the downloaded files from old script remains

### DIFF
--- a/.github/workflows/test-notebooks.yaml
+++ b/.github/workflows/test-notebooks.yaml
@@ -11,7 +11,37 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  ensure-data-is-cached:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          filter: blob:none
+          fetch-depth: 0
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
+        with:
+          enable-cache: false
+          python-version: "3.13"
+
+      - name: Restore data cache
+        id: data-cache
+        uses: actions/cache@v4
+        with:
+          path: data # IMPORTANT: this will fail if scanpy.settings.datasetdir default changes
+          key: data-${{ hashFiles('**/download_data.py') }}
+          restore-keys: |
+            data-
+          enableCrossOsArchive: true
+
+      - name: Download datasets
+        # Always run to ensure any missing files are downloaded
+        # (restore-keys may provide partial cache)
+        run: uvx hatch run data:download
+
   test:
+    needs: [ensure-data-is-cached]
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
@@ -33,8 +63,20 @@ jobs:
           enable-cache: true
           python-version: ${{ matrix.python }}
           cache-dependency-glob: pyproject.toml
+
+      - name: Restore data cache
+        id: data-cache
+        uses: actions/cache@v4
+        with:
+          path: data # IMPORTANT: this will fail if scanpy.settings.datasetdir default changes
+          key: data-${{ hashFiles('**/download_data.py') }}
+          restore-keys: |
+            data-
+          enableCrossOsArchive: true
+
       - name: Create notebooks environment
         run: uvx hatch -v env create notebooks
+
       - name: Test notebooks
         env:
           MPLBACKEND: agg


.scripts/ci/download_data.py
```diff
@@ -1,77 +1,81 @@
...
+        "Visium_FFPE_Human_Breast_Cancer",
+    ]
 
     if args.dry_run:
-        for func_name, ext in zip(all_datasets, all_extensions):
-            if func_name == "visium_hne_sdata":
-                ext = "zarr"
-            path = _ROOT / f"{func_name}.{ext}"
-            _print_message(func_name, path, dry_run=True)
+        print("\nWould download:")
```
timtreis: Doesn't logging play nicer with the GH runners?

src/squidpy/datasets/__init__.py
```diff
@@ -1,7 +1,55 @@
...
+    "visium_fluo_adata",
+    "visium_fluo_adata_crop",
+    "sc_mouse_cortex",
+    "mibitof",
+    "merfish",
+    "slideseqv2",
+    # Image datasets
+    "visium_fluo_image_crop",
+    "visium_hne_image_crop",
+    "visium_hne_image",
```
timtreis, line 54: nit, but don't fully understand the comments here? Everything visium_x is f.e. "10x Genomics". Either remove or more semantically useful split

src/squidpy/datasets/_datasets.py
```diff
@@ -0,0 +1,278 @@
...
+from squidpy.datasets._registry import get_registry
+from squidpy.read._utils import PathLike
+
+if TYPE_CHECKING:
+    import spatialdata as sd
+    from anndata import AnnData
+
+    from squidpy.im import ImageContainer
+
+__all__ = [
```
timtreis: Why is this duplicated with __init__? Will probably get out of sync, can we have 1 source of truth?

src/squidpy/datasets/_datasets.py
```diff
@@ -0,0 +1,278 @@
...
+    "visium_hne_image",
+]
+
+
+# =============================================================================
+# Hardcoded dataset name types
+# =============================================================================
+
+# 10x Genomics Visium datasets (adata_with_image type)
+VisiumDatasets = Literal[
```
timtreis, line 29: Where is this list coming from? Are we loading all of them?

src/squidpy/datasets/_datasets.py
```diff
@@ -0,0 +1,278 @@
...
+def visium(
+    sample_id: VisiumDatasets,
+    *,
+    include_hires_tiff: bool = False,
+    base_dir: PathLike | None = None,
+) -> AnnData:
+    """
+    Download Visium `datasets <https://support.10xgenomics.com/spatial-gene-expression/datasets>`_ from *10x Genomics*.
+
+    Uses the unified downloader which supports S3 with fallback to 10x Genomics servers.
```
timtreis: I'm not sure if we should have that fallback - the reason for S3 was control over the data existence. Feels overengineered here? I don't expect AWS to fail much

src/squidpy/datasets/_datasets.py
```diff
@@ -0,0 +1,278 @@
...
+
+    Parameters
+    ----------
+    sample_id
+        Name of the Visium dataset.
+    include_hires_tiff
+        Whether to download the high-resolution tissue section into
+        :attr:`anndata.AnnData.uns` ``['spatial']['{sample_id}']['metadata']['source_image_path']``.
+    base_dir
+        Directory where to download the data. If `None`, uses ~/.cache/squidpy/visium.
```
timtreis: why specifically the `visium` folder? Wouldn't do that

src/squidpy/datasets/_datasets.py
```diff
@@ -0,0 +1,278 @@
...
+
+    downloader = get_downloader()
+    return downloader.download(sample_id, base_dir, include_hires_tiff=include_hires_tiff)
+
+
+def visium_hne_sdata(folderpath: Path | str | None = None) -> sd.SpatialData:
+    """
+    Download a Visium H&E dataset as a SpatialData object.
+
+    Uses the unified downloader which supports S3 with fallback to Figshare.
```
timtreis: Would also remove that fallback.

src/squidpy/datasets/_datasets.py
```diff
@@ -0,0 +1,278 @@
...
+
+    Parameters
+    ----------
+    folderpath
+        A folder path where the dataset will be downloaded and extracted.
+        If `None`, uses the default cache directory (~/.cache/squidpy).
```
timtreis: Here the data gets downloaded straight to `cache`, in the other one into a subfolder. Should be homogenous (ideally without subfolders)

src/squidpy/datasets/_downloader.py
```diff
@@ -0,0 +1,312 @@
...
+
+    from squidpy.im import ImageContainer
+
+__all__ = ["DatasetDownloader", "download", "get_downloader"]
+
+# Default cache directory
+DEFAULT_CACHE_DIR = Path.home() / ".cache" / "squidpy"
+
+# Browser-like User-Agent to avoid 403 errors from some servers
+USER_AGENT = "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36"
```
timtreis: From experience, that's very easy to block for companies and many do. I assume you need it for the fallback?

src/squidpy/datasets/_downloader.py
```diff
@@ -0,0 +1,312 @@
...
+        img = ImageContainer()
+        img.add_img(local_path, layer="image", library_id=entry.library_id, **kwargs)
+
+        return img
+
+    def _load_spatialdata(
+        self,
+        entry: DatasetEntry,
+        path: Path | str | None = None,
+    ) -> Any:  # Returns SpatialData
```
timtreis: Why are you returning Any but saying SpatialData?

src/squidpy/datasets/_downloader.py
```diff
@@ -0,0 +1,312 @@
...
+        with tarfile.open(spatial_path) as f:
+            for member in f:
+                target = sample_dir / member.name
+                if not target.exists():
+                    f.extract(member, sample_dir)
+
+        # Optionally download high-res image (can be tif, jpg, etc.)
+        source_image_path = None
+        if include_hires_tiff:
+            # Look for any image file (image.tif, image.jpg, etc.)
```
timtreis: I think we're only loading tiffs, no?

tests/datasets/test_downloader.py
```diff
@@ -0,0 +1,149 @@
...
+        from anndata import AnnData
+
+        downloader = DatasetDownloader(cache_dir=DEFAULT_CACHE_DIR)
+        adata = downloader.download("V1_Mouse_Kidney", include_hires_tiff=False)
+
+        assert isinstance(adata, AnnData)
+        assert "spatial" in adata.uns
+
+
+class TestFallbackUrls:
```
timtreis: If you remove fallback, also remove tests

.scripts/ci/download_data.py
```diff
@@ -1,77 +1,81 @@
...
-
-        path = _ROOT / f"{func_name}.{ext}"
-        _print_message(func_name, path)
-        obj = _maybe_download_data(func_name, path)
-
-        # we could do without the AnnData check as well (1 less req. in tox.ini), but it's better to be safe
-        assert isinstance(obj, AnnData | sq.im.ImageContainer), type(obj)
-        assert path.is_file(), path
+    # Download AnnData datasets - just call the function, it handles caching
+    print("\nDownloading AnnData datasets...")
```
ilan-gold:
```suggestion
    print("Downloading AnnData datasets...")
```
why do you need `\n`?

.scripts/ci/download_data.py
```diff
@@ -1,77 +1,81 @@
...
     from anndata import AnnData
 
     import squidpy as sq
+    from squidpy.datasets._downloader import DEFAULT_CACHE_DIR
+    from squidpy.datasets._registry import get_registry
 
-    all_datasets = sq.datasets._dataset.__all__ + sq.datasets._image.__all__
-    all_extensions = ["h5ad"] * len(sq.datasets._dataset.__all__) + ["tiff"] * len(sq.datasets._image.__all__)
+    registry = get_registry()
+    print(f"Cache directory: {DEFAULT_CACHE_DIR}")
```
ilan-gold: Why so many `print` statements?

src/squidpy/datasets/_downloader.py
selmanozleyen: nope, I had this one outlier for Visium_FFPE_Human_Normal_Prostate_image.jpg

src/squidpy/datasets/_downloader.py
```diff
@@ -0,0 +1,312 @@
...
+from __future__ import annotations
+
+import shutil
+import tarfile
+from functools import lru_cache
+from pathlib import Path
+from typing import TYPE_CHECKING, Any
+
+import pooch
+from scanpy import logging as logg
```
ilan-gold: Not part of the scanpy public API, please remove

src/squidpy/datasets/_downloader.py
```diff
@@ -0,0 +1,312 @@
...
+        **kwargs: Any,
+    ) -> AnnData:
+        """Download and load an AnnData dataset."""
+        import anndata
+
+        if not entry.files:
+            raise ValueError(f"Dataset {entry.name} has no files")
+
+        file_entry = entry.files[0]
+        if path:
```
ilan-gold:
```suggestion
        if path is not None:
```
not a fan of implicit-nes

src/squidpy/datasets/_datasets.py
selmanozleyen: thx forgot to update

src/squidpy/datasets/_datasets.py
ilan-gold: I agree, we could then get rid of that agent-spoofing

src/squidpy/datasets/_downloader.py
```diff
@@ -0,0 +1,312 @@
...
+            path = Path(path)
+            target_dir = path.parent
+            # Ensure proper extension is used (e.g., .tiff, .png)
+            expected_suffix = Path(file_entry.name).suffix
+            target_name = path.name if path.suffix else f"{path.name}{expected_suffix}"
+        else:
+            target_dir = self.cache_dir / "images"
+            target_name = file_entry.name
+
+        local_path = self._download_file(file_entry, target_dir, target_name)
```
ilan-gold: deduplicate with `_load_anndata`

.scripts/ci/download_data.py
selmanozleyen: I kind of overlooked this script. Should be better

src/squidpy/datasets/_downloader.py
selmanozleyen: ok but why is it not marked with an underscore then? our notebooks also use it

src/squidpy/datasets/_downloader.py
selmanozleyen: agree

src/squidpy/datasets/_datasets.py
selmanozleyen: true

src/squidpy/datasets/__init__.py
selmanozleyen: I tried to split it based on the old files. Do you have any suggestions

src/squidpy/datasets/_datasets.py
selmanozleyen: I wanted to save the original links basically. But I can get rid of them

src/squidpy/datasets/_datasets.py
selmanozleyen: It is from the old API (main branch). Didn't want to remove it because it was public. (under _10x.py)

src/squidpy/datasets/_datasets.py
selmanozleyen: The comment was wrong in that case. I only have subfolders when the dataset is multiple files

src/squidpy/datasets/_downloader.py
timtreis: We're also moving Squidpy to use the spatialdata logger in the other functions, feel free to use that one instead

src/squidpy/datasets/_downloader.py
```diff
@@ -0,0 +1,288 @@
...
+
+__all__ = [
+    "DatasetDownloader",
+    "download",
+    "get_downloader",
+    "DEFAULT_CACHE_DIR",
+]
+
+# Default cache directory
+DEFAULT_CACHE_DIR = Path.home() / ".cache" / "squidpy"
```
Zethson: Pertpy reuses the scanpy cache directory. Would this also be a good option here? Then it's more or less 1 setting for all of them.

selmanozleyen: But the scanpy cache directory isn't global right? I wanted to also use that but it created some issues

src/squidpy/datasets/_datasets.py
selmanozleyen: Won't remove this since this was used to guide the users. Also needed for backwards compat.
image

src/squidpy/datasets/_downloader.py
timtreis: I don't think we should do that, also in general for pertpy? If I know that I downloaded test datasets f.e. from Squidpy or Pertpy, I'd prefer that I can find them in the obvious place on disk and delete them if needed.

selmanozleyen: oky I actually agree. I don't even remember why I wanted it to be global tbh

src/squidpy/datasets/_downloader.py
selmanozleyen: created an issue about this to create it separate from this PR. https://github.com/scverse/squidpy/issues/1081

src/squidpy/datasets/_datasets.py
```diff
@@ -0,0 +1,249 @@
...
+        Name of the Visium dataset.
+    include_hires_tiff
+        Whether to download the high-resolution tissue section into
+        :attr:`anndata.AnnData.uns` ``['spatial']['{sample_id}']['metadata']['source_image_path']``.
+    base_dir
+        Directory where to download the data. If `None`, uses :attr:`scanpy.settings.datasetdir`.
+
+    Returns
+    -------
+    :class:`anndata.AnnData`
```
ilan-gold, line 122: I don't think you need to specify return class types with these types of docstrings.  Does this occur elsewhere in the codebase or is it necessary for rendering?

src/squidpy/datasets/_datasets.py
```diff
@@ -0,0 +1,249 @@
...
+    loader.__name__ = dataset_name
+    return loader
+
+
+# =============================================================================
+# Image dataset functions
+# =============================================================================
+
+
+def _make_image_loader(dataset_name: str):
```
ilan-gold: This looks nearly identical to `_make_anndata_loader` - please deduplicate

src/squidpy/datasets/_downloader.py
```diff
@@ -0,0 +1,286 @@
...
+        s3_base_url: str | None = None,
+    ):
+        self.cache_dir = Path(cache_dir) if cache_dir else Path(settings.datasetdir)
+        self.cache_dir.mkdir(parents=True, exist_ok=True)
+
+        self._registry = get_registry()
+        self._s3_base_url = s3_base_url if s3_base_url is not None else self._registry.s3_base_url
+
+    def _get_first_file(self, entry: DatasetEntry) -> FileEntry:
+        """Get and validate the first file entry."""
```
ilan-gold: This does no validation. Is that intentional?

src/squidpy/datasets/_downloader.py
```diff
@@ -0,0 +1,286 @@
...
+
+    def __init__(
+        self,
+        cache_dir: Path | str | None = None,
+        s3_base_url: str | None = None,
+    ):
+        self.cache_dir = Path(cache_dir) if cache_dir else Path(settings.datasetdir)
+        self.cache_dir.mkdir(parents=True, exist_ok=True)
+
+        self._registry = get_registry()
```
ilan-gold: If a downloader is attached to the output of `get_registry`, and only one instance of the registry is shared throughout the codebase, what is the point of calling `get_registry` in other locations? Why not just make `DatasetDownloader._registry` public and always use that?

.scripts/ci/download_data.py
```diff
@@ -1,77 +1,69 @@
...
-            continue
+    for name in registry.image_datasets:
+        obj = getattr(sq.datasets, name)()
+        assert isinstance(obj, sq.im.ImageContainer)
 
-        path = _ROOT / f"{func_name}.{ext}"
-        _print_message(func_name, path)
-        obj = _maybe_download_data(func_name, path)
+    for name in registry.spatialdata_datasets:
+        getattr(sq.datasets, name)()
```
ilan-gold, line 55: Why use `get_registry` if you can just use the constants you define in `datasets/__init__.py`?

src/squidpy/datasets/_datasets.py
```diff
@@ -0,0 +1,249 @@
...
+    -------
+    :class:`anndata.AnnData`
+        Spatial AnnData object.
+    """
+    # Validate sample_id against known names
+    registry = get_registry()
+    if sample_id not in registry:
+        msg = f"Unknown Visium sample: {sample_id}. "
+        msg += f"Available samples: {registry.visium_datasets}"
+        raise ValueError(msg)
```
ilan-gold: To clarify my point about `DatasetDownloader._registry`, I think that it would be simpler to do `if sample_id not in downloader.registry` - what if the registry were different in `get_registry` from the dataset downloader registry (perhaps because of a bug or race condition)?

src/squidpy/datasets/_downloader.py
```diff
@@ -0,0 +1,286 @@
...
+        zip_path = self._download_file(file_entry, folder)
+        logg.info(f"Extracting {zip_path} to {folder}")
+        shutil.unpack_archive(str(zip_path), folder)
+
+        if not zarr_path.exists():
+            raise RuntimeError(f"Expected extracted data at {zarr_path}, but not found")
+
+        return sd.read_zarr(zarr_path)
+
+    def _load_adata_with_image(
```
ilan-gold: I think this name is confusing (and docstring) - this looks specific to 10X's format for visium, so maybe we should name it as such?

selmanozleyen: ah I wasn't aware of that. So this is a general 10x thing? If so let me rename it. But if there are other companies that use similar format would it be confusing?

src/squidpy/datasets/_downloader.py
selmanozleyen: clarified docstring

src/squidpy/datasets/_datasets.py
selmanozleyen: yes it's this one https://squidpy.readthedocs.io/en/stable/api/squidpy.datasets.visium.html

src/squidpy/datasets/_downloader.py
selmanozleyen: I updated my response @ilan-gold, I realized my first response was a bit unorganized.

Completely valid questions. Technically as is there won't be race-conditions as they are both singlethons.

But I made `registry` public since it will be that downloaders registry we are interested in. And DatasetDownloader will require a registry in constructor to make it clearer.

src/squidpy/datasets/_datasets.py
selmanozleyen: sorry my bad

src/squidpy/datasets/_datasets.py
flying-sheep: It’s superfluous, but needs some Sphinx customization to get rid of, so out of scope for this PR.

src/squidpy/datasets/_downloader.py
```diff
@@ -0,0 +1,286 @@
...
+    s3_base_url
+        Base URL for S3 bucket. If None, uses the value from datasets.yaml.
+    """
+
+    def __init__(
+        self,
+        cache_dir: Path | str | None = None,
+        s3_base_url: str | None = None,
+    ):
+        self.cache_dir = Path(cache_dir) if cache_dir else Path(settings.datasetdir)
```
flying-sheep:
```suggestion
        self.cache_dir = Path(cache_dir or settings.datasetdir)
```

src/squidpy/datasets/_downloader.py
flying-sheep: If other companies lay out their files exactly like that, then to be 10x compatible.

If they and 10x end up making a joint standard, we can name this function after that standard, otherwise it exists to load 10x data, right?

src/squidpy/datasets/_downloader.py
```diff
@@ -0,0 +1,286 @@
...
+                    fname=filename,
+                    path=str(target_dir),
+                    progressbar=True,
+                )
+                return Path(downloaded)
+            except (OSError, ValueError, RuntimeError) as e:
+                last_error = e
+                logg.warning(f"Failed to download from {url}: {e}")
+
+        raise RuntimeError(f"Failed to download {filename}. Last error: {last_error}")
```
flying-sheep: raise an ExceptionGroup instead, with all exceptions inside. That’s what it’s for!

Check the docs how to catch errors from ExceptionGroups so you know.

src/squidpy/datasets/_downloader.py
```diff
@@ -0,0 +1,286 @@
...
+    def _load_anndata(
+        self,
+        entry: DatasetEntry,
+        path: Path | str | None = None,
+        **kwargs: Any,
+    ) -> AnnData:
+        """Download and load an AnnData dataset."""
+        import anndata
+
+        file_entry = self._get_first_file(entry)
```
flying-sheep: could probably remove that helper and just do

```suggestion
        [file_entry] = entry.files
```

you’d get a `ValueError: not enough values to unpack`, but that wouldn’t happen anyway right?

.scripts/ci/download_data.py
selmanozleyen: Those are `Literal` type annotations for the IDE, docs and the users. I am not sure we should use them as our source of information in a code.

src/squidpy/datasets/_registry.py
```diff
@@ -0,0 +1,219 @@
...
+        return None
+
+    def get_file_by_suffix(self, suffix: str) -> FileEntry | None:
+        """Get a file by suffix (e.g., 'filtered_feature_bc_matrix.h5')."""
+        for f in self.files:
+            if f.name.endswith(suffix):
+                return f
+        return None
+
+    def get_file_by_prefix(self, prefix: str) -> FileEntry | None:
```
flying-sheep: Can files be more deeply nested on toplevel? Then please call this `get_file_by_name_prefix`, otherwise it’s ambiguous if it means a prefix of the relative path or the name.

src/squidpy/datasets/_registry.py
```diff
@@ -0,0 +1,219 @@
...
+
+__all__ = ["DatasetType", "FileEntry", "DatasetEntry", "DatasetRegistry", "get_registry"]
+
+
+def _get_config_path() -> Path:
+    """Get the path to datasets.yaml using importlib.resources for robustness."""
+    # Use importlib.resources for robust path resolution across different installation methods
+    # (editable installs, zip imports, etc.)
+    files = importlib.resources.files("squidpy.datasets")
+    return Path(str(files.joinpath("datasets.yaml")))
```
flying-sheep: This doesn’t make sense, there’s a reason why `files` returns a [`Traversable`](https://docs.python.org/3/library/importlib.resources.abc.html#importlib.resources.abc.Traversable) and not a local filesystem `Path`, so don’t try turning it into one. Return the `Traversable` instead and use its methods exclusively (they’re a subset of `Path`’s).

You can even use `as_file` to get a Path in a supported way, but I don’t think you need to.

src/squidpy/datasets/_registry.py
```diff
@@ -0,0 +1,219 @@
...
+            if f.name.startswith(prefix):
+                return f
+        return None
+
+    @property
+    def has_hires_image(self) -> bool:
+        """Return True if dataset has high-res image available."""
+        if not self.is_adata_with_image:
+            return False
+        return self.get_file_by_prefix("image.") is not None
```
flying-sheep: I don’t think you’re using this, you directly use `self.get_file_by_prefix("image.")`, right?

src/squidpy/datasets/_downloader.py
selmanozleyen: yeah. But I guess it always confused me because I wouldn't know if a visium 10x dataset meant a dataset from that company or a dataset that complies to that standard.

selmanozleyen: I will change it to visium_10x

src/squidpy/datasets/_downloader.py
selmanozleyen: thanks! super useful info

src/squidpy/datasets/_registry.py
selmanozleyen: great point, wasn't aware of that

src/squidpy/datasets/_registry.py
selmanozleyen: yes, I removed other redundant methods

src/squidpy/datasets/_downloader.py
selmanozleyen: I got rid of the `_get_first_file` thing. I fetch them by suffix now.

src/squidpy/datasets/_registry.py
selmanozleyen: done

src/squidpy/datasets/_registry.py
```diff
@@ -0,0 +1,198 @@
...
+from enum import Enum
+from functools import lru_cache
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+import yaml
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+    from importlib.resources import Traversable
```
flying-sheep: isn’t this the right import?

```suggestion
    from importlib.resources.abc import Traversable
```

src/squidpy/datasets/_registry.py
```diff
@@ -0,0 +1,198 @@
...
+    @classmethod
+    def from_yaml(cls, config_path: Path | str | None = None) -> DatasetRegistry:
+        """Load registry from YAML configuration file."""
+        # This case should be always true
+        # only for testing and tinkering config_path should be provided
+        if config_path is None:
+            with _get_config_traversable().open() as f:
+                config = yaml.safe_load(f)
+        else:
+            config_path = Path(config_path)
```
flying-sheep: You just use it in `open` in the next line, no need to convert it.

```suggestion
```

src/squidpy/datasets/_registry.py
```diff
@@ -0,0 +1,198 @@
...
+
+@dataclass
+class DatasetRegistry:
+    """Central registry for all squidpy datasets."""
+
+    s3_base_url: str = ""
+    datasets: dict[str, DatasetEntry] = field(default_factory=dict)
+
+    @classmethod
+    def from_yaml(cls, config_path: Path | str | None = None) -> DatasetRegistry:
```
flying-sheep: Maybe this? Or use the one from `squidpy.read._utils` instead of `os.PathLike | str` if that doesn’t lead to circular imports

```suggestion
    def from_yaml(cls, config_path: os.PathLike | str | None = None) -> DatasetRegistry:
```

src/squidpy/datasets/_registry.py
```diff
@@ -0,0 +1,198 @@
...
+import importlib.resources
+from dataclasses import dataclass, field
+from enum import Enum
+from functools import lru_cache
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+import yaml
+
+if TYPE_CHECKING:
```
flying-sheep, line 13: if you use `os.PathLike`, you need this:

```suggestion
if TYPE_CHECKING:
    import os
```

flying-sheep (approved):
Awesome! Some small nitpicks, otherwise this looks great.

I like how the `_make_loader` deduplication turned out: very elegant with the `entry.type`.

src/squidpy/datasets/_registry.py
selmanozleyen: done! also created an issue about type aliases locations https://github.com/scverse/squidpy/issues/1085. Because there is no standard currently

PR #1084, opened 2025-12-15
`make_tiles` now properly copies the transformations from source img

timtreis:


src/squidpy/experimental/im/_make_tiles.py
```diff
@@ -228,6 +228,12 @@ def _save_tiles_to_shapes(
     )
 
     sdata.shapes[shapes_key] = ShapesModel.parse(tile_gdf)
+    try:
+        transformations = get_transformation(sdata.images[image_key], get_all=True)
```
selmanozleyen: Isn't there a better way to check if key's exist than relying on exceptions? Get transformations might not be complicated but in a function that is complicated there is no way of knowing where exactly the `KeyError` came from. Maybe it is from a bug you need to fix for example.

selmanozleyen: https://github.com/scverse/squidpy/issues/1086 I made it into an issue since it seems like it could also be a spatialdata thing. I will look into it further but seems like we would need to access `spatialdata` private interface to check without making the function raise an exception

selmanozleyen:
I was trying to document why it passes when you run `uvx hatch tests`  but when you do `uvx hatch run hatch-test.py3.12-stable:run -- tests/experimental/test_make_tiles.py ` it fails. Here is the AI generated report

1. **Test files in `tests/plotting/`** set matplotlib params at **module import time**:
   ```python
   # tests/plotting/test_graph.py (line 19-20)
   sc.pl.set_rcParams_defaults()
   sc.set_figure_params(dpi=40, color_map="viridis")

   # tests/plotting/test_spatial_static.py (line 21)
   sc.set_figure_params(dpi=40, color_map="viridis")
   ```

2. **When running full test suite:**
   - `tests/plotting/` modules import first (alphabetically before `tests/experimental/`)
   - `sc.set_figure_params(dpi=40)` sets global matplotlib rcParams
   - `tests/experimental/test_make_tiles.py` runs with these settings
   - Figure size matches expected images ✅

3. **When running only `test_make_tiles.py`:**
   - `tests/plotting/` modules never import
   - `sc.set_figure_params()` is never called
   - `spatialdata_plot` uses its own defaults (larger figure)
   - Figure size doesn't match expected images ❌

### The `conftest.py` does call `sc.pl.set_rcParams_defaults()`:
```python
def pytest_sessionstart(session: pytest.Session) -> None:
    matplotlib.use("Agg")
    matplotlib.rcParams["figure.max_open_warning"] = 0
    np.random.seed(42)
    warnings.simplefilter("ignore", OldFormatWarning)
    sc.pl.set_rcParams_defaults()  # <- This is NOT enough!
```

But it does **NOT** call `sc.set_figure_params(dpi=40)` which is what sets the figure size.

## Proposed Fix

Add `sc.set_figure_params(dpi=40)` to `conftest.py`'s `pytest_sessionstart()`:

```python
def pytest_sessionstart(session: pytest.Session) -> None:
    matplotlib.use("Agg")
    matplotlib.rcParams["figure.max_open_warning"] = 0
    np.random.seed(42)
    warnings.simplefilter("ignore", OldFormatWarning)
    sc.pl.set_rcParams_defaults()
    sc.set_figure_params(dpi=40, color_map="viridis")  # <-- ADD THIS
```

src/squidpy/experimental/im/_make_tiles.py
timtreis: No you actually have a point here. That was unneccecarily defense here, we know that the image_key exists because otherwise it fails earlier and at the very least, that element would have an `Identity` transform

PR #1093, opened 2026-01-14
Experimental Feature: Function delegation to rapids_singlecell when gpu is available 

selmanozleyen:
hi @flying-sheep @ilan-gold @timtreis @Intron7 , for squidpy 2.0 we are planning to delegate to the gpu functionality when there is a device present by default. I am trying to come up with an API for this. I saw that the scanpy settings class was very complicated so I though maybe we need something simpler for our case.

This is what I came up with. I will come up with the `co_occurence` support after lunch but I wanted an initial feedback on this. 

Example usage
```python
from squidpy._utils import resolve_device_arg

def co_occurrence(
    adata: AnnData | SpatialData,
    device: Literal["cpu", "gpu"] | None = None,
) -> tuple[NDArrayA, NDArrayA] | None:

    effective_device = resolve_device_arg(device)

    if effective_device == "gpu":
        from rapids_singlecell.squidpy import co_occurrence as rsc_co_occurrence

        return rsc_co_occurrence(
            adata,
        )
     ...

```

Installation instructions
```
pip install -e '.[docs,test,dev,gpu-cuda12]' --extra-index-url https://pypi.nvidia.com
```


src/squidpy/gr/_ppatterns.py
```diff
@@ -381,6 +385,22 @@ def co_occurrence(
...
+            adata,
+            cluster_key=cluster_key,
+            spatial_key=spatial_key,
+            interval=interval,
+            copy=copy,
+            n_splits=n_splits,
+            n_jobs=n_jobs,
+            backend=backend,
+            show_progress_bar=show_progress_bar,
+        )
```
ilan-gold: I _think_ this can generally be wrapped up in a decorator where we:

1. check `effective_device`
2. Subset the inbound arguments from squidpy to those of rapids' function (I imagine `n_jobs` doesn't apply, this can probably be done with python's `inspect` module)
3. run the rapids function
4. Optionally bring the data back to the CPU (I would guess people want this, or at least it should be behind a flag)

src/squidpy/_settings/_settings.py
```diff
@@ -0,0 +1,41 @@
...
+    @device.setter
+    def device(self, value: DeviceType) -> None:
+        if value not in get_args(DeviceType):
+            raise ValueError(f"device must be one of {get_args(DeviceType)}, got {value!r}")
+        if value == "gpu" and not self.gpu_available():
+            raise RuntimeError("GPU unavailable. Install: pip install squidpy[gpu-cuda12]")
+        _device_var.set(value)
+
+    @staticmethod
+    def gpu_available() -> bool:
```
flying-sheep: could just be an `@property`, as this should be pretty fast

src/squidpy/_settings/_dispatch.py
```diff
@@ -0,0 +1,139 @@
...
+            gpu_only_values.update(
+                {k: variadic_kwargs.pop(k) for k in list(variadic_kwargs) if k in registry["gpu_only"]}
+            )
+
+            check_cpu_params(func_name, **gpu_only_values)
+
+            # Apply defaults for CPU-only params that are None
+            apply_defaults(func_name, all_args, "cpu")
+
+            return func(**all_args, **variadic_kwargs)
```
flying-sheep: How about this?

```suggestion
            device = _resolve_device(device)
            if device == "gpu":
                other_params = "cpu_only"
            else:
                other_params = "gpu_only"

            # Collect param values and check them (error if user provided)
            wrong_values = {k: all_args.pop(k) for k in list(all_args) if k in registry[other_params]}
            wrong_values.update(
                {k: variadic_kwargs.pop(k) for k in list(variadic_kwargs) if k in registry[other_params]}
            )

            check_params(func_name, device, **wrong_values)

            # Apply defaults for params that are None
            apply_defaults(func_name, all_args, device)

            if device == "gpu"
                module = importlib.import_module(gpu_module)
                func = getattr(module, _gpu_func_name)
            return func(**all_args, **variadic_kwargs)
```

src/squidpy/gr/_gpu.py
```diff
@@ -0,0 +1,159 @@
...
+    "n_jobs": GpuParamSpec(None),
+    "backend": GpuParamSpec("loky"),
+    "show_progress_bar": GpuParamSpec(True),
+}
+_SEED: dict[str, GpuParamSpec] = {"seed": GpuParamSpec(None)}
+
+# Registry: {func_name: {"cpu_only": {...}, "gpu_only": {...}}}
+# - cpu_only: parameters only supported on CPU (error on GPU if user provided a non-None value)
+# - gpu_only: parameters only supported on GPU (error on CPU if user provided a non-None value)
+GPU_PARAM_REGISTRY: dict[str, dict[str, dict[str, GpuParamSpec]]] = {
```
flying-sheep: are these reused for multiple functions? Otherwise you could move them into the decorator call to have them right above the function definition.

src/squidpy/gr/_gpu.py
```diff
@@ -0,0 +1,159 @@
...
+            "numba_parallel": GpuParamSpec(None),
+            **_SEED,
+            **_PARALLELIZE,
+        },
+        "gpu_only": {},
+    },
+}
+
+
+def check_gpu_params(func_name: str, **cpu_only_values: Any) -> None:
```
flying-sheep:
```suggestion
def check_params(func_name: str, device: Literal["cpu", "gpu"], /, **other_values: Any) -> None:
```

and then delete `check_cpu_params`

src/squidpy/gr/_ppatterns.py
```diff
@@ -104,9 +108,14 @@ def spatial_autocorr(
         Layer in :attr:`anndata.AnnData.layers` to use. If `None`, use :attr:`anndata.AnnData.X`.
     attr
         Which attribute of :class:`~anndata.AnnData` to access. See ``genes`` parameter for more information.
-    %(seed)s
+        Ignored when ``device='gpu'``.
```
flying-sheep: ignored? or raises error?

src/squidpy/_docs.py
```diff
@@ -119,6 +119,28 @@ def decorator2(obj: Any) -> Any:
...
+    For ``backend="loky"``, the number of cores used by numba for
+    each job spawned by the backend will be set to 1 in order to
+    overcome the oversubscription issue in case you run
+    numba in your function to parallelize.
+    To set the absolute maximum number of threads in numba
+    for your python program, set the environment variable:
+    ``NUMBA_NUM_THREADS`` before running the program.
+backend
+    Parallelization backend to use. See :class:`joblib.Parallel` for available options.
+    Ignored when ``device='gpu'``.
```
flying-sheep: ditto

flying-sheep (changes requested):
Nice! A bit of duplication left, but looks good!

src/squidpy/_settings/_settings.py
ilan-gold: Could also cache it? I guess people sometimes live-install things in jupyter envs

src/squidpy/gr/_gpu.py
```diff
@@ -0,0 +1,159 @@
...
+__all__ = ["GPU_PARAM_REGISTRY", "GpuParamSpec", "check_gpu_params", "check_cpu_params", "apply_defaults"]
+
+
+@dataclass
+class GpuParamSpec:
+    """Specification for a parameter's GPU compatibility."""
+
+    default: Any
+    message: str | None = None
+    validator: Callable[[Any], str | None] | None = None
```
ilan-gold: IMO verbs should be used to title functions.  `validator` makes me think this is a class

src/squidpy/gr/_gpu.py
```diff
@@ -0,0 +1,159 @@
+"""GPU parameter registry for squidpy.gr functions.
```
ilan-gold: Just a high level conceptual question about the implementation - why do we need an explicit registry of function mappings? Shouldn't this be doable on the fly?

```python
import inspect

def my_cpu_func(arg1: int, arg2: slice, cpu_only_arg: None | int = None):
    pass

def my_gpu_func(arg1: int, arg2: slice, gpu_only_arg: None | str = None):
    pass

# Check for different params:
cpu_only_params = set(inspect.signature(my_cpu_func).parameters.keys()).difference(inspect.signature(my_gpu_func).parameters.keys())

# get default param
is_default_cpu_only_param = inspect.signature(my_cpu_func).parameters[next(iter(cpu_only_params))].default

# then error out if the cpu only param is not default but the device is gpu
...
```

I guess the thing this would miss is custom validators but there appears to be only one of those and I imagine it is in @Intron7 's interest to have a matching API anyway, so you could special-case just that.

Am I missing something? I'd think that special-casing from this is probably more sustainable than a self-maintained registry? Like you have `"n_jobs": GpuParamSpec(None)` but I'm pretty sure that could be resolved by the above code.

src/squidpy/gr/_gpu.py
selmanozleyen: will call it validate_fn since it's a callable attr

src/squidpy/gr/_gpu.py
selmanozleyen: That's a good point. The reason I did it as a registry was my want to explicitly have a list of cpu/gpu only params and special cases. Sure it reduces code but having a registry to read is way easier to understand and guess the behaviour.

For example when a developer registers a rsc function they won't see any registry error and get away without thinking about the non-overlapping arguments and validation logic. Which makes debugging harder. This was my thinking.

But you are right that we can just have like a registry of special cases.

src/squidpy/_settings/_settings.py
selmanozleyen: Would a functools cache invalidate when there is live install? If so that would be the perfect solution. I agree that we can just document not supporting live install

selmanozleyen: I checked and cache can't support autoreload. Autoreload would redefine a function only when file is changed and the cache wouldn't get invalidated because only the function object itself would be the cache check.

selmanozleyen:
I checked and it works in our clusters.
I have to note, `spatial_autocorr` is only equal between gpu and cpu when `rtol=1e-3`

selmanozleyen:
closed in favor of https://github.com/scverse/squidpy/pull/1151

PR #1097, opened 2026-01-21
fix: Support SpaceRanger v3 format

### LINKED ISSUE #1089: Bug for reading the tissue_positions.csv
### Report

https://github.com/scverse/squidpy/blob/538427b470467211a106c28e3980e3eda8d60eb4/src/squidpy/read/_read.py#L83

Should we use header=0 instead? header=1 make the data loose information of the first spot after change the column name

Thank you!

### Versions

```python

```

selmanozleyen:
fixes: https://github.com/scverse/squidpy/issues/1089

The format has a change in naming 

- v1: tissue_positions.csv (no header)
- v2: tissue_positions_list.csv (with header)
- v3: tissue_positions.csv (with header) (new)

Another change is header is set to 0 instead of 1. That was most likely a bug from https://github.com/scverse/squidpy/commit/03ee8faa12d1fc7fe2b8edb4386941a1c96f49ce but it didn't show up since the spaceranger v1 datasets are at scanpy and scanpy also uses `header=0` https://github.com/scverse/scanpy/blob/f75575cac1e7f8779b33cb5570bc9c726313fd1d/src/scanpy/readwrite.py#L504

### DIFF
--- a/src/squidpy/read/_read.py
+++ b/src/squidpy/read/_read.py
@@ -85,7 +85,7 @@ def visium(
     # Detect header by checking if first cell is 'barcode' (header) or a barcode value
     with open(tissue_positions_file) as f:
         first_cell = f.readline().split(",")[0].strip()
-    has_header = first_cell == "barcode"
+    has_header = first_cell.lower() == "barcode"
 
     coords = pd.read_csv(
         tissue_positions_file,
--- a/src/squidpy/read/_read.py
+++ b/src/squidpy/read/_read.py
@@ -72,15 +72,24 @@ def visium(
         (path / f"{Key.uns.spatial}/scalefactors_json.json").read_bytes()
     )
 
+    # Space Ranger versions use different file formats:
+    #   - v1: tissue_positions.csv (no header)
+    #   - v2: tissue_positions_list.csv (with header)
+    #   - v3: tissue_positions.csv (with header)
     tissue_positions_file = (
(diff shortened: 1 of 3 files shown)


src/squidpy/read/_read.py
```diff
@@ -72,15 +72,24 @@ def visium(
...
+    #   - v3: tissue_positions.csv (with header)
     tissue_positions_file = (
         path / "spatial/tissue_positions.csv"
         if (path / "spatial/tissue_positions.csv").exists()
         else path / "spatial/tissue_positions_list.csv"
     )
 
+    # Detect header by checking if first cell is 'barcode' (header) or a barcode value
+    with open(tissue_positions_file) as f:
+        first_cell = f.readline().split(",")[0].strip()
```
timtreis, line 87: I'd probably go paranoia here and 

a) not split by ,
b) check if "arcode" is in the resulting string (intentionally without B to be capitalisation resistant)

timtreis (changes requested):
tiny nit

src/squidpy/read/_read.py
selmanozleyen: Yeah I guess you can never know with visium. I will do the check with .lower() then for case insenitivity. arcode migh confuse programers

timtreis (approved):
Lgtm

PR #1105, opened 2026-01-29
ci: mark test_parallelize as flaky

### LINKED ISSUE #1104: Test parallelize timeouts often
### Report

these tests fail randomly in Squidpy with a timeout - like they work if I just restart them often enough, but that's a bit annoying. Is this something we can solve on the code side? Or do we just have to increase the timeout? (can we even do that?)

Image

selmanozleyen:
fixes: https://github.com/scverse/squidpy/issues/1104

### DIFF
--- a/tests/utils/test_parallelize.py
+++ b/tests/utils/test_parallelize.py
@@ -67,7 +67,7 @@ def func(request) -> Callable:
 # in case of failure.
 
 
-@pytest.mark.timeout(30)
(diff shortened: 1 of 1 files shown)


selmanozleyen:
@flying-sheep I created an issue here https://github.com/scverse/squidpy/issues/1111 could we just merge it for now?

flying-sheep (approved):

PR #1117, opened 2026-02-18
fix: require spatialdata>=0.7.2 to drop the xarray-schema/pkg_resources break

### LINKED ISSUE #1115: pkg_resources not founded (deprecated since python 3.12)
### Report

ModuleNotFoundError: No module named 'pkg_resources' while importing. Full trace goes to xarray_schema __init__
Module is currently deprecated in latest python versions: https://github.com/mu-editor/mu/issues/2485
Squidpy version: 1.7.0
Python version: 3.12

### Versions

```python

```

timtreis:
Closes #1115.

`import squidpy` crashed on Python 3.12 / setuptools>=82 with `ModuleNotFoundError: No module named 'pkg_resources'`, traced into `xarray_schema` — pulled in transitively by spatialdata, which imported the removed `pkg_resources`.

spatialdata **0.7.2** dropped the `xarray-schema` dependency, so the proper fix is a lower-bound bump. The previous `>=0.7.1` floor (#1114) was set to drop `dask-expr` and still permits spatialdata 0.7.1, which keeps the bug.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,7 +66,7 @@ dependencies = [
   "scikit-image>=0.25",
   # due to https://github.com/scikit-image/scikit-image/issues/6850 breaks rescale ufunc
   "scikit-learn>=0.24",
-  "spatialdata>=0.7.1",
+  "spatialdata>=0.7.2",  # 0.7.2 dropped xarray-schema (pkg_resources break, #1115)
   "spatialdata-plot>=0.3.3",
   "statsmodels>=0.12",
   # https://github.com/scverse/squidpy/issues/526


Zethson:
I know you're labeling it as a bandaid fix but xarray_schema is dead and also got removed from SpatialData. The correct solution is to get rid of xarray_schema.

alam-shahul:
Will this be merged?

timtreis:
Probably not, https://github.com/scverse/spatialdata/commit/e88c24e800fb0be554420dedbafb1f5f302178d0 was released with spatialdata [v0.7.2](https://github.com/scverse/spatialdata/releases/tag/v0.7.2), so we'll lower-pin to that version with the next release. When the PR for that pin is merged, I'll close this issue

timtreis:
Modified this PR to just raise the lower pin of `SpatialData` which then automatically adresses the issue

selmanozleyen (approved):

PR #1120, opened 2026-02-20
feat: remove unnecessary parallelize usage in sepal

selmanozleyen:
hi, 

from the rsc sepal implementation I know we can get rid of `parallelize` easily. I also have a longer term plan to remove parallelize. I checked with this script: https://gist.github.com/selmanozleyen/c4b0e4780243fd4621f68bc2d78cae5c

Results pretified with AI:

### Benchmark results (502 genes, 2688 cells, visium_hne_adata, 3 runs, 8 cores)

| variant                                |   mean |  median |    min |    max |
|----------------------------------------|--------|---------|--------|--------|
| main, n_jobs=1 (sequential)            | 134.9s | 134.4s  | 134.3s | 136.0s |
| main, n_jobs=6 (joblib/loky)           |  40.5s |  40.2s  |  37.9s |  43.5s |
| this PR, prange (no pre-alloc)         |  37.7s |  37.6s  |  36.9s |  38.6s |
| this PR, prange + pre-alloc buffers    |  31.9s |  31.3s  |  30.9s |  33.6s |

## Motivation

This PR replaces the internal `parallelize` + `joblib.Parallel` machinery in
`sepal()` with Numba's native `prange` threading, removing the need for
multi-process orchestration (subprocess spawning, pickling, IPC).

### Key changes

1. **`prange` over genes** -- The outer gene loop is parallelized with
   `numba.prange` inside a `@njit(parallel=True)` function, replacing
   `joblib.Parallel(n_jobs=..., backend="loky")`.

2. **Pre-allocated per-thread workspace buffers** (this commit https://github.com/scverse/squidpy/pull/1120/changes/e011ddcbb6abd29f89b5beebb0ea3646946c6565) You can see why I did this in https://numba.readthedocs.io/en/stable/user/parallel.html#diagnostics under "allocation hoisting"


src/squidpy/gr/_sepal.py
```diff
@@ -160,69 +144,82 @@ def sepal(
...
-    if queue is not None:
-        queue.put(Signal.FINISH)
-
-    return np.array(score)
+    n_genes = vals_dense.shape[1]
+    n_cells = vals_dense.shape[0]
+    sat_shape = sat.shape[0]
+    n_threads = get_num_threads()
+
+    # Pre-allocate per-thread workspace to avoid allocator contention
```
flying-sheep: I wonder if there isn’t a way to tell numba to do the pre-allocation instead of doing it manually.

flying-sheep (commented):
Nice, very straightforward.

The only potential issue I see is that you densify `vals` as one, whereas the other version densified at most `n_jobs` columns of `vals` at a time.

Are there use cases for this where that matters?

src/squidpy/gr/_sepal.py
selmanozleyen: I am also wondering the same but I couldn't find a way to do it without calling `get_num_threads()`

selmanozleyen: @Intron7 do you know anything about this?

selmanozleyen:
I found out numba nogil and python threads work well enough and utilizes the cores

On sparse data (not densifying upfront)
```
Running sepal() 3 time(s)...
  run 1: 37.588s
  run 2: 39.458s
  run 3: 39.363s
```
Also we get to have our progress bar.

selmanozleyen:
hi @timtreis can you approve this so I can use some of the utils in #1023 ?

src/squidpy/gr/_sepal.py
```diff
@@ -163,69 +161,80 @@ def sepal(
...
-    if queue is not None:
-        queue.put(Signal.FINISH)
+    sparse = issparse(vals)
 
-    return np.array(score)
+    def _process_gene(i: int) -> float:
+        if sparse:
+            conc = np.ascontiguousarray(vals[:, i].toarray().ravel(), dtype=np.float64)
+        else:
+            conc = vals[:, i]
```
timtreis: You're no longer doing a .copy so you modify in place with potentially unintended consequences. I think it's also not tested because the testing data is sparse?

src/squidpy/gr/_sepal.py
```diff
@@ -163,69 +161,80 @@ def sepal(
...
+    """Run diffusion for each gene column, parallelised across threads."""
 
-    if queue is not None:
-        queue.put(Signal.FINISH)
+    sparse = issparse(vals)
 
-    return np.array(score)
+    def _process_gene(i: int) -> float:
+        if sparse:
+            conc = np.ascontiguousarray(vals[:, i].toarray().ravel(), dtype=np.float64)
```
timtreis, line 183: We're treating dense and sparse here differently, one is converted to float, the other one is taken as is - we should standardize

tests/graph/test_sepal.py
```diff
@@ -16,8 +17,14 @@ def test_sepal_seq_par(adata: AnnData):
     adata.var["highly_variable"] = rng.choice([True, False], size=adata.var_names.shape, p=[0.005, 0.995])
 
     sepal(adata, max_neighs=6)
-    df = sepal(adata, max_neighs=6, copy=True, n_jobs=1)
-    df_parallel = sepal(adata, max_neighs=6, copy=True, n_jobs=2)
+
+    prev_threads = numba.get_num_threads()
```
timtreis: This feels flaky? Can we not just run it once with 1 core, once with 2 cores and then compare? We might also need a fixture to make sure that the GH runner actually has 2 cores

src/squidpy/_utils.py
```diff
@@ -242,6 +243,46 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
 
 
+def thread_map(
+    fn: Callable[..., Any],
+    items: Iterable[Any],
+    *,
+    n_jobs: int = 1,
+    show_progress_bar: bool = False,
+    unit: str = "item",
+    total: int | None = None,
```
timtreis, line 253: We pass `total` but never use

src/squidpy/gr/_sepal.py
```diff
@@ -82,8 +82,8 @@ def sepal(
     use_raw
         Whether to access :attr:`anndata.AnnData.raw`.
     %(copy)s
-    %(parallelize)s
-
+    %(n_jobs)s
+    %(show_progress_bar)s
```
timtreis, line 86: Does this still work without the %(parallelize)s?

timtreis (changes requested):

src/squidpy/gr/_sepal.py
selmanozleyen: yes see the docs

src/squidpy/_utils.py
selmanozleyen: it will be used in the ligrec one

tests/graph/test_sepal.py
selmanozleyen: you are right, i fixed it

src/squidpy/gr/_sepal.py
selmanozleyen: ok done

src/squidpy/gr/_sepal.py
selmanozleyen: should be solved now

timtreis (approved):
lgtm

PR #1121, opened 2026-02-23
fix: remove unused parameters in co_occurence

### LINKED ISSUE #1110: Unused parallelization parameters in `co_occurrence`
### Report

**File:** `src/squidpy/gr/_ppatterns.py`

**Description:** The `co_occurrence` function accepts `n_jobs`, `backend`, and `show_progress_bar` parameters that are documented as CPU-only but are currently unused on the CPU path. The function now uses a pure numba implementation (`_co_occurrence_helper`) that doesn't leverage joblib parallelization.

selmanozleyen:
fixes: https://github.com/scverse/squidpy/issues/1110 and https://github.com/scverse/squidpy/issues/1048 .

n_splits usage was removed here https://github.com/scverse/squidpy/commit/32789aefd53b797ceee8c3153beded2f46a2cd4c


flying-sheep (commented):
this will break code that uses these parameters – sure they are unused, but you could warn first that they do nothing and will be removed, and then remove them later,

timtreis (approved):
lgtm

durr1602:
just a quick question, was there anything else to replace the progress bar? i'm not even sure there was one with the squidpy version I'm testing, but with this new version there doesn't seem to be any sort of log (at least on my end).

selmanozleyen:
It should log something at least looking at "start = logg.info(f"Calculating co-occurrence probabilities for `{len(interval)}` intervals")". Maybe you should check stderr?

durr1602:
Yes, that's what I'd thought as well.. weird. I'm running a Jupyter notebook in Positron. I've tried both the new Positron notebook editor (which is still under active dev I believe) and the legacy editor, but no output with either. I tried to capture both stdout and stderr using the magic `%%capture` but both are empty..

I thought I'd just ask, if the issue is not reproducible on your end, it's not that big of a deal. Let me know if you think it warrants opening an issue.

PR #1124, opened 2026-02-23
chore: extend the timeout for loky progress init

selmanozleyen:
I added these tests to ensure robustness but I think they are too harsh for `parallelize` at this point and they just end up being annoying. However I think the fact that a simple test file can break so easily underlines why nobody should use parallelize.

related: https://github.com/scverse/squidpy/issues/1123


flying-sheep:
ah nice! once you get rid of parallel, you can still get rid of this

flying-sheep (approved):

PR #1125, opened 2026-02-23
feat: remove ligrec parallelize

selmanozleyen:
[benchmark code](https://gist.github.com/selmanozleyen/97e3abec793555ab1fa9ec1e88575aaa)

[main](https://gist.github.com/selmanozleyen/33c63461d0ad15f59ad123fabc06e283) results
[pr](https://gist.github.com/selmanozleyen/f48d8f8f86c7b8d7bdb9ae225e3eceb8) results

Results compared:
```
python benchmarks/bench_ligrec.py --compare benchmarks/results


scenario         n_jobs   main (s)     PR (s)  speedup   change
---------------------------------------------------------------
large                 1      2.548      3.865    0.66x  -51.7%
large                 4      1.561      1.259    1.24x +  19.3%
large                 8      1.706      1.122    1.52x +  34.2%
many_perms            1      1.620      2.247    0.72x  -38.7%
many_perms            4      0.924      0.637    1.45x +  31.1%
many_perms            8      0.966      0.647    1.49x +  33.1%
medium                1      0.261      0.425    0.61x  -62.8%
medium                4      0.581      0.183    3.18x +  68.5%
medium                8      0.687      0.173    3.97x +  74.8%
xlarge                1      8.139      9.259    0.88x  -13.8%
xlarge                4     10.498      3.077    3.41x +  70.7%
xlarge                8     10.188      2.768    3.68x +  72.8%
```

both faster and cleaner code. this removes parallelize.

update: the reason main is faster when n_jobs=1 is because main sets also  `numba_parallel=True` so it's because it's still numba parallel even though it's one process.


src/squidpy/gr/_ligrec.py
```diff
@@ -658,24 +570,74 @@ def ligrec(
...
                 copy=copy,
                 key_added=key_added,
-                **kwargs,
+                n_jobs=n_jobs,
+                show_progress_bar=show_progress_bar,
             )
         )
 
 
+@njit(nogil=True, cache=True)
```
ilan-gold: Why not parallel=True + prange? Because this is being run in a thread pool? Why not just make every individual step parallel?
https://numba.pydata.org/numba-doc/dev/user/parallel.html?highlight=njit#explicit-parallel-loops

Would this require rewriting into a reduction of some sort to prevent overlapping writes?

I see in the benchmarks that the speedups with more jobs is not really scaling linearly, which is not what I would expect.

src/squidpy/gr/_ligrec.py
```diff
@@ -719,145 +676,67 @@ def _analysis(
...
+                perm,
+                inv_counts,
+                mean_obs,
+                interactions,
+                interaction_clusters,
+                valid,
+                local_counts,
+            )
+            pbar.update(1)
+        return local_counts
```
ilan-gold: Why can't this also be numba-ified with an outer-loop of some sort? Why do we still need a thread pool? I thought "one giant kernel" was the goal

Is shuffling not parallelizable? Certainly there are ways around this like `argsort` + randomindices or somethign? Other than that, I don't really see why the`range(chunk_sizes[t])` couldn't be parallelized.  Is it the validity of `local_counts`? Seems like there should be ways around this

selmanozleyen: to have a responsive progress bar and to have the same shuffling results as old version.

src/squidpy/gr/_ligrec.py
selmanozleyen: > I see in the benchmarks that the speedups with more jobs is not really scaling linearly, which is not what I would expect.

Thats a good point worth investigating

src/squidpy/gr/_ligrec.py
ilan-gold: Could you explain a bit more

1. Why is the "same results" thing a hard blocker? `clustering` seems small so copy+shuffle should be cheap as a pre-processing step i.e., do all the "shuffle" stuff ahead of time / outside `numba`
2. Would you expect a giant kernel to be faster? My gut is "yes" given Severin's experience/our experience with `co_occurrence` but I'm all ears

selmanozleyen: > clustering seems small so copy+shuffle should be cheap as a pre-processing step i.e.,
> perm = clustering.copy()

Copies per thread. To allocate ahead of time you'd need `n_perms * len(clustering)`. So it can blow up if the user give high `n_perms`. You can also do per thread in place shuffling but that changes the behaviour and results. 


> Would you expect a giant kernel to be faster? My gut is "yes" given Severin's experience/our experience with co_occurrence but I'm all ears

For sure.

src/squidpy/gr/_ligrec.py
selmanozleyen: > Why not parallel=True + prange? Because this is being run in a thread pool? Why not just make every individual step parallel?
> https://numba.pydata.org/numba-doc/dev/user/parallel.html?highlight=njit#explicit-parallel-loops
> 
> Would this require rewriting into a reduction of some sort to prevent overlapping writes?

For keeping the progress bar. Also fits current code more naturally, I agree that it's better and faster but in the sepal PR but this pattern helped me keep the progress bar and avoid duplicate code in a straightforward way. For example I had to create a function that would densify per thread but pushing this step to python reduced the duplicate code.

src/squidpy/gr/_ligrec.py
ilan-gold: @timtreis How many permutations are realistic? I would the memory hit can't be that high here

Separate from our "internal knowledge", I have to wonder if this is wroth trying out and seeing where it breaks/degrades.  If there's a way to do this that doesn't involve changing results but you get more speed at the risk of higher memory usage, I think we should understand that tradeoff

this might be a nice way to retain the progress bar without depending on the linked package:

Instead of explicit threads each working on a "chunk", we would redo the algorithm so that we sequentially proceed over the "chunks" but process each "chunk" at once in using parallelized `numba` instead of `for _ in range(chunk_sizes[t])`. This could help alleviate the memory concerns by doing allocations for only `clustering * chunk_sizes[t]` amount of data instead of the full `n_perms * len(clustering)`.

Do I ahve this right?

src/squidpy/gr/_ligrec.py
ilan-gold: Two options at a high level:

1. Refactor for the `numba` itself to be parallel and work on "batches"
2. Does https://github.com/mortacious/numba-progress work?

src/squidpy/gr/_ligrec.py
```diff
@@ -719,145 +676,67 @@ def _analysis(
...
+
+    base_chunk, remainder = divmod(n_perms, n_jobs)
+    chunk_sizes = np.full(n_jobs, base_chunk, dtype=np.int64)
+    chunk_sizes[:remainder] += 1
+
+    pbar = tqdm(total=n_perms, unit="permutation", disable=not show_progress_bar)
+
+    def _worker(t: int) -> NDArrayA:
+        local_counts = np.zeros((n_inter, n_cpairs), dtype=np.int64)
+        rs = np.random.RandomState(None if seed is None else t + seed)
```
ilan-gold: Why do we need to use the old `RandomState`? Can we get the same result by using `Generator`?

selmanozleyen: > Can we get the same result by using Generator?

I am not sure but isn't this what Phil tried to do in his PR? By using the legacy RandomState internally by default when no rng was given. I think he also said the guarantees for reproducibility is different than RandomState (I verified it and its correct). Like Generator doesn't guarantee reproducibility across different versions of numpy.

src/squidpy/gr/_ligrec.py
selmanozleyen: I will have a look at these questions by doing some runs. Didn't fully understand some parts but will come back to you. But to be clear: would not reproducing old results be a dealbreaker for you? I didn't understand your stance on this.

ilan-gold: > But to be clear: would not reproducing old results be a dealbreaker for you? I didn't understand your stance on this.

That's not really my call here.  That being said, what I was trying to propose was a way of using `numba`'s `parallel=True` mechanism _without_ breaking backwards compatible reproducibility.  It seems like all the reproducibility concerns boil down to the shuffling, so I'm proposing basically front loading the shuffling so that the kernel can run in parallel on multiple pre-shuffled permutations

src/squidpy/gr/_ligrec.py
flying-sheep: My use of the legacy RandomState is carefully isolated so I never directly call its APIs. The idea is that support for it can just be removed one day without changing code in any way other than removing some `if legacy` branches. I did this by making sure that when a user passes a `Generator`, no RandomState object is created anywhere in scanpy’s code.

src/squidpy/gr/_ligrec.py
flying-sheep: I think it’s worth doing a quick and dirty evaluation if “one giant kernel” is feasible. In my attempt to do it for `downsample_counts`, I was blocked by numba lacking an important API.

ilan-gold: Could you highlihgt what was mising? The RNG API you added? 

https://github.com/scverse/scanpy/pull/4004

I don't see anything specific in the above but maybe I'm missing it.

src/squidpy/gr/_ligrec.py
ilan-gold: Right, so unblocking this PR _could_ involve making a separate PR that does here what Phil did for scanpy i.e., RNG compat, but I'm not sure I understand enough clearly what the blocker is around reproducibility and random state and furthemore, we haven't seen a "monolith" implementation.

And if there is no way to resolve the weird performance scaling, we should understand exactly why!

src/squidpy/gr/_ligrec.py
timtreis: > @timtreis How many permutations are realistic? I would the memory hit can't be that high here

no idea tbh, I'd probably also just try a few numbers and see when it breaks/converges

> numba's parallel=True mechanism

I'll complain about this every single time I see it 🥸 If we don't give the user the option to specify how many cores something uses, we should not parallelise - especially not if it's so opaque to a user like `numba` is. We cannot assume that we control all ressources on a given system

ilan-gold: https://numba.pydata.org/numba-doc/dev/user/threading-layer.html what goes wrong here?

flying-sheep: I kinda hate that all these parallelization libraries (dask, numba) have some underdocumented implicit global pool as well. Like, if you call `numba.set_num_threads()`, does it actually do that or just create a threadpool with 128 threads and mask some out? I’m sure the answer is “it depends”.

I wish there was a way for a threading framework to claim a thread and then to say to anything wanting to spawn a sub-thread “you now live in a world where there’s only 4 cores”, and that way there would be no accidental overcommitting.

I understand your radical stance in that light @timtreis. Forcing users to understand that dask, numba, and who knows what else all have to be babysat is a really sad consequence of our choice to adopt numba.

> Could you highlight what was missing?

good idea, I wrote it up here: https://github.com/scverse/scanpy/pull/4004#issuecomment-4155880691

timtreis: Haven't looked much under the hood because I haven't paralleised in numba yet outside of analysis scripts, but according to the link that @ilan-gold posted, [numba.set_num_threads()](https://numba.pydata.org/numba-doc/dev/user/threading-layer.html#numba.set_num_threads) should probably be the equvivalent of 

> you now live in a world where there’s only 4 cores

(which perfectly summarises what I want it do be 😢 )

So if that's actually true (or rather "we cannot prove the opposite") I'm fine with using it given that the user gets a `n_cores: int = -1` which is strictly adhered to

timtreis: >  Forcing users to understand that dask, numba, and who knows what else all have to be babysat is a really sad consequence of our choice to adopt numba.

I think it's our job as devs to fully hide this from the user. Most users only care about things being fast and things being useable in a pipeline, so limiting the amount of cores something uses is imho mandatory. But if your usecase goes beyond that, like if you need control over the exact backend or threads vs processes, it's maybe easier to just reimplement it from scratch given your constraints

src/squidpy/gr/_ligrec.py
selmanozleyen: > so unblocking this PR could involve making a separate PR that does here what Phil did for scanpy i.e., RNG compat

I dont understand what you mean by unblocking. If we want to add Generator support we might need something like what @flying-sheep did, yes. So it is out of scope.

I am going to verify if this scaling is because the benchmark is limited or not. Come up with new results and post it here.

selmanozleyen:
overview_time_vs_n_jobs



@ilan-gold here is the scaling behaviour benched more properly. My earlier results werent simply big enough to show scaling.


These are from the the PR. In main because the number of clusters are taken dynamically it can even take 2 minutes to compile which is crazy to me. So I didn't even bother comparing it with main.

I will open an issue to consider using monolith kernels but that requires a complete reimplementation that needs more effort than this. Here I abstracted away from some details and reused the old working parts. This PR is good enough for a speedup but also gets rid of multiprocessing and dynamic kernel creation in a sloppy way. So I think it is worth merging. We know that ligrec isn't a very hot function of squidpy anyway.

bench code: https://github.com/selmanozleyen/bench_squidpy

selmanozleyen:
Here are the latest results with one big numba kernel. One nice thing is it's also bit-by-bit same as main in terms of reproducibility.

```
scenario         n_jobs   main (s)     PR (s)  speedup   change
---------------------------------------------------------------
medium                1      0.261      0.321    0.81x  -23.0%
medium                4      0.581      0.124    4.67x  +78.6%
medium                8      0.687      0.107    6.41x  +84.4%
large                 1      2.548      3.218    0.79x  -26.3%
large                 4      1.561      1.310    1.19x  +16.0%
large                 8      1.706      0.936    1.82x  +45.1%
xlarge                1      8.139      7.978    1.02x   +2.0%
xlarge                4     10.498      2.950    3.56x  +71.9%
xlarge                8     10.188      2.300    4.43x  +77.4%
many_perms            1      1.620      1.873    0.86x  -15.6%
many_perms            4      0.924      0.639    1.45x  +30.8%
many_perms            8      0.966      0.550    1.76x  +43.1%
```

flying-sheep:
> One nice thing is it's also bit-by-bit same as main in terms of reproducibility.

that is really nice! all that work trying to use generators right is paying off

timtreis:
> Here are the latest results with one big numba kernel. One nice thing is it's also bit-by-bit same as main in terms of reproducibility.
> 
> ```
> scenario         n_jobs   main (s)     PR (s)  speedup   change
> ---------------------------------------------------------------
> medium                1      0.261      0.321    0.81x  -23.0%
> medium                4      0.581      0.124    4.67x  +78.6%
> medium                8      0.687      0.107    6.41x  +84.4%
> large                 1      2.548      3.218    0.79x  -26.3%
> large                 4      1.561      1.310    1.19x  +16.0%
> large                 8      1.706      0.936    1.82x  +45.1%
> xlarge                1      8.139      7.978    1.02x   +2.0%
> xlarge                4     10.498      2.950    3.56x  +71.9%
> xlarge                8     10.188      2.300    4.43x  +77.4%
> many_perms            1      1.620      1.873    0.86x  -15.6%
> many_perms            4      0.924      0.639    1.45x  +30.8%
> many_perms            8      0.966      0.550    1.76x  +43.1%
> ```

Quick question on this: The decrease in n_job=1 performance is because on `main`, it was parallelizing across n>1 anyway?

tests/conftest.py
```diff
@@ -257,10 +257,17 @@ def complexes(adata: AnnData) -> Sequence[tuple[str, str]]:
 
 
 @pytest.fixture(scope="session")
-def ligrec_no_numba() -> Mapping[str, pd.DataFrame]:
-    with open("tests/_data/ligrec_no_numba.pickle", "rb") as fin:
-        data = pickle.load(fin)
-        return {"means": data[0], "pvalues": data[1], "metadata": data[2]}
+def ligrec_pvalues_reference() -> Mapping[str, pd.DataFrame]:
```
timtreis, line 260: maybe describe structure/contents of file in comment here

selmanozleyen: But it is documented below no? 

> 
> axes, stored as an AnnData (X=pvalues, layers["means"]=means) with the index
>     # levels kept as obs/var columns.

timtreis (approved):
lgtm

PR #1130, opened 2026-02-27
Add explicit error handling in _make_tiles.py

### LINKED ISSUE #1086: Handle cases explicitly for `KeyError` or `ValueError` in `_make_tiles`
### Report

Handle cases explicitly for `KeyError` or `ValueError` in `_make_tiles`. For example in https://github.com/scverse/squidpy/blob/d4f256f03d7e6359e957084989e6f2104a1dc5b3/src/squidpy/experimental/im/_make_tiles.py#L840-L851

Seperating this issue to speed up development

timtreis:
## Summary
- Adds upfront parameter validation with clear `KeyError`/`ValueError` messages to `make_tiles()`, `make_tiles_from_spots()`, `_get_largest_scale_dimensions()`, and `_get_spot_coordinates()`
- Validates that image/spots keys exist in `sdata`, `tile_size` has positive values, and `min_tissue_fraction` is in [0, 1]
- Adds 5 test cases covering all new error conditions

Closes #1086

## Test plan
- [x] `test_make_tiles_invalid_image_key` — bad image key raises `KeyError`
- [x] `test_make_tiles_invalid_tile_size` — non-positive tile size raises `ValueError`
- [x] `test_make_tiles_invalid_min_tissue_fraction` — out-of-range fraction raises `ValueError`
- [x] `test_make_tiles_from_spots_invalid_spots_key` — bad spots key raises `KeyError`
- [x] `test_make_tiles_from_spots_invalid_image_key` — bad image key raises `KeyError`

### DIFF
--- a/src/squidpy/experimental/im/_make_tiles.py
+++ b/src/squidpy/experimental/im/_make_tiles.py
@@ -359,7 +359,6 @@ def make_tiles(
     make_tiles_from_spots
         Create tiles centered on Visium spots instead of a regular grid.
     """
-    # --- upfront parameter validation ---
     if image_key not in sdata.images:
         raise KeyError(f"Image key '{image_key}' not found in sdata.images. Available: {list(sdata.images.keys())}")
     if tile_size[0] <= 0 or tile_size[1] <= 0:
--- a/src/squidpy/experimental/im/_make_tiles.py
+++ b/src/squidpy/experimental/im/_make_tiles.py
@@ -171,9 +171,7 @@ def _get_largest_scale_dimensions(
 ) -> tuple[int, int]:
     """Get the dimensions (H, W) of the largest/finest scale of an image."""
     if image_key not in sdata.images:
-        raise KeyError(
-            f"Image key '{image_key}' not found in sdata.images. Available: {list(sdata.images.keys())}"
-        )
+        raise KeyError(f"Image key '{image_key}' not found in sdata.images. Available: {list(sdata.images.keys())}")
     img_node = sdata.images[image_key]
 
     # Use _get_element_data with "scale0" which is always the largest scale
@@ -363,9 +361,7 @@ def make_tiles(
     """
     # --- upfront parameter validation ---
     if image_key not in sdata.images:
-        raise KeyError(
-            f"Image key '{image_key}' not found in sdata.images. Available: {list(sdata.images.keys())}"
-        )
+        raise KeyError(f"Image key '{image_key}' not found in sdata.images. Available: {list(sdata.images.keys())}")
     if tile_size[0] <= 0 or tile_size[1] <= 0:
         raise ValueError(f"tile_size must have positive values, got {tile_size}.")
     if not 0 <= min_tissue_fraction <= 1:
@@ -540,9 +536,7 @@ def make_tiles_from_spots(
     if spots_key not in sdata.shapes:
         raise KeyError(f"Spots key '{spots_key}' not found in sdata.shapes")
     if image_key is not None and image_key not in sdata.images:
-        raise KeyError(
-            f"Image key '{image_key}' not found in sdata.images. Available: {list(sdata.images.keys())}"
-        )
+        raise KeyError(f"Image key '{image_key}' not found in sdata.images. Available: {list(sdata.images.keys())}")
     if not 0 <= min_tissue_fraction <= 1:
         raise ValueError(f"min_tissue_fraction must be in [0, 1], got {min_tissue_fraction}.")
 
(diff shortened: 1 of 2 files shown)


src/squidpy/experimental/im/_make_tiles.py
```diff
@@ -170,6 +170,8 @@ def _get_largest_scale_dimensions(
     image_key: str,
 ) -> tuple[int, int]:
     """Get the dimensions (H, W) of the largest/finest scale of an image."""
+    if image_key not in sdata.images:
+        raise KeyError(f"Image key '{image_key}' not found in sdata.images. Available: {list(sdata.images.keys())}")
```
selmanozleyen, line 174: these can be one liners with a common check function. also easier to read

src/squidpy/experimental/im/_make_tiles.py
```diff
@@ -357,6 +359,13 @@ def make_tiles(
     make_tiles_from_spots
         Create tiles centered on Visium spots instead of a regular grid.
     """
+    if image_key not in sdata.images:
+        raise KeyError(f"Image key '{image_key}' not found in sdata.images. Available: {list(sdata.images.keys())}")
+    if tile_size[0] <= 0 or tile_size[1] <= 0:
+        raise ValueError(f"tile_size must have positive values, got {tile_size}.")
+    if not 0 <= min_tissue_fraction <= 1:
+        raise ValueError(f"min_tissue_fraction must be in [0, 1], got {min_tissue_fraction}.")
```
selmanozleyen, line 367: this one can also be deduplicated as well

src/squidpy/experimental/im/_make_tiles.py
timtreis: I prefer not to code-golf at the sake of readability. Generally I find the asser thing, text pattern ugly

src/squidpy/experimental/im/_make_tiles.py
timtreis: how so?

src/squidpy/experimental/im/_make_tiles.py
selmanozleyen: It's not just code golfing, it makes sense as a function itself. When the error message or access pattern changes you could change it from the function instead modifying it everywhere. Also I find it more readable this way. I can just skip reading those functions as they aren't part of the core algo. These kinds of trivial checks bloat the core function.

Idk something like this
```python
def check_element_exists(
    sdata,
    key,
    element_type,
):
    store = getattr(sdata, element_type)
    if key not in store:
        available = list(store.keys())
        raise KeyError(
            f"Key {key!r} not found in sdata.{element_type}. "
            f"Available keys: {available}"
        )
```

src/squidpy/experimental/im/_make_tiles.py
selmanozleyen: something like this

```python
def check_within_bound(name,val,min,max):
    if not min <= val <= max:
        raise ValueError(f"{name} must be in ..")
```
This way you can also modify if you want to call it `"min_tissue_fraction"` or `"Min tissue fraction"` for other use cases.

src/squidpy/experimental/im/_make_tiles.py
selmanozleyen: I am wondering if we should just let it error from spatialdata? Like the error message would be almost the same

timtreis: > I am wondering if we should just let it error from spatialdata? Like the error message would be almost the same

I see the rational but dislike it in this case because sdata core has quite a few error messages that a biologist probably wouldn't intuitively understand and they also usually don't proactively offer help by f.e. giving the valid keys. There's also quite a few empty asserts which I find quite problematic.

> When the error message or access pattern changes you could change it from the function instead modifying it everywhere. Also I find it more readable this way. I can just skip reading those functions as they aren't part of the core algo. These kinds of trivial checks bloat the core function.

I do see that point, but I think then it should be a separate PR where we do this across the entire codebase. wdyt?

selmanozleyen: Sure it can be a separate PR but isn't this Pr already small. Why not start here? 

Update: sorry was in the bus so the comment was incomplete bc of that

src/squidpy/experimental/im/_make_tiles.py
selmanozleyen: Or `check_valid_fraction` where you dont need to add the bounds and is 0 to 1 by default which is reasonable imo. Since it's used more than once won't bloat the core algo and can be even used more by other functions potentially.

selmanozleyen (changes requested):
sorry but I stand by the comments I made earlier. 

From the other codes I have read, I find these changes reasonable to expect.

For example see:
https://github.com/scverse/annbatch/blob/d47240f883f64b86524d6bc32731ce3e15be2984/src/annbatch/utils.py#L114

I don't think this is overengineering or anything I think it is a necessity

selmanozleyen:
> I just think it's suboptimal to do one function here and then others in a separate PR. If we do something like that, it should be across the entire package. I can just do that in this PR and rename it, but I'd rather not sneak in such a major refactor under "added error handling"

ok if you say so. I just feel like there are already lot's one can refactor we might as well start when we add new code but if you think you can do the PR then sure.

selmanozleyen (approved):

PR #1132, opened 2026-03-02
chore: ligrec add explicit reproducibility tests

selmanozleyen:
Adds file for pval reprod tests not to rely on plotting tests as it should be independent.

Also removes kwargs from ligrec to make the taken parameters more explicit.

No longer does these to be backwards compatible for now:
~https://github.com/scverse/squidpy/issues/1131 and creates new files to check full reproducibility.~

~This PR is also important because after this we can verify if  https://github.com/scverse/squidpy/pull/1125 works or not.~


ilan-gold (commented):
So this PR will break reproducibility? Just judging by the `_images` changed

ilan-gold:
> Backwards reproducibility: yes. Images were the only thing to test reproducibility but if matplotlib had an update it would also fail so I also added proper reprod tests

I'm not sure I follow here - "Backwards reproducibility" seems like the only thing that would matter here.  I don't think we would want new installations of `squidpy` after this PR to change results when the same function is run on the same data.  Isn't that what would happen here?

>  Images were the only thing to test reproducibility but if matplotlib had an update it would also fail so I also added proper reprod tests

But has matplotlib had an update? Why weren't the tests proper before?

tests/conftest.py
```diff
@@ -263,6 +263,12 @@ def ligrec_no_numba() -> Mapping[str, pd.DataFrame]:
         return {"means": data[0], "pvalues": data[1], "metadata": data[2]}
 
 
+@pytest.fixture(scope="session")
+def ligrec_pvalues_reference() -> Mapping[str, pd.DataFrame]:
+    with open("tests/_data/ligrec_pvalues_reference.pickle", "rb") as fin:
+        return pickle.load(fin)
```
ilan-gold: Can we store this as a zarr file or hdf5 or even just an anndata?

selmanozleyen:
I added this feature in https://github.com/scverse/squidpy/pull/1125 so no need for this PR anymore

PR #1133, opened 2026-03-02
Fix SpatialData table access and and fix getting full key names

nilchia:

## Description

1. Fixes an AttributeError when using SpatialData objects with squidpy graph functions. The error occurred because the code was trying to access `.table` attribute on SpatialData objects, but SpatialData uses `.tables`, which is a dictionary.

The fix was adding the `table_key` as an argument.
The error was:
```
AttributeError: 'SpatialData' object has no attribute 'table'. Did you mean: 'tables'?
```

2. Fixes a bug where passing the full connectivity key name (e.g., 'spatial_connectivities') to functions like `sq.gr.nhood_enrichment()` would cause a KeyError due to suffix duplication.

The error was:
```
>>> sq.gr.nhood_enrichment(
...     adata,
...     cluster_key='region',
...     connectivity_key='spatial_connectivities',  # Full key name
...     ...
... )
KeyError: "Spatial connectivity key `spatial_connectivities_connectivities` not found in `adata.obsp`. 
Please run `squidpy.gr.spatial_neighbors(..., key_added='spatial')` first."
```

## How has this been tested?

Tested locally.


nilchia:
@selmanozleyen Thank you.
I'll try to check if other functions work with spatialdata as well and will keep you posted.

Could you please tell me which functions will be deprecated in squidpy? I see some plotting functions will be moved to spatialdata-plot.

I've already added the spatialdata tool suite to Galaxy [here](https://github.com/galaxyproject/tools-iuc/pull/7455)
and I'm adding squidpy [here](https://github.com/galaxyproject/tools-iuc/pull/7730)

PR #1134, opened 2026-03-02
chore: fix spatialdata access in squidpy via a helper function

selmanozleyen:
Fixes  https://github.com/scverse/squidpy/pull/1133 and implements the table extraction in a decorator so we won't need to ctrl+f our way to fix a similar bug when it happens

update: decided not to use a decorator as it uses metaprogramming with little to no benefit over using a function. Only plus was not adding `table_key` to the signature which isn't a plus imo. Explicitly typing a signature of a function is not something I consider as duplicate code


src/squidpy/gr/_utils.py
```diff
@@ -14,13 +14,42 @@
...
 from scipy.sparse import csc_matrix, csr_matrix, spmatrix
+from spatialdata import SpatialData
 
 from squidpy._compat import ArrayView, SparseCSCView, SparseCSRView
 from squidpy._docs import d
 from squidpy._utils import NDArrayA
 from squidpy._validators import assert_non_empty_sequence
 
 
+def extract_adata(adata: AnnData | SpatialData, *, table_key: str = "table") -> AnnData:
```
timtreis: Not sure if `extract` captures the intent here, given that it might already be an `adata`? Maybe `assure_adata`?

src/squidpy/gr/_sepal.py
```diff
@@ -43,6 +44,8 @@ def sepal(
     n_jobs: int | None = None,
     backend: str = "loky",
     show_progress_bar: bool = True,
+    *,
+    table_key: str = "table",
```
timtreis, line 48: This is no longer convention in `SpatialData` due to inherently multi-table techs like VisiumHD or CosMx 🤔 But it might be good to just let the user run into the error msg that tells them about their options if they don't specify it so I think we can leave it like this

src/squidpy/gr/_niche.py
```diff
@@ -169,12 +170,8 @@ def calculate_niche(
     if resolutions is None:
         resolutions = [0.5]
 
-    if isinstance(data, SpatialData):
-        orig_adata = data.tables[table_key]
-        adata = orig_adata.copy()
-    else:
-        orig_adata = data
-        adata = data.copy()
+    orig_adata = extract_adata(data, table_key=table_key or "table")
```
timtreis: Why do we have a different logic here than in the other function where `table` is just the default arg?

src/squidpy/gr/_build.py
```diff
@@ -436,6 +397,62 @@ def _build_connectivity(
...
+    elements_to_coordinate_systems: dict[str, str] | None,
+    table_key: str = "table",
+    spatial_key: str = Key.obsm.spatial,
+    library_key: str | None = None,
+) -> tuple[AnnData, str | None]:
+    if not isinstance(adata, SpatialData):
+        return adata, library_key
+
+    assert elements_to_coordinate_systems is not None, (
+        "Since `adata` is a :class:`spatialdata.SpatialData`, `elements_to_coordinate_systems` must not be `None`."
```
timtreis, line 409: "adata is sdata" is weird? Maybe it should be `resolve_data/input`?

src/squidpy/gr/_niche.py
```diff
@@ -843,18 +840,14 @@ def _validate_niche_args(
...
-            if library_key not in data.obs.columns:
-                raise ValueError(f"'library_key' must be a column in 'adata.obs', got {library_key}")
-        elif isinstance(data, SpatialData):
-            if table_key is None:
-                raise ValueError("'table_key' is required when 'data' is a SpatialData object")
-            if table_key not in data.tables:
-                raise ValueError(f"'table_key' must be a valid table key in 'data', got {table_key}")
-            if library_key not in data.tables[table_key].obs.columns:
-                raise ValueError(f"'library_key' must be a column in 'adata.obs', got {library_key}")
+        adata = extract_adata(data, table_key=table_key or "table")
```
timtreis: Let extract_adata handle that

src/squidpy/gr/_build.py
```diff
@@ -436,6 +397,62 @@ def _build_connectivity(
...
+    for region_ in ordered_regions_in_table:
+        cs = elements_to_coordinate_systems[region_]
+        centroid = get_centroids(adata[region_], coordinate_system=cs)[["x", "y"]].compute()
+
+        # TODO: remove this after https://github.com/scverse/spatialdata/issues/614
+        if remove_centroids[region_]:
+            centroid = centroid[1:].copy()
+        centroids.append(centroid)
+
+    table.obsm[spatial_key] = np.concatenate(centroids)
```
timtreis, line 450: Does this mutate in-place?

docs/conf.py
```diff
@@ -49,7 +49,7 @@
     numpy=("https://numpy.org/doc/stable", None),
     statsmodels=("https://www.statsmodels.org/stable", None),
     scipy=("https://docs.scipy.org/doc/scipy", None),
-    pandas=("https://pandas.pydata.org/pandas-docs/stable", None),
```
timtreis, line 52: is this drive-by?

timtreis (changes requested):

src/squidpy/gr/_build.py
selmanozleyen: yes but this is from old code, we can create an issue to fix this later

src/squidpy/gr/_build.py
selmanozleyen: yes but the error message would still have to be the same since the public signature takes adata, otherwise the error message would be confusing

src/squidpy/gr/_niche.py
selmanozleyen: I am not sure, it was already like this. Let's make it table by default then?

src/squidpy/gr/_niche.py
selmanozleyen: true

src/squidpy/gr/_utils.py
selmanozleyen: what about `maybe_extract_adata`?

src/squidpy/gr/_build.py
timtreis: Yeah, create issue pls

docs/conf.py
selmanozleyen: yeah I think so, I don't know why it was erroring without this locally

src/squidpy/gr/_build.py
selmanozleyen: this is done in many places lets make an issue about it and move on for here. Ideally we would call it data since it's positional argument anyway

src/squidpy/gr/_utils.py
selmanozleyen: calling it extract_adata_if_sdata

src/squidpy/gr/_build.py
selmanozleyen: oky done

src/squidpy/gr/_sepal.py
flying-sheep: isn’t using `"table"` as default misleading then?

flying-sheep (approved):
Looks great apart from the table issue: https://github.com/scverse/squidpy/pull/1134#discussion_r3026696839

I think since there’s no default anymore, it might make sense to require specifying `table=...` when `adata` is a SpatialData (and raise a `TypeError` otherwise, as Python does when a required parameter is missing)

selmanozleyen:
Thank you @flying-sheep ! Also what do you think of changing the parameter name from `adata` to `data`? Since it's a positional argument would the change be smooth?

flying-sheep:
Makes sense to me! Make sure to add a `/` to the parameter specification to enforce that people aren’t actually using the name.

And put the `/` after *all* the “main” parameters if there are more than just `data`. Depends on your judgement what a “main” parameters are – it’s often the same list as the mandatory parameters, unless a function has too many mandatory parameters, or multiple mandatory parameters have the same type and the order could be confusing.

I think most good APIs have 1–2 positional-only parameters (might have defaults), and some keyword-only parameters that all have defaults. Most functions have zero positional-or-keyword parameters, but for some it makes sense.

e.g. `var_by_distance` would look like this:

```python
def var_by_distance(
    adata: AnnData,
    groups: str | list[str] | NDArrayA,
    /,
    *,
    cluster_key: str | None = None,
    ...
) -> ...:
```

whereas `nhood_enrichment` would look like this, so people can call it either as `nhood_enrichment(adata, "k")` or `nhood_enrichment(adata, cluster_key=..., library_key=...)` to clarify which key is which.

```python
def nhood_enrichment(
    data: AnnData | SpatialData,
    /,
    cluster_key: str,
    *,
    library_key: str | None = None,
    connectivity_key: str | None = None,
    ...
) -> ...:
```

selmanozleyen:
Hi @timtreis is this good for merge?

timtreis (approved):

PR #1139, opened 2026-03-10
Consolidate validation utilities into squidpy/_validators.py

timtreis:
## Summary

- Create canonical `squidpy/_validators.py` module for generic validation utilities, fixing layering violations where `pl/`, `im/`, and `tl/` modules imported from `gr/_utils.py`
- Move 6 existing validators (`_assert_positive`, `_assert_non_negative`, `_assert_in_range`, `_assert_non_empty_sequence`, `_check_tuple_needles`, `_get_valid_values`) from `gr/_utils.py` with re-exports for backwards compatibility
- Add 4 new validators (`_assert_isinstance`, `_assert_one_of`, `_assert_key_in_adata`, `_assert_key_in_sdata`) and replace ~20 ad-hoc isinstance checks in `gr/_niche.py`, ~12 sdata key checks in `_make_tiles.py`, and scattered key/range checks across `pl/` and `gr/` modules

Follows up on the discussion in #1130.

## Test plan

- [x] `tests/test_validators.py` — 37 new unit tests covering all validator functions + edge cases + re-export identity
- [x] `tests/graph/test_utils.py` — 8 existing tests still pass (45 total)
- [x] Full CI test suite passes
- [x] Pre-commit (ruff/isort) passes


src/squidpy/experimental/im/_make_tiles.py
```diff
@@ -17,6 +17,7 @@
 from spatialdata.transformations import get_transformation, set_transformation
 
 from squidpy._utils import _yx_from_shape
+from squidpy._validators import _assert_in_range, _assert_key_in_sdata, _assert_positive
```
selmanozleyen: can we make these functions public within the module e.g. `assert_in_range` ? the module `_validators` is already private

src/squidpy/experimental/im/_make_tiles.py
```diff
@@ -780,10 +765,7 @@ def _make_tiles(
         return _TileGrid(H, W, tile_size=tile_size)
 
     # Path 2: Center grid on tissue mask centroid
-    if image_mask_key not in sdata.labels:
-        raise KeyError(
-            f"Mask key '{image_mask_key}' not found in sdata.labels. Available keys: {list(sdata.labels.keys())}"
-        )
+    _assert_key_in_sdata(sdata, image_mask_key, attr="labels")
```
selmanozleyen: so the difference between assert and raising is assert can technically be disabled for production level code. This isn't a thing in python probably but calling it assert is technically not right and can be misleading

timtreis: I think I'm more using it here as the English "assert that sth is xyz" and less so the Synthax `assert` 😅

src/squidpy/experimental/im/_make_tiles.py
timtreis: Yeah can do 👌

src/squidpy/experimental/im/_make_tiles.py
selmanozleyen: yeah I thought so too but still wanted to point it out. I don't mind honestly but I'd prefer validate

timtreis: hm, I'll leave it "assert" here I think. To me, "validate" implies more in-depth logic, f.e. checking that it's actually a valid object to perform the operation on etc

src/squidpy/_validators.py
```diff
@@ -52,15 +52,15 @@ def assert_non_empty_sequence(
...
+    if not res:
         raise ValueError(f"No {name} have been selected.")
 
     return res
 
 
 def get_valid_values(needle: Sequence[Any], haystack: Sequence[Any]) -> Sequence[Any]:
     res = [n for n in needle if n in haystack]
-    if not len(res):
+    if not res:
```
selmanozleyen: This is something I often notice with Claude that is stubborn but is commonly known as bad practice. Here `len(res)>0` or something explicit is needed.

Like checks for 0 or None shouldn't be done this say. Only if res is a boolean this should be written this way

selmanozleyen: @timtreis only this is left

selmanozleyen (commented):
This PR is very satisfying. Only need to use `len(res)==0` then it's all good

selmanozleyen (approved):

PR #1140, opened 2026-03-12
chore: update pytest config for pytest 9

flying-sheep:
just make the pytest config a bit prettier

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -225,17 +225,21 @@ lint.flake8-tidy-imports.ban-relative-imports = "all"
 lint.isort.required-imports = [ "from __future__ import annotations" ]
 
 [tool.pytest]
-ini_options.filterwarnings = [
+strict = true
+filterwarnings = [
   "error::numba.NumbaPerformanceWarning",
   "ignore::UserWarning",
   "ignore::anndata.OldFormatWarning",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -225,18 +225,22 @@ lint.flake8-tidy-imports.ban-relative-imports = "all"
...
-ini_options.python_files = "test_*.py"
-ini_options.testpaths = [ "tests/" ]
-ini_options.xfail_strict = true
-ini_options.addopts = [
+python_files = [ "test_*.py" ]
+testpaths = [ "tests/" ]
+addopts = [
   "--ignore=docs",
 ]
+markers = [
```
selmanozleyen, line 240: we don't support gpu marker yet I think

flying-sheep: you must import some unofficial anndata testing utils then which causes this marker to be defined as a side effect.

selmanozleyen: yes true we use `assert_adata_equal` which isn't public it turns out

selmanozleyen: opened an issue about it https://github.com/scverse/squidpy/issues/1143

selmanozleyen (approved):
thanks! I created an issue for using the private test func

PR #1144, opened 2026-03-23
feat: parallelize per-library spatial graph construction

Marius1311:
Add `n_jobs` parameter to `spatial_neighbors()` to compute per-library graphs in parallel via joblib. Defaults to 1 (sequential, no behavior change). Set to -1 to use all CPUs.

When `library_key` is set, each library's graph is already computed independently, so this is a trivially parallel workload. For datasets with many libraries (e.g., multi-sample spatial transcriptomics), this gives a near-linear speedup.

### DIFF
--- a/src/squidpy/gr/_build.py
+++ b/src/squidpy/gr/_build.py
@@ -14,6 +14,7 @@ import pandas as pd
 from anndata import AnnData
 from anndata.utils import make_index_unique
 from fast_array_utils import stats as fau_stats
+from joblib import Parallel, delayed
 from numba import njit, prange
 from scipy.sparse import (
     SparseEfficiencyWarning,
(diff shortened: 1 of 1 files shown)


selmanozleyen:
Thank you but we are working on making updates to `spatial_neighbours`: https://github.com/scverse/squidpy/pull/1147 Stay tuned, after these changes your version should be trivial to implement!

grst:
Hi @Marius1311, 

#1147 has been merged. 
Your implementation would have to be adapted accordingly, to parallelize this section: 
https://github.com/scverse/squidpy/blob/da789d0b10fb8436ac9aa26080a45464430eb1f3/src/squidpy/gr/_build.py#L793-L799

Do you want to work on this?

Marius1311:
thanks @grst! In my own tests, my parallelization actually didn't lead to any speed improvements, not sure why!

Marius1311:
So we can close this if you like.

grst:
I observed something similar, and I think a large part of the runtime is spent on subsetting the spatialdata/anndata object before any nearest neighbor code is even triggered. 

I still think it should be possible to optimize this, but then it's likely not a trivial fix as thought initially.

grst:
Closing in favor of https://github.com/scverse/squidpy/pull/1198

PR #1147, opened 2026-04-02
spatial_neigbours extensibility features and clarification

### LINKED ISSUE #1102: Documentation for `sq.gr.spatial_neighbors()` likely needs more clarification about the interplay of arguments
### Description of feature

Hello, I think that it would be beneficial to have more explanation about the arguments of `sq.gr.spatial_neighbors()`, like `coord_type`, `n_neighs`, `radius`, `n_rings`, etc. in the documentation. Current documentation (https://squidpy.readthedocs.io/en/stable/api/squidpy.gr.spatial_neighbors.html#squidpy.gr.spatial_neighbors) does a good job of explaining their usage for the most part, but behaviour for some combinations of these arguments are not super clear from it. For example:

- For `coord_type = 'grid'`, it is not clear what would happen if for a hexagonal grid, someone provided `n_neighs = 5`. Which out of the neighbors will be excluded. Maybe we can limit `n_neighs` to take values from {4,6} only (as only they form a continuous grid), when `coord_type = 'grid'`, or give a warning when that is not the case? In general, it would be nice to know that for a given ring, in which order are spots chosen from it (eg- clockwise, starting from right).
- It would be helpful to make more explicit, that the neighborhood of a cell never includes itself. So neighborhood enrichment never considers the celltype of center cell itself. I had to double check this for my usage, but I don't know if others also faced such a confusion ever.
- For `coord_type = 'generic'`, what would happen if one supplies both, `n_neighs` and `radius`? How will the neighborhood be determined then?

I believe that it could be useful to have a separate tutorial explaining the interplay of all these arguments for spot-based and image-based data, especially because `sq.gr.spatial_neighbors()` is such an important first step in most of squidpy's functionality. If not, then having more details in the documentation of `sq.gr.spatial_neighbors()` would definitely be super useful.

@selmanozleyen, let me know what you think. If you think this could be useful, I am willing to give a shot and submit a PR for the same. Just wanted to assess the maintainers' opinion before I do that.

Thanks a lot!

selmanozleyen:
fixes: https://github.com/scverse/squidpy/issues/1102 and https://github.com/scverse/squidpy/issues/1047 

It's backward compatible and I am curious what the community might bring to this!

- I clarified the docs on argument precedence
- Added extensibility features
- Added tests to ensure the function behaves as written in the docs
- No breaking changes


src/squidpy/gr/_build.py
```diff
@@ -107,29 +160,107 @@ def spatial_neighbors(
...
+          ``coord_type='generic'``, ``delaunay=False``, ``radius`` set.
+          Uses ``radius`` and builds a radius-based neighbor graph.
+          If ``radius`` is a tuple, the graph is built with the maximum
+          radius and then pruned to the interval
+          ``[min(radius), max(radius)]``.
+        - Generic Delaunay mode:
+          ``coord_type='generic'``, ``delaunay=True``.
+          Builds a Delaunay triangulation graph. ``n_neighs`` is not
+          used for the triangulation itself, but is still accepted for
+          backward compatibility. If ``radius`` is a tuple, it is used
```
grst: what means "accepted"? Does it change the results in any way or is it ignored?

src/squidpy/gr/_build.py
```diff
@@ -107,29 +160,107 @@ def spatial_neighbors(
...
+        ``percentile``, ``transform``, ``set_diag``) must be left as ``None``.
     key_added
         Key which controls where the results are saved if ``copy = False``.
     %(copy)s
 
+    Notes
+    -----
+    ``spatial_neighbors`` has 4 graph-construction modes:
+
+        - Grid mode:
```
grst, line 280: This documentation helps a lot in clarifying the different modes, but it doesn't solve the underlying problem that the original API design is confusing. 

It of course depends to what extent you want to make API modifications or keep such things for squidpy 2.0, but have you considered deprecating the function altogether (it could still be kept for backwards-compatibility, but internally delegate to the builders) and build a new API instead?

I'm thinking along the lines of either

```python
def spatial_neighbors(mode: Literal["grid", "knn", "radius", "delaunay"] | GraphBuilder, **kwargs):
    """Determine the GraphBuilder based on mode or pass one directly -- kwargs is passed to the GraphBuilder"""
    pass
```

or 

```python
def spatial_neighbors_grid(n_neighs: int, n_rings: int): 
    pass

def gr.spatial_neighbors_knn(n_neighs: int):
    pass

gr.spatial_neighbors_delauney():
    pass

gr.spatial_neighbors_radius(radius: float):
    pass
```

src/squidpy/gr/neighbors.py
```diff
@@ -0,0 +1,431 @@
...
+
+- :class:`KNNBuilder` -- k-nearest-neighbor graph.
+- :class:`RadiusBuilder` -- radius-based graph.
+- :class:`DelaunayBuilder` -- Delaunay triangulation graph.
+- :class:`GridBuilder` -- grid (Visium-style) graph.
+
+To implement a custom strategy, subclass :class:`GraphBuilder` and override
+:meth:`~GraphBuilder._build_graph`.  The returned object can then be passed
+to ``spatial_neighbors(adata, builder=my_builder)``.
+"""
```
grst: Can this be shown in the sphinx docs somewhere?

src/squidpy/gr/neighbors.py
```diff
@@ -0,0 +1,431 @@
...
+
+        return self._apply_transform(Adj), Dst
+
+    @abstractmethod
+    def _build_graph(self, coords: NDArrayA) -> tuple[csr_matrix, csr_matrix]:
+        """Construct raw adjacency and distance matrices."""
+
+    def _apply_filters(self, Adj: csr_matrix, Dst: csr_matrix) -> None:
+        """Apply builder-specific post-processing filters."""
+        return None
```
grst: Is it possible to add the private functions to the sphinx docs? For developers who want to override these functions it would be good to know these.

src/squidpy/gr/neighbors.py
```diff
@@ -0,0 +1,431 @@
...
+        if return_distance:
+            Dst = csr_matrix((dists, (row_indices, col_indices)), shape=(N, N))
+
+    # radius-based filtering needs same indices/indptr: do not remove 0s
+    Adj.setdiag(1.0 if set_diag else Adj.diagonal())
+    if return_distance:
+        Dst.setdiag(0.0)
+        return Adj, Dst
+
+    return Adj
```
grst: Any reason not to split the `if` up and put it into the respective `Builder` classes? Like this the branching logic is just moved to a lower level.

src/squidpy/gr/neighbors.py
```diff
@@ -0,0 +1,431 @@
...
+    radius: float | tuple[float, float] | None
+    n_neighs: int
+
+    @property
+    def coord_type(self) -> CoordType:
+        return CoordType.GENERIC
+
+    def _apply_filters(self, Adj: csr_matrix, Dst: csr_matrix) -> None:
+        if isinstance(self.radius, Iterable):
+            _filter_by_radius_interval(Adj, Dst, self.radius)
```
grst: Not sure if this is worth the additional abstraction layer? This essentially just adds a single function call to `_filter_by_radius_interval` which could be easily added to the respective classes directly.

src/squidpy/gr/neighbors.py
```diff
@@ -0,0 +1,431 @@
...
+    minn, maxx = sorted(radius)[:2]
+    mask = (Dst.data < minn) | (Dst.data > maxx)
+    a_diag = Adj.diagonal()
+
+    Dst.data[mask] = 0.0
+    Adj.data[mask] = 0.0
+    Adj.setdiag(a_diag)
+
+
+def _build_grid(
```
grst: I would move all logic that is builder-specific (including specific helper funcitons) inside the respective builder classes.

src/squidpy/gr/neighbors.py
```diff
@@ -0,0 +1,431 @@
...
+) -> tuple[csr_matrix, csr_matrix]:
+    if n_rings > 1:
+        Adj: csr_matrix = _build_connectivity(
+            coords,
+            n_neighs=n_neighs,
+            neigh_correct=True,
+            set_diag=True,
+            delaunay=delaunay,
+            return_distance=False,
+        )
```
grst: to absorb the `_build_connectivity` function into the builder classes as suggested below, would it be an option to call a `Builder` directly here?

grst (commented):
Thanks @selmanozleyen, that goes in the right direction. I put up some design questions up for discussion!

src/squidpy/gr/_build.py
selmanozleyen: Latter is also what we discussed with Tim! I agree that it makes more sense. I thought it could be another PR on top but while I have your attention let me also incorparate that

src/squidpy/gr/_build.py
selmanozleyen: It is effectively ignored. I should add future warning to deprecate it later!

src/squidpy/gr/neighbors.py
selmanozleyen: We already have lot's in a PR. Basically I wanted it to be easily reviewable.But the branches can go to classes almost as is and since you are fine with it I will do that

src/squidpy/gr/neighbors.py
selmanozleyen: actually agree

src/squidpy/gr/neighbors.py
selmanozleyen: I didn't expose them so I can iterate over them but I let me expose them directly then. I will move this from a module doc to an example page.

src/squidpy/gr/_build.py
```diff
@@ -140,70 +348,329 @@ def spatial_neighbors(
...
-            "Since `adata` is a :class:`spatialdata.SpatialData`, `elements_to_coordinate_systems` must not be `None`."
+    if builder is None:
+        warnings.warn(
+            "Calling `spatial_neighbors` without a `builder` argument is deprecated "
+            "and will be removed in squidpy v1.7.0. Use one of the mode-specific "
+            "functions instead: `spatial_neighbors_knn`, `spatial_neighbors_radius`, "
+            "`spatial_neighbors_delaunay`, or `spatial_neighbors_grid`.",
+            FutureWarning,
+            stacklevel=2,
         )
```
shashkat: Here, it may also be useful to give a warning if someone specifies a builder object and also specifies other arguments like n_neighs. As in such a case, those arguments won't have an effect, and the logic of graph building would come only from the builder object.

src/squidpy/gr/_build.py
```diff
@@ -140,70 +348,329 @@ def spatial_neighbors(
...
+            transform=transform,
+            set_diag=set_diag,
         )
-        elements, table = match_element_to_table(adata, list(elements_to_coordinate_systems), table_key)
-        assert table.obs_names.equals(adata.tables[table_key].obs_names), (
-            "The spatialdata table must annotate all elements keys. Some elements are missing, please check the `elements_to_coordinate_systems` dictionary."
+    else:
+        n_neighs = n_neighs if n_neighs is not None else 6
+        delaunay = delaunay if delaunay is not None else False
+        n_rings = n_rings if n_rings is not None else 1
```
shashkat: Minor suggestion: We may not need the overriding lines for n_neighs and n_rings. This else block would be entered only when the user doesn't specify builder object to `spatial_neighbors()` I believe. And in that case n_neighs and n_rings would already have values 6 and 1 respectively if not supplied by the user. 
If default values of n_neighs or n_rings is changed in future, then this part can possibly cause issues as it will override the (different) defaults.

src/squidpy/gr/neighbors.py
```diff
@@ -0,0 +1,383 @@
...
+        self.n_neighs = n_neighs
+
+    @property
+    def coord_type(self) -> CoordType:
+        return CoordType.GENERIC
+
+    def build_graph(self, coords: NDArrayA) -> tuple[csr_matrix, csr_matrix]:
+        N = coords.shape[0]
+        r = self.radius if isinstance(self.radius, int | float) else max(self.radius)
+        tree = NearestNeighbors(n_neighbors=self.n_neighs, radius=r, metric="euclidean")
```
shashkat: Since we are using `tree.radius_neighbors()` below, is the `n_neighbors = self.n_neighs` having an effect here? 
Is it intentional to have some of these arguments still valid for RadiusBuilder, DelaunayBuilder etc? I see that for the parameter n_neighs, there is a warning in DelaunayBuilder that it will be ignored. If its intentional to keep these arguments for now, then possibly can add warning for n_neighs in RadiusBuilder too.

src/squidpy/gr/_build.py
```diff
@@ -59,6 +64,117 @@ class SpatialNeighborsResult(NamedTuple):
...
+        coord_type = CoordType.GRID if has_spatial_uns else CoordType.GENERIC
+    else:
+        coord_type = CoordType(coord_type)
+
+    common: dict[str, Any] = {
+        "n_neighs": n_neighs,
+        "transform": transform,
+        "set_diag": set_diag,
+    }
+
```
shashkat, line 97: WRT the removal of non necessary arguments from RadiusBuilder, DelaunayBuilder etc, whenever it is decided to remove them (now or later), it would also be nice to make the logic here more explicit. By more explicit logic I mean `if` `else` hierarchy closely resembling the decision tree we have possible. I made a rough version according to my understanding:

Image

shashkat (commented):
Thanks a lot for doing this :)
Here are some minor suggestions from me!

docs/extensibility.md
```diff
@@ -0,0 +1,76 @@
...
+### What to override
+
+| Method / property | Required | Purpose |
+|---|---|---|
+| {attr}`~squidpy.gr.neighbors.GraphBuilder.coord_type` | yes | Return the {class}`~squidpy._constants._constants.CoordType` this builder supports. |
+| {meth}`~squidpy.gr.neighbors.GraphBuilder.build_graph` | yes | Construct and return ``(Adj, Dst)`` as {class}`~scipy.sparse.csr_matrix` pair. |
+| {meth}`~squidpy.gr.neighbors.GraphBuilder.apply_filters` | no | Post-processing on the raw ``Adj``/``Dst`` (e.g. radius-interval pruning). Called before percentile filtering and transform. |
+
+The base class handles percentile filtering, adjacency transforms, and
+{class}`~scipy.sparse.SparseEfficiencyWarning` suppression automatically.
```
grst: Could you include the definition of `Ast` and `Dst` somewhere please? I.e. what should they contain, whay type should they have?

docs/extensibility.md
```diff
@@ -0,0 +1,76 @@
...
+
+        Adj = csr_matrix(
+            (np.ones(len(row), dtype=np.float32), (row, col)),
+            shape=(N, N),
+        )
+        Dst = csr_matrix((d, (row, col)), shape=(N, N))
+
+        Adj.setdiag(1.0 if self.set_diag else Adj.diagonal())
+        Dst.setdiag(0.0)
+        return Adj, Dst
```
grst: Now that you implemented it already as an example, why not make it generally available via the API? 
But that's for another PR :)

src/squidpy/gr/_build.py
```diff
@@ -140,70 +348,329 @@ def spatial_neighbors(
         - :attr:`anndata.AnnData.obsp` ``['{{key_added}}_distances']`` - the spatial distances.
         - :attr:`anndata.AnnData.uns`  ``['{{key_added}}']`` - :class:`dict` containing parameters.
     """
-    if isinstance(adata, SpatialData):
-        assert elements_to_coordinate_systems is not None, (
-            "Since `adata` is a :class:`spatialdata.SpatialData`, `elements_to_coordinate_systems` must not be `None`."
+    if builder is None:
+        warnings.warn(
```
grst: Is there any reason we would not deprecate it entirely in favor of using the new functions, or a builder class directly?

src/squidpy/gr/_build.py
grst: Since all of this is deprecated anyway, I wouldn't put too much effort into making this part "nice"

src/squidpy/gr/_build.py
```diff
@@ -140,70 +348,329 @@ def spatial_neighbors(
...
+    elements_to_coordinate_systems: dict[str, str] | None = None,
+    table_key: str | None = None,
+    library_key: str | None = None,
+    n_neighs: int = 6,
+    percentile: float | None = None,
+    transform: str | Transform | None = None,
+    set_diag: bool = False,
+    key_added: str = "spatial",
+    copy: bool = False,
+) -> SpatialNeighborsResult | None:
```
grst, line 489: For all of these new functions, I'd recommend to make everything but `adata` a keyword-only argument. That allows you later to reorder arguments without breaking backwards-compatibility.

src/squidpy/gr/_build.py
```diff
@@ -140,70 +348,329 @@ def spatial_neighbors(
...
+    return _spatial_neighbors(
+        adata,
+        spatial_key=spatial_key,
+        elements_to_coordinate_systems=elements_to_coordinate_systems,
+        table_key=table_key,
+        library_key=library_key,
+        builder=builder,
+        key_added=key_added,
+        copy=copy,
+    )
```
grst: With `_spatial_neighbors` you are calling part of the old logic that we want to get rid of with these changes. 
I think you could call `_run_spatial_neighbors` directly without losing anything? 

(applies to all of the new user-facing functions)

src/squidpy/gr/_build.py
```diff
@@ -265,241 +725,13 @@ def spatial_neighbors(
 def _spatial_neighbor(
```
grst: Seems like an unneceesary abstraction

src/squidpy/gr/neighbors.py
```diff
@@ -0,0 +1,383 @@
...
+
+
+class GridBuilder(GraphBuilder):
+    """Build a grid-based spatial graph."""
+
+    def __init__(
+        self,
+        n_neighs: int = 6,
+        n_rings: int = 1,
+        delaunay: bool = False,
```
grst, line 310: What's the difference between a GridBuilder with Delauney mode and using a DelauneyBuilder directly?

grst (commented):
It's taking a nice shape now... mostly minor things at this point!

src/squidpy/gr/_build.py
selmanozleyen: Yes that is handled in fact we throw a ValueError

src/squidpy/gr/_build.py
selmanozleyen: But now we have None as default for them so this is the only place the defaults are resolved now.

selmanozleyen: Oh now I see what you mean the builders also have defaults within but those are for the deprecated API anyway. After that path it won't use builders

src/squidpy/gr/_build.py
selmanozleyen: makes sense

src/squidpy/gr/_build.py
selmanozleyen: new functions yes, but if we have builder abstraction we still need an entry function to execute builder because thats we extract sdata/adata in `spatial_neigbors`

src/squidpy/gr/_build.py
selmanozleyen: oky done

src/squidpy/gr/_build.py
grst: ok, I see what you mean. 
I'm wondering if from an API perspective it made more sense if that entry-function were part of the `gr.neighbors` module or even a static method on the `GraphBuilder` class. Not entirely sure though what's best. 

I'm inclined to keep the old and the new logic separately and keep `gr.spatial_neighbors` as a deprecated function that is only for backwards-compatibility that can be removed without any interdependencies in a 2.0 release. 

Ultimately I don't have a super strong opinion here and maybe worth asking @timtreis for his perspective.

src/squidpy/gr/neighbors.py
selmanozleyen: you are right it was a legacy thing. I removed it now

src/squidpy/gr/neighbors.py
selmanozleyen: In general
- DelaunayBuilder: generic point-cloud graph from a Delaunay triangulation
- GridBuilder(delaunay=True) : grid-mode graph that happens to use Delaunay only as its base adjacency.

Distance wise
- DelaunayBuilder computes real Euclidean edge lengths
- GridBuilder uses graph/ring distances, not geometric distances

docs/extensibility.md
selmanozleyen: ok done, I also renamed them to lowercase.

docs/extensibility.md
selmanozleyen: sure :) https://github.com/scverse/squidpy/issues/1153

docs/extensibility.md
```diff
@@ -0,0 +1,88 @@
...
+{class}`~scipy.sparse.SparseEfficiencyWarning` suppression automatically.
+
+Here ``adj`` and ``dst`` are square sparse matrices of shape ``(n_obs, n_obs)``
+with matching sparsity structure:
+
+- ``adj`` is the connectivity / adjacency matrix. Non-zero entries mark edges in
+  the graph, and built-in builders typically use ``1.0`` for present edges.
+- ``dst`` is the distance matrix for those same edges. For generic graphs this is
+  usually the Euclidean edge length. For grid builders it may instead encode
+  graph-distance semantics such as ring number.
```
grst, line 41: any recommendations on dtypes?

src/squidpy/gr/_build.py
grst: > I'm wondering if from an API perspective it made more sense if that entry-function were part of the gr.neighbors module or even a static method on the GraphBuilder class. Not entirely sure though what's best.

I think a good candidate could be a new function `gr.spatial_neighbors_from_builder(adata, builder)`, or `gr.spatial_neighbors_custom(adata, builder)`. That would be quite explicit and the old `gr.spatial_neighbors` function could be left as legacy-only.

src/squidpy/gr/_build.py
```diff
@@ -140,366 +367,361 @@ def spatial_neighbors(
...
+    table_key: str | None = None,
+    library_key: str | None = None,
+    radius: tuple[float, float] | None = None,
+    percentile: float | None = None,
+    transform: str | Transform | None = None,
+    set_diag: bool = False,
+    key_added: str = "spatial",
+    copy: bool = False,
+) -> SpatialNeighborsResult | None:
+    """Create a Delaunay triangulation graph from spatial coordinates.
```
grst, line 625: I think this could deserve 1-2 sentences how delauney triangulation works and what radius means in this case.

src/squidpy/gr/_build.py
```diff
@@ -140,366 +367,361 @@ def spatial_neighbors(
...
+    %(adata)s
+    %(spatial_key)s
+    %(sdata_params)s
+    %(library_key)s
+    n_neighs
+        Number of neighboring tiles. Defaults to ``6``.
+    n_rings
+        Number of rings of neighbors. Defaults to ``1``.
+    delaunay
+        Whether to compute the grid graph from Delaunay triangulation.
```
grst: I think a bit more verbose description would be nice and how n_neighs and n_rings work out and what difference it makes in practice to use delauney. 

For instance, I can't really imaging what would happen if I used n_neighs=3 with n_rings=2 on visium, or if I used 'grid' neighbors on "continuous" (e.g. xenium) coordinates.

src/squidpy/gr/neighbors.py
```diff
@@ -0,0 +1,368 @@
...
+        if self.transform == Transform.COSINE:
+            return cast(csr_matrix, _transform_a_cosine(adj))
+        if self.transform == Transform.NONE:
+            return adj
+
+        raise NotImplementedError(f"Transform `{self.transform}` is not yet implemented.")
+
+
+class KNNBuilder(GraphBuilder):
+    """Build a generic k-nearest-neighbor spatial graph."""
```
grst: Similar to the functions above, I think the classes would deserve a bit more detailed documentation on how the distance calculation works. Or at least refer to the respective high-level function to avoid redundancies.

src/squidpy/gr/neighbors.py
grst: thanks, would be good to have that in the docs!

grst (commented):
Only some doc suggestions left from my side. 

Apart from that LGTM, I'll leave final approval to @timtreis

docs/extensibility.md
selmanozleyen: I updated the API. On the GraphBuilder itself I wouldn't want to give such specification as it depends can change based on use case. But I added 

```
- For CSR-based builders, ``adj`` often behaves like a boolean or indicator
  matrix describing whether an edge is present, even if it is stored with a
  numeric dtype such as ``float32``. ``dst`` stores edge-associated values such
  as distances and will often use a floating-point dtype. The exact dtype choice
  is left to the builder implementation and may depend on performance, memory,
  and numerical accuracy requirements.
```
for csr builders

src/squidpy/gr/_build.py
selmanozleyen: I think this is the best way to go

selmanozleyen:
Thanks a lot @grst ! I really appreciate your reviews. I assesed them and even made a more general class to support Cupy arrays in future.

grst (approved):
Love it!

docs/extensibility.md
```diff
@@ -0,0 +1,104 @@
...
+provides a faster exact fixed-radius search based on PCA-based pruning. The
+example below swaps the backend while keeping full compatibility with the rest
+of the Squidpy graph pipeline:
+
+```python
+import numpy as np
+from scipy.sparse import csr_matrix
+from snnpy import build_snn_model
+
+from squidpy._constants._constants import CoordType
```
timtreis: Should we have the import of internal constants as the default behaviour?

src/squidpy/gr/_build.py
```diff
@@ -59,6 +64,127 @@ class SpatialNeighborsResult(NamedTuple):
...
+    common: dict[str, Any] = {"transform": transform, "set_diag": set_diag}
+
+    if coord_type == CoordType.GRID:
+        if percentile is not None:
+            raise ValueError(
+                "`percentile` is not supported for grid coordinates. It only applies to generic (non-grid) graphs."
+            )
+        return GridBuilder(n_neighs=n_neighs, **common, n_rings=n_rings, delaunay=delaunay)
+    if delaunay:
+        # below check should be removed once legacy mode spatial_neighbors is deprecated
```
timtreis: Rephrase as a `TODO ...` so IDEs pick it up

src/squidpy/gr/_build.py
```diff
@@ -59,6 +64,127 @@ class SpatialNeighborsResult(NamedTuple):
...
+        if n_neighs_was_set:
+            warnings.warn(
+                "Parameter `n_neighs` is ignored when `delaunay=True` and will be removed in squidpy v2.0.0.",
+                FutureWarning,
+                stacklevel=3,
+            )
+        return DelaunayBuilder(**common, radius=radius, percentile=percentile)
+    if radius is not None:
+        # below check should be removed once legacy mode spatial_neighbors is deprecated
```
timtreis: see above

src/squidpy/gr/_build.py
```diff
@@ -59,6 +64,127 @@ class SpatialNeighborsResult(NamedTuple):
...
+                "Parameter `n_neighs` is ignored when `radius` is set and will be removed in squidpy v2.0.0.",
+                FutureWarning,
+                stacklevel=3,
+            )
+        return RadiusBuilder(**common, radius=radius, percentile=percentile)
+    return KNNBuilder(n_neighs=n_neighs, **common, percentile=percentile)
+
+
+def _resolve_spatial_data(
+    adata: AnnData | SpatialData,
```
timtreis: I prefer the pattern of

A) Function runs on `AnnData` -> `adata`
B) Function runs on `SpatialData` -> `sdata`
C) Function runs on `AnnData` or `SpatialData`-> `data`

For C) you can then internally use `adata`/`sdata` where appropiate but it shouldn't be default to plug A into a variable that's named after B

src/squidpy/gr/_build.py
```diff
@@ -107,29 +243,100 @@ def spatial_neighbors(
...
     set_diag
         Whether to set the diagonal of the spatial connectivities to `1.0`.
+        Defaults to ``False``.
     key_added
         Key which controls where the results are saved if ``copy = False``.
     %(copy)s
 
+    Notes
+    -----
+    ``spatial_neighbors`` has 4 graph-construction modes:
```
timtreis, line 278: Do we have a good explanation of these somewhere?

src/squidpy/gr/_build.py
```diff
@@ -140,366 +347,470 @@ def spatial_neighbors(
...
-
-    start = logg.info(
-        f"Creating graph using `{coord_type}` coordinates and `{transform}` transform and `{len(libs)}` libraries."
+    warnings.warn(
+        "Calling `spatial_neighbors` is deprecated and will be removed in squidpy "
+        "v1.7.0. Use `spatial_neighbors_knn`, `spatial_neighbors_radius`, "
+        "`spatial_neighbors_delaunay`, `spatial_neighbors_grid`, or "
+        "`spatial_neighbors_from_builder` instead.",
+        FutureWarning,
+        stacklevel=2,
```
timtreis, line 356: I think your other future warning was `stacklevel=3`, how do you decide this?

src/squidpy/gr/_build.py
```diff
@@ -140,366 +347,470 @@ def spatial_neighbors(
...
-        "params": {
-            "n_neighbors": n_neighs,
-            "coord_type": coord_type.v,
-            "radius": radius,
-            "transform": transform.v,
-        },
-    }
+@d.dedent
+def spatial_neighbors_from_builder(
+    adata: AnnData | SpatialData,
```
timtreis: see other comment

src/squidpy/gr/_build.py
```diff
@@ -140,366 +347,470 @@ def spatial_neighbors(
...
+        builder=builder,
+        spatial_key=spatial_key,
+        library_key=library_key,
+        key_added=key_added,
+        copy=copy,
+    )
+
+
+def _prepare_spatial_neighbors_input(
+    adata: AnnData | SpatialData,
```
timtreis: see other comment

src/squidpy/gr/_build.py
```diff
@@ -140,366 +347,470 @@ def spatial_neighbors(
...
+    *,
+    spatial_key: str,
```
timtreis, line 458: Align annotations across functions

src/squidpy/gr/_build.py
```diff
@@ -140,366 +347,470 @@ def spatial_neighbors(
...
+        copy=copy,
+    )
 
 
-def symmetric_normalize_csr(adj: spmatrix) -> csr_matrix:
-    """
-    Return D^{-1/2} * A * D^{-1/2}, where D = diag(degrees(A)) and A = adj.
+@d.dedent
+def spatial_neighbors_grid(
+    adata: AnnData | SpatialData,
```
timtreis: see other comment

src/squidpy/gr/neighbors.py
```diff
@@ -0,0 +1,440 @@
...
+    ``GridBuilder(delaunay=True)``, this builder uses geometry-based connectivity
+    and stores real Euclidean edge lengths.
+
+    See :func:`~squidpy.gr.spatial_neighbors_delaunay` for the user-facing API or
+    :func:`~squidpy.gr.spatial_neighbors_from_builder` for direct builder usage.
+    """
+
+    def __init__(
+        self,
+        radius: float | tuple[float, float] | None = None,
```
timtreis, line 255: is this used?

src/squidpy/gr/neighbors.py
```diff
@@ -0,0 +1,440 @@
+"""Graph construction strategies for spatial neighbor graphs.
```
timtreis, line 1: Quite some duplicates code, can you use a factory or prototcol instead?

timtreis (changes requested):

src/squidpy/gr/neighbors.py
grst: Not sure if I agree. What exactly do you think needs deduplication here?

timtreis: 1) The two typevars [here](https://github.com/scverse/squidpy/pull/1147/changes/BASE..95012dc54ec158044d06fbf8bf54bdec2e53cec5#diff-56425659b6d6708bbff1eaa181d8271e8ab5b150b000dcb4c3529e2bb730393dR39) aren't modified anywhere. It could just be an ABC. 

2) `apply_percentile` is the only function using the `coord_type`. Since at the point of `_resolve_graph_builder` we decide on the downstream path, we don't need that guard. With that guard removed, the coord_type becomes superflous and would further simplify the class.

3) Both `apply_filters` are identical, could be deduplicated by moving it into `build` of the ABC.

4) `KNNBuilder.build_graph` and `RadiusBuilder.build_graph` share quite some overlap in building their nn-graphs. Just the delauny stuff would need to be routed differently.

but I guess the biggest win would be removing the non-CSR build option? Is this sth we'll realistically ever encounter?

src/squidpy/gr/neighbors.py
selmanozleyen: yes by `_filter_by_radius_interval`. If I did this from scratch maybe I'd make it more complicated by having this filtering as a separate transform but this feels more appropiate for old implementation

src/squidpy/gr/neighbors.py
timtreis: The factory/protocol comment doesn't make much sense in hind-sight after having gone through it in detail - still, quite some opportunity to remove LOCs

docs/extensibility.md
selmanozleyen: this is a good point, I will think about it in the week

src/squidpy/gr/_build.py
selmanozleyen: stacklevel 2: user code -> spatial_neighbors() -> warnings.warn(...)

vs

stacklevel 3: user code -> spatial_neighbors() -> _resolve_graph_builder() -> warnings.warn(...)

src/squidpy/gr/neighbors.py
selmanozleyen: > but I guess the biggest win would be removing the non-CSR build option?

I'd like to also give possibility to support cupy sparse or whatever sparse types jax might have for example

selmanozleyen: But I noticed I use scipy operations also in the builder path so let me just have a second look.

src/squidpy/gr/_build.py
timtreis: Do we have this written down somewhere? I haven't really thought about this yet so I wonder if we should add that to some "how to Squidpy" doc somewhere?

src/squidpy/gr/_build.py
```diff
@@ -627,17 +707,35 @@ def spatial_neighbors_grid(
     %(sdata_params)s
     %(library_key)s
     n_neighs
-        Number of neighboring tiles. Defaults to ``6``.
+        Number of neighboring tiles used to form the base grid connectivity.
+        Defaults to ``6``. On a Visium-like hexagonal grid, ``6`` corresponds to
+        the immediate surrounding spots, while smaller values such as ``3`` make
+        the first-ring graph deliberately sparser.
```
shashkat, line 713: Maybe it would be useful to mention that there is no guarantee of what subset of neighbors would be chosen from a given ring, if n_neighs is less than the actual number of neighbors of a cell in the grid.
Just like this snippet from the latest-commit documentation of spatial_neighbors function:
```
        - values such as ``n_neighs=4`` and ``n_neighs=6`` are the
          intended square-grid and hex-grid choices, respectively;
        - other values are accepted for backward compatibility, but
          their geometric interpretation is not guaranteed to match a
          continuous ring on the grid;
        - no clockwise or other within-ring ordering is part of the
          public API.
```

src/squidpy/gr/neighbors.py
```diff
@@ -25,70 +27,124 @@
...
-    @property
     @abstractmethod
-    def coord_type(self) -> CoordType:
-        """Coordinate system supported by this builder."""
+    def build_graph(self, coords: CoordT) -> tuple[GraphMatrixT, GraphMatrixT]:
+        """Construct raw adjacency and distance matrices."""
+
+    def postprocessors(self) -> Sequence[GraphPostprocessor[GraphMatrixT]]:
+        """Return post-build processing steps for ``(adj, dst)``."""
+        return self._postprocessors
```
shashkat, line 87: I am not sure if this method is offering any functionality. 
We can possibly call the attribute _postprocessors as postprocessors and replace line 77 from `for postprocessor in self.postprocessors():` to `for postprocessor in self.postprocessors:`.
What I understand is that this method can allow one to reorder or filter out the entities in _postprocessors attribute. But that can anyways be done when the postprocessors entities are created in the __init__ method of a subclass of GraphBuilder.

src/squidpy/gr/neighbors.py
```diff
@@ -25,70 +27,124 @@
...
         self,
         transform: str | Transform | None = None,
         set_diag: bool = False,
         percentile: float | None = None,
+        postprocessors: Sequence[GraphPostprocessor[GraphMatrixT]] = (),
     ) -> None:
         self.transform = Transform.NONE if transform is None else Transform(transform)
         self.set_diag = set_diag
         self.percentile = percentile
+        self._postprocessors: list[GraphPostprocessor[GraphMatrixT]] = list(postprocessors)
```
shashkat, line 73: Is the `transform` argument used here? I think the transformation requirements would be encoded in `postprocessors` anyways, so we can remove it.

src/squidpy/gr/neighbors.py
```diff
@@ -114,97 +175,107 @@ def build_graph(self, coords: NDArrayA) -> tuple[csr_matrix, csr_matrix]:
...
-                FutureWarning,
-                stacklevel=2,
-            )
-        else:
-            n_neighs = 6
-        assert_positive(n_neighs, name="n_neighs")
-        super().__init__(transform=transform, set_diag=set_diag, percentile=percentile)
+        postprocessors: list[GraphPostprocessor[csr_matrix]] = []
+        if isinstance(radius, tuple):
+            postprocessors.append(DistanceIntervalPostprocessor(tuple(sorted(radius))))
```
shashkat, line 262: If someone passes a single float or int to radius, DelaunayBuilder doesn't do anything with it I believe. Only if radius is a tuple[float, float], does it add the filter. I think we can fix this by doing something like the following.
change this:
```py
if isinstance(radius, tuple):
    postprocessors.append(DistanceIntervalPostprocessor(tuple(sorted(radius))))
```
to this:
``` py
if isinstance(radius, (int, float)):
    radius = (0, radius)
postprocessors.append(DistanceIntervalPostprocessor(tuple(sorted(radius))))
```
Then, we can also change the type suggestion for radius in spatial_neighbors_delaunay to float | tuple[float, float].

shashkat (commented):
It looks amazing to me now! Just a few more minor suggestions. I really liked the idea of using specialized Postprocessor classes!

src/squidpy/gr/_build.py
shashkat: Ahh you are right. None was indeed default for them at this commit, so this is the only place they were resolved. Yeah, I agree, even though the Builders would have defaults, it would be required to resolve None values before calling the builders.

src/squidpy/gr/neighbors.py
shashkat: Thanks a lot!

src/squidpy/gr/_build.py
shashkat: Yeah that makes sense!

src/squidpy/gr/neighbors.py
shashkat: I am wondering what could be a possible use case of having delaunay=True in GridBuilder (and spatial_neighbors_grid). Is it something like when the grid has some sparsity due to some spots being absent or removed due to low quality? Do we think it would be useful to add an example of possible use cases for this argument in documentation of spatial_neighbors_grid?

src/squidpy/gr/neighbors.py
```diff
@@ -0,0 +1,494 @@
...
+    Two observations are connected when their Euclidean distance falls within
+    the specified radius. See :func:`~squidpy.gr.spatial_neighbors_radius` for the
+    user-facing API or :func:`~squidpy.gr.spatial_neighbors_from_builder` for
+    direct builder usage.
+    """
+
+    def __init__(
+        self,
+        radius: float | tuple[float, float],
+        transform: str | Transform | None = None,
```
grst, line 212: Adding these arguments here adds again some complexity/branching to the RadiusBuilder class, which makes it harder to reason about it and implement it on other backends. 

An alternative would be to support these arguments only in the high-level function and instead support chaining of Builders and Postprocessors. Preprocessors would inherit the API from Builders and could be instantiated with a Builder or another already initialized preprocessor. E.g.  

```python
transform_interval_builder: Builder = TransformPreprocessor(
    DistanceIntervalPostprocessor(RadiusBuilder(...))
)
```

or

```python
transform_interval_builder: Builder = RadiusBuilder(...).chain(
    DistanceIntervalPostprocessor
).chain(
    TransformPreprocessor
)
```

grst: I don't have a super strong opinion on this, just as an additional idea.

src/squidpy/gr/neighbors.py
selmanozleyen: thanks very useful stuff. I will keep the old behaviour on the deprecated function though

src/squidpy/gr/_build.py
selmanozleyen: I will instead type it as Literal[4,6]

selmanozleyen:
These classes would be also ideally serializable so we dont need to rely on a forman under adata.uns which we are scared to change. I will make this as an issue to tackle when the time comes

src/squidpy/gr/neighbors.py
selmanozleyen: I see the concern but I wouldn't worry about the extensibility of RadiusBuilder since I don't see much use in inheriting it. Transform is like another parameter (e.g. radius) for now. We store it also to save the metadata. I know it's not a good design but I don't want to break to much of the old format/branching. Better designs can now be in the next new classes!

src/squidpy/gr/neighbors.py
selmanozleyen: It can but we want to save it for metadata. Otherwise same could be also said with radius and stuff

src/squidpy/gr/neighbors.py
selmanozleyen: it's mainly for future because we'd want to access the postprocessors in general

src/squidpy/gr/_build.py
```diff
@@ -140,366 +353,477 @@ def spatial_neighbors(
...
+    neighs_key = Key.uns.spatial_neighs(key_added)
+    conns_key = Key.obsp.spatial_conn(key_added)
+    dists_key = Key.obsp.spatial_dist(key_added)
 
-def _transform_a_cosine(a: spmatrix) -> spmatrix:
-    return cosine_similarity(a, dense_output=False)
+    neighbors_dict = {
+        "connectivities_key": conns_key,
+        "distances_key": dists_key,
+        "params": {
```
timtreis: old method also stored "coord_type", not including it here will break things that rely on it downstream

src/squidpy/gr/_build.py
```diff
@@ -140,366 +353,477 @@ def spatial_neighbors(
...
+            "n_neighbors": getattr(builder, "n_neighs", 6),
```
timtreis: Why is 6 hardcoded? Not all method have 6 neighbors

src/squidpy/gr/_build.py
```diff
@@ -140,366 +353,477 @@ def spatial_neighbors(
...
-    neigh_correct: bool = False,
+@d.dedent
+def spatial_neighbors_radius(
+    data: AnnData | SpatialData,
+    *,
+    spatial_key: str = Key.obsm.spatial,
+    elements_to_coordinate_systems: dict[str, str] | None = None,
+    table_key: str | None = None,
+    library_key: str | None = None,
+    radius: float | tuple[float, float] = 1.0,
```
timtreis: Why the default of 1 here? Isn't it None in the others?

src/squidpy/gr/_build.py
```diff
@@ -140,366 +353,477 @@ def spatial_neighbors(
...
-    Return D^{-1/2} * A * D^{-1/2}, where D = diag(degrees(A)) and A = adj.
+@d.dedent
+def spatial_neighbors_grid(
+    data: AnnData | SpatialData,
+    *,
+    spatial_key: str = Key.obsm.spatial,
+    elements_to_coordinate_systems: dict[str, str] | None = None,
+    table_key: str | None = None,
+    library_key: str | None = None,
+    n_neighs: Literal[4, 6] = 6,
```
timtreis: Why 4, 6 as a Literal?

timtreis (changes requested):
We're not really explicitly testing the new pattern, just indirectly through the equvialence tests. 

Also, we should align on the deprecate warning version

Claude:
```
Extensibility doc: SNN example will not import / run as written

  docs/extensibility.md SNNRadiusBuilder.build_graph doesn't initialise the matrices' set_diag consistently with the rest of the codebase — adj.setdiag(1.0 if self.set_diag else adj.diagonal()) will materialise an explicit zero diagonal because
  csr_matrix.diagonal() of a freshly constructed matrix returns a dense vector of zeros, then setdiag(0.0) is set on dst (fine). Minor, but worth either (a) testing this snippet in CI as a doctest, or (b) simplifying to mirror what
  RadiusBuilder.build_graph actually does. Also the example imports snnpy (optional 3p dep) and silently breaks doc-build if anyone runs the snippet — flag clearly that it is illustrative.
```

src/squidpy/gr/_build.py
selmanozleyen: From the docs what I understood is only those two makes sense

selmanozleyen: oky I fixed it

src/squidpy/gr/_build.py
selmanozleyen: fixed

src/squidpy/gr/_build.py
selmanozleyen: you are right, I made it a non-defaulted positional arg

grst: FWIW, it's also possible to have keyword-only arguments without default.

selmanozleyen: I agree, being explicit is nice

src/squidpy/gr/_build.py
selmanozleyen: I revised the uns structure. It should be fine now

selmanozleyen:
Hi @timtreis this is ready for review again

timtreis (approved):
lgtm

PR #1151, opened 2026-04-07
Add pluggable backend dispatch system

Intron7:
### Summary

- Adopt `scverse-backends>=0.0.3` for Squidpy backend dispatch.
- Align the trusted RAPIDS backend identity with Scanpy: canonical `rapids-singlecell`, with `cuda`, `rapids`, and `rapids_singlecell` aliases.
- Dispatch the four Squidpy graph functions implemented by RAPIDS SingleCell: `spatial_autocorr`, `co_occurrence`, `ligrec`, and `calculate_niche`.
- Preserve the historical joblib `backend` usage for `spatial_autocorr`, `co_occurrence`, and `ligrec` through explicit compatibility handling.
- Use `scverse_backends.testing.run_conformance` with Squidpy-owned CPU-vs-backend checks.

### Backend Contract

Backends register an adapter under the `squidpy.backends` entry-point group. The companion RAPIDS SingleCell change supplies that adapter; this host PR keeps the dispatcher configuration and aliases identical to the Scanpy integration.

### Verification

- `uv run --group test pytest -q tests/test_backends.py`
- `uv run --group test pytest -q tests/graph/test_ppatterns.py`
- `uv run --group test pytest -q tests/graph/test_ligrec.py -k 'parallel_backend or legacy_parallel_backend'`
- `uv run ruff check ...`


src/squidpy/_backends/_settings.py
```diff
@@ -0,0 +1,142 @@
...
+        canonical = resolve_backend_name(value)
+
+        # Completely unknown name
+        if canonical is None:
+            raise ValueError(f"Unknown backend {value!r}. Available backends: {available_backend_names() or ['cpu']}.")
+
+        # Trusted but not installed
+        if canonical in TRUSTED_BACKENDS and get_backend(canonical) is None:
+            package = TRUSTED_BACKENDS[canonical]["package"]
+            raise ValueError(
```
flying-sheep: Should be `ImportError`

src/squidpy/_backends/_registry.py
```diff
@@ -0,0 +1,113 @@
...
+_alias_map: dict[str, str] = {}  # alias -> canonical_name
+_discovered = False
+
+# Trusted (verified) backends and their known aliases.
+# Backends not in this list still work but emit a one-time warning on first use.
+# To become trusted, submit a PR adding your backend here and pass the
+# conformance test suite (squidpy.testing.backend_conformance).
+TRUSTED_BACKENDS: dict[str, dict[str, Any]] = {
+    "rapids_singlecell": {
+        "aliases": ["rapids-singlecell", "rsc", "cuda", "gpu"],
```
flying-sheep, line 23: why have multiple? maybe just have a error message helper that recognizes them and does “backend 'gpu' does not exist, did you mean 'rapids-singlecell'?”

src/squidpy/_backends/_dispatch.py
flying-sheep, line 1: didn’t @selmanozleyen already build something like this? this looks like it’s similar code, so if the other version is merged, this should be unified with that.

src/squidpy/_backends/_dispatch.py
```diff
@@ -0,0 +1,432 @@
...
+    obj = candidate
+    while obj is not None:
+        if obj is wrapper:
+            return candidate
+        obj = getattr(obj, "__wrapped__", None)
+
+    return wrapper
+
+
+def _update_signatures() -> None:
```
selmanozleyen: This private function is not used in this file but it is exported. Therefore it shouldn't be private. The module _dispatch is already private so it is fine to remove the `_` here

src/squidpy/_backends/_dispatch.py
```diff
@@ -0,0 +1,432 @@
...
+
+
+# All functions decorated with @dispatch, so we can update their signatures later
+_dispatched_functions: list[Callable] = []
+
+
+def _get_param_sets(
+    func: Callable,
+    adapter_method: Callable,
+    func_name: str,
```
selmanozleyen, line 27: this is unused

selmanozleyen:
compared to https://github.com/scverse/squidpy/pull/1093 I agree that renaming it to `backend` rather than `device` is way better.

However, currently there is a blocker because the `backend` argument is reserved for the parallelize backend. Which is also one more motivation to get rid of it. To fix it quickly we can basically rename it to `parallelize_backend` and give warnings until we hit a good enough version and introduce `backend` again which doesn't sound ideal but it can be ok with sq 2.0

src/squidpy/_backends/_dispatch.py
```diff
@@ -0,0 +1,432 @@
...
+
+    return wrapper
+
+
+def _update_signatures() -> None:
+    """Merge GPU-only params from discovered backends into dispatched function signatures.
+
+    Called once automatically after backend discovery so that ``help()`` /
+    IDE tooltips show the full parameter list (CPU + GPU + backend) with
+    documentation.
```
selmanozleyen, line 275: I am not sure about dynamically updating the signature and docstrings. I like the approach with just linking to the dispatched backend better in :https://github.com/scverse/squidpy/pull/1093

@flying-sheep @ilan-gold wdyt? For example if I compiled the docs with gpu will it compile a mix of rsc and squidpy docs?

src/squidpy/_backends/_dispatch.py
Intron7: @selmanozleyen approach is similar but less extensible. It's still a draft PR like this one. This approach would immediately open the door for other backends. With 0 updates needed in squidpy.

src/squidpy/_backends/_dispatch.py
selmanozleyen: I would ideally want to expose everything, even if dispatched to another function in the public function signature

src/squidpy/_backends/_registry.py
```diff
@@ -0,0 +1,113 @@
...
+            logger.debug("Failed to load backend entrypoint %r", ep.name, exc_info=True)
+
+    # Merge backend-specific params into dispatched function signatures
+    if _backends:
+        from squidpy._backends._dispatch import _update_signatures
+
+        _update_signatures()
+
+
+def _check_trusted(name: str) -> None:
```
selmanozleyen: again, should be `check_trusted`

Zethson:
> I agree that renaming it to backend rather than device is way better.

I think `device` has the advantage that the actual (CUDA) device could be specified -> like a specific GPU number. This would be weird to pass to `backend`, right?

But I think the right answer is to do what the rest of the pydata ecosystem does and just follow these patterns. I don't see any reason for us to stick out.

Generally, I hope that a generalized version of this could be implemented into https://github.com/scverse/scverse-misc so that it would eventually be reused across all of our packages trivially.

Edit: I just read on zulip:

> My plan is to longterm move this over into a dedicated scverse_backends package that every package can import

Yeah SGTM but maybe `scverse-misc` could also be the place.

selmanozleyen:
> I think device has the advantage that the actual (CUDA) device could be specified -> like a specific GPU number. This would be weird to pass to backend, right?

But this way we can leave it to the user to either set the context with whatever backend of their choice. If we had jax backend for example we could do it either in cpu or gpu.

src/squidpy/_backends/_dispatch.py
selmanozleyen: to be more clear func_name is not being used here

src/squidpy/_backends/_dispatch.py
```diff
@@ -0,0 +1,432 @@
...
+    """
+    func_name = func.__name__
+
+    @functools.wraps(func)
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        local_backend = kwargs.pop("backend", None)
+        effective = local_backend or settings.backend
+
+        if effective == "cpu":
+            return func(*args, **kwargs)
```
selmanozleyen, line 395: This has a subtle bug if we want to generalize. We only bind `*args` by order and don't ever check if the order and the names match. 

For example:

CPU: func(a, b)
GPU: func(b, a)
and both a and b are integers:

func(5, 10, backend="gpu")
becomes:

GPU sees b=5, a=10
If the backend computes something like a / b, you silently get 10 / 5 instead of 5 / 10. No exception, just incorrect output.

or imagine this

if you have 
CPU: threshold, n_perms
GPU: n_perms, threshold

these won't error at the level they should. Maybe it will say n_perms should be an integer inside the function but it shouldn't have been dispatchable to begin with.

selmanozleyen:
@Zethson We should also document the contract here. There are lots of assumptions hidden in this PR if we want to generalize.

For example currently the dispatch is position based for args, see the review I did, if it's not addressed and generalized it will introduce subtle bugs. 

To expose these assumptions ideally we should have a `dispatchable(f_base, f_backend_impl)` that checks if can be dispatchable (not only true false but also with a state: throws warning during runtime). This doesn't have to be written in code but I think it's easier to formulate it this way for me.

For two functions `f_base(*base_args, **base_kwargs), f_backend_impl(*backend_args, **backend_kwargs)` to be dispatchable:

- can have `base_args` that aren't in `backend_args` which we can call `base_only_args`.  These are dropped:
     - silently if their defaults are equal
     - with warning if their defaults don't match
- same for `base_only_kwargs` we have two options.
- currently we can similarly have `backend_only_args` **but we should think about this more**, current behaviour is we can't document these in the base implementation therefore we silently update the signature of the base function for them to be included. **This is a big redflag for me**. I will write some of my suggestions to fix this below
- the intersection by arg name of `base_args` and `backend_args` should be in same order i.e., (this isn't currently done but I assume it will be fixed)
```python
shared = [name for name in base_args if name in backend_arg_names]
backend_shared = [name for name in backend_args if name in base_arg_names]
# we want
shared == backend_shared
``` 

These are the current terms for the contract. But instead of update-the-signature trick we can disallow `backend_only_args` to exist and only support `backend_only_kwargs`. Then we can have in the base signature `backend_kwargs` in`f_base(...,backend_kwargs)`. We can update then documentation of backend_kwargs dynamically that might explain what these kwargs do but we won't be changing our function signatures this way.

PR #982, opened 2025-03-28
[EXPERIMENTAL]: Integrate cp-measure

timtreis:
Associated notebook: https://github.com/scverse/squidpy_notebooks/blob/add_cpmeasure_notebook/tutorials/tutorial_cpmeasure.ipynb

PR #1157, opened 2026-04-10
Add tiling QC metric for tile-boundary segmentation artifacts

timtreis:
## Summary

- Adds `sq.experimental.tl.calculate_tiling_qc` — per-cell scoring that detects cells artificially cut by tile boundaries during segmentation, using collinearity-based straight-edge detection on contours
- Adds `sq.experimental.pl.tiling_qc` — diagnostic plot via spatialdata-plot where the tile grid emerges from high-scoring cells without requiring tile-border metadata
- Includes cell-aware tiling infrastructure (`_tiling.py`) for scalable labels-only tile extraction — will be shared with #982 when merged
- Scores stored in `.obs` of a QC AnnData table (`{labels_key}_qc`) with proper `spatialdata_attrs` linking and algorithm params in `.uns["tiling_qc"]`

## Metrics

| Metric | Description |
|--------|-------------|
| `max_straight_edge_ratio` | Longest collinear boundary segment / equivalent diameter |
| `cardinal_alignment_score` | Axis-alignment of that segment (1 = cardinal, 0 = diagonal) |


src/squidpy/experimental/im/_tiling.py
```diff
@@ -0,0 +1,427 @@
...
+    tile_to_cells: dict[tuple[int, int], set[int]] = {}
+    for lid, key in cell_to_tile.items():
+        tile_to_cells.setdefault(key, set()).add(lid)
+
+    n_rows = (H + tile_size - 1) // tile_size
+    n_cols = (W + tile_size - 1) // tile_size
+
+    specs: list[TileSpec] = []
+    for row in range(n_rows):
+        for col in range(n_cols):
```
selmanozleyen, line 264: Why not loop over the element you are going to operate on?
```
for (row, col), owned in tile_to_cells.items():
    by0 = row * tile_size
    bx0 = col * tile_size
    by1 = min(by0 + tile_size, H)
    bx1 = min(bx0 + tile_size, W)

    cy0 = max(by0 - margin, 0)
    cx0 = max(bx0 - margin, 0)
    cy1 = min(by1 + margin, H)
    cx1 = min(bx1 + margin, W)

    specs.append(
        TileSpec(
            base=(by0, bx0, by1, bx1),
            crop=(cy0, cx0, cy1, cx1),
            owned_ids=frozenset(owned),
        )
    )
```

src/squidpy/experimental/tl/_tiling_qc.py
```diff
@@ -0,0 +1,657 @@
...
+        raise ValueError("No cells scored - labels may be empty or all below min_area.")
+
+    combined = pd.concat(tile_dfs, axis=0).sort_index()
+
+    if combined.index.duplicated().any():
+        dups = combined.index[combined.index.duplicated()].unique().tolist()
+        raise RuntimeError(f"Duplicate cell IDs across tiles - tile ownership may be broken. Duplicates: {dups}")
+
+    # --- Validation ---
+    if not outlier_use_cut and not outlier_use_smoothed:
```
selmanozleyen, line 562: can we move the validations above before doing computations?

src/squidpy/experimental/tl/_tiling_qc.py
```diff
@@ -0,0 +1,657 @@
...
+        raise ValueError("At least one outlier gate must be enabled (outlier_use_cut or outlier_use_smoothed).")
+    if outlier_use_cut and nmads_cut <= 0:
+        raise ValueError(f"nmads_cut must be positive, got {nmads_cut}.")
+    if outlier_use_smoothed and nmads_smoothed <= 0:
+        raise ValueError(f"nmads_smoothed must be positive, got {nmads_smoothed}.")
+
+    # --- Spatial context post-processing ---
+    n_cells = len(combined)
+    k = 10
```
selmanozleyen, line 571: why magic numbers here?

src/squidpy/experimental/tl/_tiling_qc.py
```diff
@@ -0,0 +1,657 @@
...
+        # Build is_outlier from enabled gates (AND when both active)
+        is_outlier = np.ones(n_cells, dtype=bool)
+
+        if outlier_use_cut:
+            median_c = np.median(cut_scores)
+            mad_c = np.median(np.abs(cut_scores - median_c))
+            if mad_c < 1e-12:
+                is_outlier[:] = False
+            else:
+                is_outlier &= cut_scores >= median_c + nmads_cut * mad_c * 1.4826
```
selmanozleyen, line 602: what's 1.4826? why magic numbers in general?

src/squidpy/experimental/tl/_tiling_qc.py
timtreis: `k = 10` because we're calculating the local reference context for a given cell here. In a perfect grid a cell would have 8 neighbours, but biology is fuzzy, So I gave it a little wiggle room. Less and we might not capture all cut cells around it, more and we'd be starting to waste compute

selmanozleyen: Can we have them at the top of the function then and maybe not call it `k` maybe something that suggests its a magic number that can be changed from there?

src/squidpy/experimental/tl/_tiling_qc.py
timtreis: That "magic number" is basically the conversion factor from MAD to standard deviation. You can use it like `sd  ~ 1.4826 x MAD`, cheap and fast approxipation of the real thing

src/squidpy/experimental/tl/_tiling_qc.py
selmanozleyen: and also document the reasoning for hardcoding them, it is fine to hard-code at this stage when done intentionally I guess

timtreis: Yeah, could rename 🤔

selmanozleyen: I mean not only for this but as hardcoded values being on top of the function to make it more explicit.

timtreis: I wouldn't move it far away from this function because it's really only used here and nowehre else

timtreis: upgraded to user-exposed parameter and therefore now documented in the function

src/squidpy/experimental/tl/_tiling_qc.py
timtreis: renamed to constant

LLehner (approved):
Just a general question: how is it currently decided whether a function should go into ```experimental```?

timtreis:
> Just a general question: how is it currently decided whether a function should go into `experimental`?

currently pretty much "anything that actually uses sdata"

src/squidpy/_utils.py
```diff
@@ -45,6 +46,19 @@ def wrapper(*args: Any, **kw: Any) -> Any:
 NDArrayA = NDArray[Any]
 
 
+def cpu_count() -> int:
```
selmanozleyen: isn't this change unrelated to this PR? Do we really need it? Like have you seen os.cpu_count cause errors so far?

src/squidpy/experimental/pl/_tiling_qc.py
```diff
@@ -0,0 +1,79 @@
...
+import spatialdata as sd
+
+__all__ = ["tiling_qc"]
+
+
+def tiling_qc(
+    sdata: sd.SpatialData,
+    labels_key: str,
+    qc_key: str | None = None,
+    score_col: str = "nhood_outlier_fraction",
```
selmanozleyen:
```suggestion
    score_col: Literal["nhood_outlier_fraction", "smoothed_cut_score", "cut_score", "max_straight_edge_ratio", "cardinal_alignment_score"] = "nhood_outlier_fraction",
```

src/squidpy/experimental/tl/_tiling_qc.py
```diff
@@ -0,0 +1,668 @@
...
+    distance_tol: float = _DEFAULT_DISTANCE_TOL,
+    min_area: int = _MIN_CELL_AREA,
+    downsample: int = 1,
+    outlier_use_cut: bool = True,
+    outlier_use_smoothed: bool = True,
+    nmads_cut: float = 1.5,
+    nmads_smoothed: float = 3,
+    n_neighbors: int = 10,
+    n_jobs: int = -1,
+    client: Client | None = None,
```
selmanozleyen: I don't think getting client as an arg is the right approach here. In dask we usually would expect setting the client with a context manager I think. Something like this to get the client object if it exists:

```
# removing client from the function signature and not importing dask distributed at top level.
client = None
try:
    import dask.distributed as dd
    client = dd.get_client() # value error if there is no client active
except (ImportError, ValueError):
    client = None
```
but I am not sure of this suggestion. so @flying-sheep what do you think?

src/squidpy/experimental/tl/_tiling_qc.py
```diff
@@ -0,0 +1,668 @@
...
+    n_jobs
+        Number of threads for tile processing.  ``-1`` (default) uses
+        all available CPUs.  Ignored when ``client`` is provided.
+    client
+        A :class:`dask.distributed.Client` for distributed execution.
+        When provided, tile processing is submitted to this client,
+        ``n_jobs`` is ignored, and progress is reported via the dask
+        dashboard.  Workers must have access to the underlying data
+        store (e.g. shared filesystem or cloud storage for zarr).
+    adata_key_added
```
selmanozleyen: Not a strong suggestion but an opinion. I think key_added would be nicer. Or table_key_added. Because now I am not sure if this adds to the adata or adds the adata in this name

src/squidpy/_utils.py
selmanozleyen: update: I now remember _get_n_cores. Why don't we modify that one instead?

src/squidpy/experimental/im/_tiling.py
```diff
@@ -0,0 +1,419 @@
...
+    Reads only the smallest resolution, then scales coordinates to *target_scale*.
+    """
+    available = list(labels_node.keys())
+    if not available:
+        return {}
+
+    def _scale_idx(k: str) -> int:
+        num = "".join(c for c in k if c.isdigit())
+        return int(num) if num else 0
+
```
selmanozleyen: This is a bit weird to me. 
We should either:
- Match only the expected format, e.g. ^scale(\d+)$, and fail clearly for unexpected keys.
- or choose the coarsest level by actual spatial shape, like the nearby _choose_label_scale_for_image() logic does, since “coarsest” is a property of the array dimensions, not the key string.

selmanozleyen (approved):
Besides the  things I mentioned it looks good to me. Really depends on the feedback this will get, otherwise I feel like it would be early optimization to dig deeper.

src/squidpy/_utils.py
timtreis: folded into that one 👍

src/squidpy/experimental/tl/_tiling_qc.py
timtreis: True, pivot to letting dask figure this out in the background

src/squidpy/experimental/tl/_tiling_qc.py
timtreis: Renamed to table_key_added

PR #1162, opened 2026-04-16
General purpose `sq.tl.align()` function with `STAlign` core logic

timtreis:
Implements the following functions:

```
sq.experimental.tl.align_by_landmarks(
    sdata,
    cs_name_ref="section_a",
    cs_name_query="section_b",
    landmarks_ref=((x1, y1), (x2, y2), (x3, y3)),
    landmarks_query=((x1, y1), (x2, y2), (x3, y3)),
    model="similarity",  # or "affine" for 6-DOF
)
```
This registers a spatialdata Affine transformation on every element in cs_b, mapping it into cs_a. After the call, sdata is updated in-place, all elements in section_b now know how to transform into section_a's coordinate system.

```  
result = sq.experimental.tl.align_obs(
    adata_ref,
    adata_query,
    flavour="stalign", # will later accept moscot
    inplace=False, # otherwise, creates new aligned adata in sdata
    output_mode="obs",
)
```
output_mode controls what you get back:
- "obs": a new AnnData with baked-in aligned coordinates
- "affine": registers a transform on the SpatialData element (only works if the backend produces an affine; LDDMM doesn't)
- "return": raw AlignResult with the displacement field and full solver state for power users

```
sq.experimental.tl.align_images(
    ...
)
```
Currently dead but will be implemente via STalign. Already in here to better align function internals

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -104,6 +104,8 @@ jobs:
             env: { name: hatch-test.py3.12-stable }
           - os: macos-latest
             env: { name: hatch-test.py3.13-pre } # pre-release only needs one OS
+          - os: macos-latest
+            env: { name: hatch-test.py3.13-stable-jax } # JAX only needs one OS
           - os: ubuntu-latest
             env: { name: hatch-test.py3.13-stable } # skipping because we run this as a coverage job
     name: ${{ matrix.env.label }} (${{ matrix.os }})
--- a/hatch.toml
+++ b/hatch.toml
@@ -40,12 +40,20 @@ deps = ["stable"]
 python = ["3.11", "3.12", "3.13"]
 
 [[envs.hatch-test.matrix]]
+deps = ["stable"]
+python = ["3.13"]
+extras = ["jax"]
+
+[[envs.hatch-test.matrix]]
 deps = ["pre"]
 python = ["3.13"]
 
 [envs.hatch-test.overrides]
 matrix.deps.env-vars = [
   { key = "UV_PRERELEASE", value = "allow", if = ["pre"] },
+]
+matrix.extras.features = [
+  { value = "jax", if = ["jax"] },
 ]
 
 [envs.notebooks]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,6 +102,9 @@ optional-dependencies.docs = [
   "sphinxcontrib-bibtex>=2.3",
   "sphinxcontrib-spelling>=7.6.2",
 ]
+optional-dependencies.jax = [
+  "jax",
+]
 optional-dependencies.leiden = [
   "leidenalg",
   "spatialleiden>=0.4",
--- a/src/squidpy/experimental/__init__.py
+++ b/src/squidpy/experimental/__init__.py
@@ -6,6 +6,6 @@ These features may change or be removed in future releases.
 
 from __future__ import annotations
 
-from . import im, pl
+from . import im, pl, tl
 
-__all__ = ["im", "pl"]
+__all__ = ["im", "pl", "tl"]
--- a/src/squidpy/experimental/tl/__init__.py
+++ b/src/squidpy/experimental/tl/__init__.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Any
 
-from squidpy.experimental.tl._align import align_by_landmarks, align_images, align_obs
+from squidpy.experimental.tl._align import align_by_landmarks, align_obs
 
 if TYPE_CHECKING:
     from squidpy.experimental.tl._align._backends._stalign_tools import (
@@ -20,7 +20,6 @@ __all__ = [
     "STalignRegistrationConfig",
     "STalignResult",
     "align_by_landmarks",
-    "align_images",
     "align_obs",
 ]
 
--- a/src/squidpy/experimental/tl/__init__.py
+++ b/src/squidpy/experimental/tl/__init__.py
@@ -1,0 +1,56 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+from squidpy.experimental.tl._align import align_by_landmarks, align_images, align_obs
+
+if TYPE_CHECKING:
+    from squidpy.experimental.tl._align._backends._stalign_tools import (
+        STalignConfig,
+        STalignPreprocessConfig,
+        STalignPreprocessResult,
+        STalignRegistrationConfig,
+        STalignResult,
+    )
+
+__all__ = [
+    "STalignConfig",
+    "STalignPreprocessConfig",
+    "STalignPreprocessResult",
+    "STalignRegistrationConfig",
+    "STalignResult",
+    "align_by_landmarks",
+    "align_images",
+    "align_obs",
+]
+
+_STALIGN_REEXPORTS = frozenset(
+    {
+        "STalignConfig",
+        "STalignPreprocessConfig",
+        "STalignPreprocessResult",
+        "STalignRegistrationConfig",
+        "STalignResult",
+    }
+)
+
+
+def __getattr__(name: str) -> Any:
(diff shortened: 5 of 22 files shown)


selmanozleyen (changes requested):
- need to rename backend -> method everywhere
- it is not moscot since we don't need moscot, both stalign and gridot use jax but Ideally we'd want possible numba implementations?
- why is AlignPair only for `AnnData | DataTree`? Landmarks should be another possibility here
- again why in STAlign class there is align_obs and align_images but not align_landmarks? (it's fine to have partial support since GridOT class won't have maybe align_landmarks as well)
- why `model`? I'd name it ransac or whatever and expect the inputs based on that. My method should clarify what input I need. GridOT assumes the point clouds are in the same space for example and different algorithms can minimize different models of the input given.

I'd rather have

For **inputs**: `AlignObsPair, AlignImagesPair, AlignLandmarkPair` as possible inputs which are separated nicely.

For **algorithms**: `StAlignStrategy, GridOtStrategy, LandmarkAffineStrategy, LandmarkSimilarityStrategy, *NumbaStAlignStrategy, *OpticalFlowStrategy...` in which we can encapulate the heavy dependencies to classes.

We are also not nicely conceptualizing the relation of alignment to input to what the user actually gave. When the input is spatialdata the alignment pairs can be all, when it's anndata it's either the obs or uns for landmarks, or when it's numpy arrays its landmarks right? Why not go on that pattern?

If I have spatialdata I can have a landmark based algorithm run and apply that to my images and my anndata depending on what I will give.

Cc @FrancescaDr @sarajimenez

What do you guys also think? What makes sense for you usually?

selmanozleyen:
Would this kind of refactor make sense:

## Public surface: 2 functions, 3 output modes

The split is driven by *input source*, not by *element kind*:

| input source                        | function                |
| ----------------------------------- | ----------------------- |
| `AnnData` / `SpatialData` container | `align(...)`            |
| in-memory landmark arrays / tuples  | `align_by_landmarks(...)` |

The `obs` vs `image` distinction is one kwarg of `align()`, not a
separate function -- both consume containers and produce a transform.
Output is a single enum across both functions:

| `output_mode`  | returns                                         |
| -------------- | ----------------------------------------------- |
| `"object"`     | raw `AlignResult` (no container is touched)     |
| `"copy"`       | modified copy of the writeback target           |
| `"inplace"`    | mutates the target, returns `None`              |

```python
def align(
    data_ref: AnnData | SpatialData,
    data_query: AnnData | SpatialData,
    *,
    method: str = "stalign",
    on: Literal["obs", "image"] = "obs",
    ref_key: str | None = None,           # SpatialData only: which table / image
    query_key: str | None = None,
    output_mode: Literal["object", "copy", "inplace"] = "inplace",
    key_added: str | None = None,
    **method_kwargs,
) -> AlignResult | AnnData | SpatialData | None: ...

def align_by_landmarks(
    landmarks_ref: np.ndarray | Sequence[tuple[float, float]],
    landmarks_query: np.ndarray | Sequence[tuple[float, float]],
    *,
    method: Literal["similarity", "affine"] = "similarity",
    data: AnnData | SpatialData | None = None,   # required for copy / inplace
    cs_name_ref: str | None = None,
    cs_name_query: str | None = None,
    output_mode: Literal["object", "copy", "inplace"] = "object",
    key_added: str | None = None,
    method_kwargs
) -> AlignResult | AnnData | SpatialData | None: ...
```

`align_by_landmarks` defaults to `output_mode="object"` because the only
input is arrays in memory -- there is no implicit container target.
Choosing `"copy"` or `"inplace"` requires `data=`.

Naming notes:

- `ref_key` / `query_key` (not `element_ref` / `element_query`): "element"
  is `SpatialData` jargon; `_key` is the standard scverse suffix
  (`cluster_key`, `library_key`, `spatial_key`, `connectivity_key`).
- `data=` (not `target=`) on `align_by_landmarks`: from the user's
  perspective they are passing *the data on which to apply the
  alignment*, not a "writeback dump". The `data` slot also matches
  `align`'s `data_ref` / `data_query` vocabulary.

## Strategy hierarchy: two families, pure `fit`

`Strategy.fit*` is a *pure* function -- raw fit args in, `AlignResult`
out, no container in scope.  Containers are routed separately as a
writeback `AlignTarget` and only enter the picture after the fit.  The
landmark family in particular never sees an `AnnData` or `SpatialData`
during the fit.

```
AlignStrategy (ABC)
    name: ClassVar[str]
    requires: ClassVar[tuple[str, ...]]
├── AlignDataStrategy (ABC)
│       on: ClassVar[frozenset[str]]   # subset of {"obs", "image"}
│       def fit_obs(self, ref: AnnData,        query: AnnData)        -> AlignResult: ...
│       def fit_image(self, ref: xr.DataArray, query: xr.DataArray)   -> AlignResult: ...
│   ├── StAlignStrategy             # on={"obs","image"}; requires=("jax",)
│   ├── NumbaStalignStrategy        # on={"obs","image"}; requires=("numba",)
│   └── GridOtStrategy              # on={"obs"};         requires=("jax",)
└── AlignLandmarkStrategy (ABC)
        def fit(self, landmarks_ref: np.ndarray, landmarks_query: np.ndarray) -> AlignResult: ...
    ├── LandmarkSimilarityStrategy
    └── LandmarkAffineStrategy
```


### Writeback target (separate concept, kept)

```python
@dataclass(frozen=True)
class AlignTarget:
    container: AnnData | SpatialData
    key: str | None = None           # element key (SpatialData only; ignored for AnnData)
```

Two fields, both intent-neutral.  Coordinate-system names live on the
`AlignResult` (the affine carries `source_cs` / `target_cs`), not on the
writeback target -- they describe *the transform*, not *where to put
it*.

`AlignTarget` survives across the resolver -> strategy -> writeback
boundary (the fit args do not), so it earns being a typed object.
Resolvers in `_io.py` produce `(ref_data, query_data, target | None)`:

| public function             | fit args                            | target source                              |
| --------------------------- | ----------------------------------- | ------------------------------------------ |
| `align(...)`                | extracted from `data_query`         | `data_query` itself -> `AlignTarget`       |
| `align_by_landmarks(...)`   | `landmarks_ref`, `landmarks_query`  | `data=` kwarg -> `AlignTarget` (or `None`) |

Two flat `(method) -> Strategy` registries (one per family), no
`(method, mode)` tuple keys -- the family already pins it.

## Result hierarchy: polymorphic `apply_to(target)`

```
AlignResult (ABC)
    metadata: dict
    def apply_to(self, target: AlignTarget, *, key_added=None) -> AnnData | SpatialData
├── AffineResult         # 3x3; SpatialData -> set_transformation, AnnData -> obsm
├── DisplacementResult   # (N, 2); per-obs deltas; obsm-only on either container
└── DiffeomorphismResult # velocity field + integrator (future, lightweight)
```

*Invariant: results and inputs carry pointers / parametric forms, not
bulk data.* `DiffeomorphismResult` keeps the velocity field; the dense
`(H, W, 2)` warp is materialised only when actually needed.

Each result subclass dispatches on `target.container` (and on
`target.key`) inside `apply_to`, absorbing the three writeback paths
currently in `_io.py::apply_affine_to_cs`.  Coordinate-system info comes
from the result's own transform (`source_cs`, `target_cs`), not from the
target.  `_writeback` collapses to:

```python
def _writeback(target: AlignTarget | None, result: AlignResult, *, output_mode, key_added):
    if output_mode == "object":
        return result
    if target is None:
        raise ValueError(f"output_mode={output_mode!r} requires a target container.")
    if output_mode == "copy":
        target = _copy_for_writeback(target)
    return result.apply_to(target, key_added=key_added)
```

No `isinstance` ladder on the result type, no
`output_mode in {"obs", "affine"}` branch -- the result class owns the
"can I be applied to this target?" question and raises a single, typed
`NotImplementedError` when a combination doesn't make sense (e.g.
`DisplacementResult.apply_to(<SpatialData target without a `key`>)`).

selmanozleyen:
closed in favor of: https://github.com/scverse/squidpy/pull/1195

PR #1163, opened 2026-04-16
Draft: add contour-based ring density and smooth signed-distance density for SpatialData

### LINKED ISSUE #1160: Feature request: contour-based inward/outward ring density for cells and transcripts around polygon annotations
## Summary

I would like to propose contour-based inward/outward ring density around polygon
annotations in `SpatialData`.

This is related to #837, but it narrows the request to a concrete downstream
workflow: instead of only exposing distances to polygon annotations, the goal is
to compute signed distance and ring-based density profiles for both cells and
transcripts around a contour.

## Use cases

- Tumor boundary analyses where we want to quantify cell or transcript density
  as we move inward or outward from an annotated interface.
- Lumen-edge analyses where density should be summarized in concentric inward and
  outward rings.
- Tissue compartment or manual contour annotations where the user wants a radial
  density profile rather than only a nearest-distance column.

## Proposed inputs

- Polygon or MultiPolygon annotations from `SpatialData` used as the contour
  reference.
- Cells counted by centroid assignment.
- Transcripts counted by point coordinates.
- User-defined ring width and inward/outward extent in the dataset coordinate
  system.

## Proposed outputs

A per-ring table, optionally stratified by library or sample, with at least:

- `ring_start`
- `ring_end`
- `count`
- `area`
- `density = count / area`

Signed-distance semantics:

- boundary = `0`
- inside = negative
- outside = positive

## Why this is not a duplicate of #837

#837 already points toward polygon-aware distance calculations. This request is a
more concrete downstream proposal built on that direction:

1. add a contour-aware signed-distance primitive
2. add ring-density aggregation on top of that primitive

So I see this as a focused extension of #837 rather than a separate geometry
system.

## Minimal API sketch

Low-level primitive:

```python
sq.tl.var_by_distance(
    sdata,
    groups=...,
    cluster_key=...,
    annotations="tumor_boundary",
    coordinate_system="global",
)
```

High-level helper:

```python
sq.tl.ring_density(
    sdata,
    contour_key="tumor_boundary",
    target="cells",  # or "transcripts"
    ring_width=25.0,
    inward=100.0,
    outward=100.0,
)
```

## Proposed v1 scope

In scope:

- `SpatialData`-first API
- Polygon and MultiPolygon contour support
- signed distance to the contour boundary
- centroid-based cell density
- point-based transcript density
- per-ring `count`, `area`, and `density`

Out of scope for v1:

- weighting cells by polygon overlap with rings
- 3D contours
- smoothing or KDE-based density
- automatic conflict resolution across multiple overlapping contours

## Related context

- Existing issue: #837
- Nearby but different PR: #807
- Related SpatialData discussion: [Some suggestions and proposals for annotations in SpatialData #975](https://github.com/scverse/spatialdata/issues/975) - broader discussion of annotation/linking internals that may become relevant for contour-based workflows and table-to-element linking semantics.

If this direction sounds reasonable, I would be happy to follow up with a minimal
Draft PR that keeps the first implementation intentionally small.

hutaobo:
## Description

This draft PR proposes contour-aware density quantification around polygon annotations in `SpatialData`.

It adds two related tools:

1. `sq.tl.ring_density(...)`
   Computes inward/outward ring-based density profiles around polygon contours for cells or transcripts, using signed-distance semantics relative to the contour boundary.

2. `sq.tl.smooth_density_by_distance(...)`
   Computes a continuous signed-distance density profile for transcripts using Gaussian kernel smoothing on signed distance, while keeping an area-normalized density interpretation.

The main motivation is to support workflows such as tumor boundary, lumen edge, or compartment-interface analyses where users want density profiles around annotated contours rather than only nearest-distance values.

This draft intentionally includes both a discrete ring-based summary and a continuous smooth alternative so the API direction can be reviewed together. I realize that issue #1160 scoped smoothing as out of scope for a minimal v1, so I would especially welcome feedback on whether the smooth method should be split into a follow-up PR or deferred.

## How has this been tested?

Tested locally with:
- focused synthetic tests for `ring_density`
- focused synthetic tests for `smooth_density_by_distance`
- combined local test run:
  - `pytest -q tests/tools/test_ring_density.py tests/tools/test_smooth_density_by_distance.py`

The smooth method tests cover:
- output structure and metadata
- `.uns` storage when `copy=False`
- boundary-enriched smooth behavior
- consistency across Dask partitions
- local shell-area geometry normalization
- reflection-based boundary correction

I also ran a real-data smoke test on a local `SpatialData` object with:
- `protein_cluster_contours`
- `assigned_structure == "Structure 4"`
- transcript feature `VIM`

This produced a continuous signed-distance output table with 23 contours and 33 distance grid points.

## Closes

Closes #1160

### DIFF
--- a/src/squidpy/tl/__init__.py
+++ b/src/squidpy/tl/__init__.py
@@ -2,5 +2,7 @@
 
 from __future__ import annotations
 
+from squidpy.tl._ring_density import ring_density
 from squidpy.tl._sliding_window import _calculate_window_corners, sliding_window
+from squidpy.tl._smooth_density_by_distance import smooth_density_by_distance
 from squidpy.tl._var_by_distance import var_by_distance
--- a/src/squidpy/tl/_ring_density.py
+++ b/src/squidpy/tl/_ring_density.py
@@ -101,9 +101,7 @@ def ring_density(
     assert_one_of(target, ["cells", "transcripts"], name="target")
 
     feature_values = (
-        assert_non_empty_sequence(feature_values, name="feature_values")
-        if feature_values is not None
-        else None
+        assert_non_empty_sequence(feature_values, name="feature_values") if feature_values is not None else None
     )
 
     contours = _prepare_contours(
--- a/src/squidpy/tl/_ring_density.py
+++ b/src/squidpy/tl/_ring_density.py
@@ -1,0 +1,484 @@
+from __future__ import annotations
+
+from collections.abc import Iterable, Sequence
+from typing import Any, Literal
+
+import geopandas as gpd
+import numpy as np
+import pandas as pd
+from scanpy import logging as logg
+from shapely import STRtree, affinity, boundary, contains, distance, points
+from spatialdata import SpatialData
+from spatialdata.transformations import get_transformation
+
+from squidpy._docs import d
+from squidpy._validators import (
+    assert_key_in_sdata,
+    assert_non_empty_sequence,
+    assert_non_negative,
+    assert_one_of,
+    assert_positive,
+)
+from squidpy.gr._utils import _save_data
+
+__all__ = ["ring_density"]
+
+
+@d.dedent
+def ring_density(
+    sdata: SpatialData,
+    contour_key: str,
+    target: Literal["cells", "transcripts"] = "cells",
+    table_key: str | None = None,
+    points_key: str = "transcripts",
+    shape_key: str | None = None,
+    ring_width: float = 50.0,
+    inward: float = 0.0,
+    outward: float = 50.0,
+    coordinate_system: str = "global",
+    contour_query: str | None = None,
+    feature_key: str | None = None,
+    feature_values: str | Sequence[str] | None = None,
+    metadata_keys: str | Sequence[str] | None = None,
+    key_added: str = "ring_density",
+    copy: bool = False,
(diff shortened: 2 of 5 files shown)


hutaobo:
Thanks for taking a look. I would especially appreciate feedback on three design questions:\n\n1. Does  belong in  as proposed?\n2. Should  stay in this same PR for joint API review, or would you prefer it split or deferred?\n3. Does it make sense to eventually factor the signed-distance computation into a lower-level helper shared by both tools?\n\nI also generated local example figures for the ring-based and smooth profiles and can add them in a follow-up comment if that would be helpful.

hutaobo:
Thanks for taking a look. I would especially appreciate feedback on three design questions:

1. Does `sq.tl.ring_density(...)` belong in `tl` as proposed?
2. Should `sq.tl.smooth_density_by_distance(...)` stay in this same PR for joint API review, or would you prefer it split or deferred?
3. Does it make sense to eventually factor the signed-distance computation into a lower-level helper shared by both tools?

I also generated local example figures for the ring-based and smooth profiles and can add them in a follow-up comment if that would be helpful.

timtreis:
Hey @hutaobo, could you demonstrate on real data which biological insights one is getting from that?

PR #1170, opened 2026-05-08
Add tile-cut stitching follow-up to calculate_tiling_qc

timtreis:
## Summary

Builds on #1157.  Adds a follow-up pass that recovers cells which segmentation tiling broke into 2-4 pieces by detecting facing cut edges across tile boundaries and assigning each candidate pair a transparent geometric score.  Worst case (4-tile corner) is handled.

Two public functions:

- **`squidpy.experimental.tl.stitch_tile_cuts(sdata, labels_key, ...)`** -- reads `is_outlier=True` cells from the QC table, extracts cut-edge candidates via bbox-edge alignment, scores each pair with a transparent geometric composite (see below), and assembles confident pairs into 2-4-piece groups via union-find with corner-junction validation.  Writes 4 `.obs` columns to the existing QC table -- `stitch_group_id`, `is_stitched`, `n_pieces`, `stitch_confidence` -- plus a `.uns["tiling_stitch"]` audit trail (params, score formula, run summary).  **The labels element is never mutated.**
- **`squidpy.experimental.im.make_stitched_labels(sdata, labels_key, ..., merge_strategy="sum", inplace=True)`** -- opt-in materialisation of a stitched labels element via a lazy dask LUT, plus a collapsed AnnData with one row per unique `stitch_group_id` (unstitched cells pass through unchanged, stitched groups collapse).  Numeric `.obs` columns and `.X` aggregate via `merge_strategy` (sum/min/max/mean/median/first or callable); group-invariant + non-numeric columns always take "first".  Preserves `.uns`, `.var`, and any user-added obs columns.

`calculate_tiling_qc` re-runs now warn and drop stale stitch columns when the QC table is overwritten.

## How `stitch_confidence` is computed

For each candidate pair the four geometric / shape-quality features below are averaged into a single score in `[0, 1]`.  **No coefficients are fitted or shipped** -- the formula is the entire model and is recorded in `.uns["tiling_stitch"]["score_formula"]`.

| feature | what it captures | range |
|---|---|---|
| `iou` | 1-D intersection-over-union of the two cut-edge extents along the boundary | [0, 1] |
| `endpoint_match` | how closely the chord endpoints coincide -- true cuts share endpoints, unrelated cells don't | [0, 1] |
| `merge_compactness` | `4*pi*A / P^2` of the union mask after morphologically closing the seam gap.  Real cells are reasonably compact; false merges produce weird perimeters | [0, 1] |
| `merge_solidity` | union mask area / convex hull area.  Real cells are convex-ish; false merges have concave joins | [0, 1] |

```
stitch_confidence = (iou + endpoint_match + merge_compactness + merge_solidity) / 4
```

A `gap_score` is also computed but only used as a hard filter (already inside `max_gap` by construction); it does not enter the score.  The two `merge_*` features are computed by materialising a tight crop around the union of the candidate pieces, closing the gap with a `disk(3)` structuring element, and running `regionprops` on the largest connected component.

`min_confidence` is a threshold on this mean; `0.7` is the default starting point.  **Tune for your data** -- the score is heuristic, dataset-independent, and not a calibrated probability.  Review false positives / negatives via `make_stitched_labels` and the visual test fixture, then adjust.


selmanozleyen (changes requested):
I understand you want to keep PR count small but can we separate this into 3 PRs? otherwise there is no way for me to really review it.

1- Refactor (docs and tests also)

    _tiling_qc.py
    tests/experimental/test_tiling_qc.py
    (optionally) docs entry for TilingQCParams
    Add shared resolve_labels_array helper

2 - “Add stitch_tile_cuts + StitchParams (stitching algorithm)”

    _tiling_stitch.py
    tl/init.py export changes
    tests/experimental/test_tiling_stitch.py (+ baseline image)
    Update docs/api.md for stitch_tile_cuts / StitchParams
    _tiling_qc.py warning function _warn_if_dropping_stitch_columns(...)

3- “Add make_stitched_labels (materialize stitched labels + collapsed table)”

    _stitched_labels.py
    im/init.py export changes
    tests/experimental/test_stitched_labels.py
    Update docs/api.md for make_stitched_labels

About stitch_confidence I'd also want to understand the math and intuition behind. Isn't there a way to maybe make AI write a html file to visualize/simulate the score of this function? Or some notebook to understand a bit.
Because from what I understand it's a novel formula to compute it. Therefore I'd also like to review the math/science of it.

timtreis:
hm yeah, makes sense 🤔 Will split it into smaller ones but leave this one open as a reference until the others are merged 👌

timtreis:
Closed in favour of #1188, #1193, and #1207

PR #1172, opened 2026-05-11
Add derive_mpp_from_shapes utility

timtreis:
## Summary

Adds `squidpy.experimental.utils.derive_mpp_from_shapes`, a utility that computes microns-per-pixel of a target coordinate system from a shapes element plus one known physical measurement (spot pitch or spot diameter). Works for Visium, Visium HD bins, and any regular-grid shapes layout without relying on `scalefactors.json`.

- **Pitch path** (recommended): transforms centroids into the target CS, measures the median k=1 NN distance, returns `um_between_centers / pitch_px`. Subsamples to 5000 centroids on large grids (relevant at Visium HD bin counts).
- **Diameter path**: for Point shapes uses `2 * median(radius) * sqrt(|det A|)`; for Polygon shapes uses `sqrt(median(area) * |det A|)`, avoiding a per-geometry shapely call at HD scale.
- Anisotropic transformations are hard-rejected with a clear error and per-axis would-be mpps, since a single scalar mpp is not well-defined under non-uniform scale or shear.
- 3D shapes and `MultiPolygon` geometries are rejected with informative errors.

### DIFF
--- a/src/squidpy/experimental/utils/_derive_mpp.py
+++ b/src/squidpy/experimental/utils/_derive_mpp.py
@@ -31,7 +31,11 @@ def derive_mpp_from_shapes(
     corresponding geometric quantity in the target coordinate system and dividing the physical value
     by it.
 
-    Exactly one of ``um_between_centers`` or ``um_diameter`` must be provided.
+    Exactly one of ``um_between_centers`` or ``um_diameter`` must be provided. Prefer
+    ``um_between_centers`` when the technology's canonical pitch is known: it averages over the
+    realised grid and is robust to per-spot calibration noise in the stored radius. ``um_diameter``
+    depends on a single stored scalar and can disagree by a fraction of a percent on real Visium
+    data where the radius and grid pitch are calibrated separately.
 
     The function requires the transformation between the shapes' native frame and ``coordinate_system``
     to be a similarity (uniform scale plus optional rotation and translation). Non-uniform scales,
--- a/src/squidpy/experimental/utils/_derive_mpp.py
+++ b/src/squidpy/experimental/utils/_derive_mpp.py
@@ -1,0 +1,150 @@
+from __future__ import annotations
+
+import numpy as np
+import shapely.affinity
+import spatialdata as sd
+from scipy.spatial import cKDTree
+from spatialdata.models import ShapesModel
+from spatialdata.models._utils import get_axes_names
+from spatialdata.transformations import get_transformation
+
+from squidpy._validators import assert_key_in_sdata
+
+__all__ = ["derive_mpp_from_shapes"]
+
+_ANISOTROPY_TOL = 1e-3
+
+
+def derive_mpp_from_shapes(
+    sdata: sd.SpatialData,
+    shapes_element: str,
+    coordinate_system: str,
+    *,
+    um_between_centers: float | None = None,
+    um_diameter: float | None = None,
+) -> float:
+    """
+    Derive microns-per-pixel for a coordinate system from a shapes element with a known physical scale.
+
+    Given a shapes element (e.g. Visium spots, Visium HD bins) whose physical spacing or diameter is
+    known, this function returns the microns-per-pixel of ``coordinate_system`` by measuring the
+    corresponding geometric quantity in the target coordinate system and dividing the physical value
+    by it.
+
+    Exactly one of ``um_between_centers`` or ``um_diameter`` must be provided.
+
+    The function requires the transformation between the shapes' native frame and ``coordinate_system``
+    to be a similarity (uniform scale plus optional rotation and translation). Non-uniform scales,
+    shear, or other anisotropy raise ``ValueError``: a single scalar microns-per-pixel is not
+    well-defined in that case.
+
+    Parameters
+    ----------
+    sdata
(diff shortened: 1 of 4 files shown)


src/squidpy/experimental/utils/_derive_mpp.py
```diff
@@ -0,0 +1,146 @@
...
+    The function requires the transformation between the shapes' native frame and ``coordinate_system``
+    to be a similarity (uniform scale plus optional rotation and translation). Non-uniform scales,
+    shear, or other anisotropy raise ``ValueError``: a single scalar microns-per-pixel is not
+    well-defined in that case.
+
+    Parameters
+    ----------
+    sdata
+        SpatialData object containing the shapes element.
+    shapes_element
```
selmanozleyen: maybe name it `shapes_element_key`? to match the existing patterns. idk

tests/experimental/test_derive_mpp.py
```diff
@@ -0,0 +1,165 @@
+from __future__ import annotations
+
+import math
+
+import geopandas as gpd
+import numpy as np
+import pytest
+from shapely.geometry import MultiPolygon, Point, Polygon
```
selmanozleyen: This gives a warning so I'd suggest this:
```suggestion
from shapely import MultiPolygon, Point, Polygon
```

The warning
```
  /Users/selman/miniforge3/lib/python3.12/site-packages/geopandas/_compat.py:7: DeprecationWarning: The 'shapely.geos' module is deprecated, and will be removed in a future version. All attributes of 'shapely.geos' are available directly from the top-level 'shapely' namespace (since shapely 2.0.0).
```

src/squidpy/experimental/utils/_derive_mpp.py
```diff
@@ -0,0 +1,146 @@
...
+    return _mpp_from_diameter(gdf, A, um_diameter)
+
+
+def _mpp_from_pitch(gdf: gpd.GeoDataFrame, A: np.ndarray, t: np.ndarray, um_between_centers: float) -> float:
+    n = len(gdf)
+    if n < 2:
+        raise ValueError("Pitch is undefined for a single shape; pass `um_diameter` instead.")
+    centroids = gdf.geometry.centroid
+    xy_native = np.column_stack([centroids.x.to_numpy(), centroids.y.to_numpy()])
+    if n > _PITCH_MAX_SAMPLES:
```
selmanozleyen, line 162: This is what my ai agent told me:
```
large um_between_centers inputs return biased mpp because nearest neighbors are computed after random subsampling. For Visium HD-scale grids, most true adjacent neighbors are dropped, so the sampled nearest-neighbor distance is much larger than the real pitch. I reproduced a 400 x 400 square grid with pitch 8: full median NN is 8.0, sampled median NN is 22.63, so mpp becomes 0.354 instead of 1.0.
```

I don't think this branch is being tested. Maybe we should either throw a warning if we are going to keep it as is and make sure it is somewhat acceptable in unit tests

src/squidpy/experimental/utils/_derive_mpp.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def _mpp_from_diameter(gdf: gpd.GeoDataFrame, A: np.ndarray, um_diameter: float) -> float:
+    geom_types = set(gdf.geometry.geom_type.unique())
+    if geom_types == {"Point"}:
+        if "radius" not in gdf.columns:
+            raise ValueError("Point shapes element is missing the 'radius' column required for diameter-based mpp.")
+        scale = float(np.sqrt(abs(np.linalg.det(A))))
+        diam_target = float(np.median(2.0 * gdf["radius"].to_numpy())) * scale
+    elif geom_types <= {"Polygon"}:
```
selmanozleyen: why not `elif geom_types == {"Polygon"}:` to be more clear?

src/squidpy/experimental/utils/_derive_mpp.py
```diff
@@ -0,0 +1,146 @@
...
+        # area transforms by |det A| under any affine, so we avoid per-geometry shapely calls
+        det = float(abs(np.linalg.det(A)))
```
selmanozleyen: This is only true for rectangular polygones right? Maybe we need to document that?

selmanozleyen (approved):
Looks good except the points I mentioned. Maybe 

```
xy_target = xy_native @ A.T + t
query_xy = xy_target
if n > _PITCH_MAX_SAMPLES:
    rng = np.random.default_rng(0)
    query_xy = xy_target[rng.choice(n, size=_PITCH_MAX_SAMPLES, replace=False)]
nn_dist = cKDTree(xy_target).query(query_xy, k=2)[0][:, 1]
```
would be better for accuracy? Idk

src/squidpy/experimental/utils/_derive_mpp.py
timtreis: Will rename to `shapes_key`

src/squidpy/experimental/utils/_derive_mpp.py
timtreis: True, need to sample only local neighbours

PR #1175, opened 2026-05-12
fix: remove test_raw_has_different_n_obs: raw assignment check since its part of anndata now

selmanozleyen:
ligrec tests used to raise an error with mismatched n_obs with raw in test_raw_has_different_n_obs. Since https://github.com/scverse/anndata/pull/2351 the raw n_obs checks are inside anndata. Also our ci's use 12.11> so we can just remove this from our tests.


timtreis (approved):
lgtm

PR #1179, opened 2026-05-15
chore: add flaky marker to pyproject

selmanozleyen:
similar to the gpu marker we needed to add even though the CI's worked here: https://github.com/scverse/squidpy/pull/1140. We need to add the flaky marker explicitly here.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -108,11 +108,11 @@ optional-dependencies.leiden = [
 ]
 optional-dependencies.test = [
   "coverage[toml]>=7",
-  "flaky",
   "pytest>=7",
   # Just for VS Code
   "pytest-cov>=4",
   "pytest-mock>=3.5",
+  "pytest-rerunfailures",
   "pytest-timeout>=2.1",
   "pytest-xdist>=3",
   "scanpy[leiden]",
@@ -262,7 +262,6 @@ addopts = [
 markers = [
   "internet: tests that require internet",
   "gpu: tests that require GPU",
-  "flaky: tests that may be retried",
 ]
 
 [tool.coverage]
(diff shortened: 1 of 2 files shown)


flying-sheep (changes requested):
The dependency isn’t called `flaky`, it’s `pytest-rerunfailures`. And the marker section is only for markers *we* add, plugins will register their markers:

1. pytest will discover and load the plugin if the package is installed: https://github.com/pytest-dev/pytest-rerunfailures/blob/9f792d9efe6bf0218e7ba2734257af2d5165ca3f/pyproject.toml#L48
2. the plugin will register the marker: https://github.com/pytest-dev/pytest-rerunfailures/blob/9f792d9efe6bf0218e7ba2734257af2d5165ca3f/src/pytest_rerunfailures.py#L343-L350

selmanozleyen:
> The dependency isn’t called `flaky`, it’s `pytest-rerunfailures`. And the marker section is only for markers _we_ add, plugins will register their markers:
> 
>     1. pytest will discover and load the plugin if the package is installed: https://github.com/pytest-dev/pytest-rerunfailures/blob/9f792d9efe6bf0218e7ba2734257af2d5165ca3f/pyproject.toml#L48
> 
>     2. the plugin will register the marker: https://github.com/pytest-dev/pytest-rerunfailures/blob/9f792d9efe6bf0218e7ba2734257af2d5165ca3f/src/pytest_rerunfailures.py#L343-L350

But this means we got flaky from somewhere else because pytest-rerunfailures takes `reruns` as parameter but it worked ok when I used to give it `max_runs`

flying-sheep (approved):

PR #1185, opened 2026-05-20
chore: switch to dependency groups

flying-sheep:
Switches to dependency groups so the development dependency groups `test`/`dev`/`docs` are no longer “features” that users can see and install when installing squidpy.

Also formatted `hatch.toml` in the way that pyproject-fmt does. I can undo that if you don’t like it.


selmanozleyen:
thank you!

selmanozleyen (approved):

PR #1190, opened 2026-05-27
[2/4] Add calculate_image_features with skimage + squidpy-native features

timtreis:
PR 2/4 of the #982 split (plan: PR-1 tests, PR-2 skimage+squidpy features, PR-3 alignment, PR-4 cp_measure).

Introduces `squidpy.experimental.im.calculate_image_features` for per-cell feature extraction from segmentation masks. This is the foundation that PR-3 (alignment) and PR-4 (cp_measure) build on; **cp_measure integration itself lands in PR-4** so the new hard-dependency conversation (`cp-measure`, `centrosome`) is isolated.

## What's in

- skimage `regionprops` features (`skimage:label`, `skimage:label+image`, fine-grained per-property).
- squidpy-native features: `squidpy:summary`, `squidpy:texture` (GLCM), `squidpy:color_hist`.
- Orchestration: `DropReport`, validation, lazy `_prepare_lazy`, centroid computation, parallel tile dispatch via `joblib`.
- Public function `calculate_image_features(sdata, image_key, labels_key=None, shapes_key=None, features=..., align_mode="strict", ...)`.

## What's deferred to follow-up PRs

- **`cpmeasure:*` feature flags** (PR-4). Flag names are recognised here and raise `NotImplementedError` pointing to the follow-up, not "unknown feature".
- **`align_mode="rasterize"`** (PR-3). Raises `NotImplementedError`. PR-2 supports only `align_mode="strict"` (image and labels must share the same pixel grid).
- **`_classify_dropped_cells`** (per-cell drop-reason diagnostic). Deferred; `DropReport` still tracks `empty_tiles`.
- **`features=None`** now raises. The cp_measure-as-default behaviour lands with PR-4.

## No new hard dependencies

PR-2 adds zero new `pyproject.toml` deps. `cp-measure` and `centrosome` land with PR-4.


src/squidpy/experimental/im/_feature.py
```diff
@@ -0,0 +1,739 @@
...
+# ---------------------------------------------------------------------------
+
+
+def calculate_image_features(
+    sdata: SpatialData,
+    image_key: str,
+    labels_key: str | None = None,
+    shapes_key: str | None = None,
+    scale: str | None = None,
+    channels: list[str] | list[int] | None = None,
```
selmanozleyen: `list[int]` errors I think, is this defered to the next PR's?

src/squidpy/experimental/im/_feature.py
```diff
@@ -0,0 +1,739 @@
...
+        Which features to compute.  Required (``None`` is rejected).
+        Accepts a list of strings:
+
+        - ``"skimage:label"`` (all mask props), ``"skimage:label:area"``
+          (single prop), ``"skimage:label+image"`` (all intensity props),
+          ``"skimage:label+image:intensity_mean"`` (single prop)
+        - ``"squidpy:summary"``, ``"squidpy:texture"``,
+          ``"squidpy:color_hist"``
+
+        ``cpmeasure:*`` flag names are recognised but currently raise
```
selmanozleyen: There should be ideally an explanation on each one here.

selmanozleyen (commented):
**label + image naming**
Names `skimage:label` and `skimage:label+image` really don't make sense to me. If we are always going to need a label and an image why would we need to name it `label+image`? Why not call it morphology for "label" and "intensity" for "label+image" for example? Adding the + sign makes the syntax more confusing and is named based on implementation specific jargon.

**mandatory `image_key`**
I can see users getting annoyed when they just want "skimage:label" but don't want to provide image_key. Why is it not optional? I am not sure why the implementation here always needs an image for features that don't need an image. This needs to be solved somehow. Would this make sense `calculate_image_features(sdata, labels_key="seg", features=["skimage:morphology"])
`?

selmanozleyen:
Another comment I'd like to make is the folder and file structure. At this point I would rather have smaller but more files and folder nesting. I am saying this in general but also for this function I am not sure why it's not called `_calculate_image_features.py`. If there is another things coming to the file, I'd still like to be able to know which function would belong where. For example it is also annoying that the spatial_neighbours function is in a file called just _build.py

src/squidpy/experimental/im/_feature.py
```diff
@@ -0,0 +1,739 @@
...
+
+    if label_props is not None:
+        regions = measure.regionprops(labels)
+        rows = {r.label: _regionprops_to_row(r, label_props) for r in regions}
+        parts.append(pd.DataFrame.from_dict(rows, orient="index"))
+
+    if intensity_props is not None:
+        for ch_idx, ch_name in enumerate(channel_names):
+            regions = measure.regionprops(labels, intensity_image=image[ch_idx])
+            rows = {r.label: _regionprops_to_row(r, intensity_props) for r in regions}
```
selmanozleyen: This is potentially very slow!

Use skimage.measure.regionprops_table (which the legacy API used). It executes the loops internally in Cython, returning flat NumPy arrays directly. You can pass the multi-channel image as intensity_image=np.moveaxis(image, 0, -1) to get all channels at once, which is orders of magnitude faster and memory-efficient.

Check out ImageContainer, I think this is done in a more optimized way there

src/squidpy/experimental/im/_calculate_image_features.py
```diff
@@ -0,0 +1,739 @@
...
+    n_pixels = labels_da.sizes.get("y", 1) * labels_da.sizes.get("x", 1)
+    if n_pixels <= 4096 * 4096:
+        lbl_np = labels_da.values
+        if lbl_np.ndim > 2:
+            lbl_np = lbl_np.squeeze()
+        return compute_cell_info(lbl_np)
+
+    # Large single-scale → tiled centroid computation
+    logg.info("Computing centroids in tiled mode (large single-scale labels).")
+    return compute_cell_info_tiled(labels_da)
```
selmanozleyen, line 614: compute_cell_info_tiled might need a second look for optimization.

src/squidpy/experimental/im/_calculate_image_features.py
```diff
@@ -0,0 +1,739 @@
...
+        return pd.DataFrame()
+    return pd.concat(parts, axis=1)
+
+
+# ---------------------------------------------------------------------------
+# squidpy per-cell features
+# ---------------------------------------------------------------------------
+
+
+def _compute_squidpy_per_cell(
```
selmanozleyen, line 362: This function can be numbafied. In fact I can do it if you want. (Maybe even in a separate PR)

src/squidpy/experimental/im/_feature.py
timtreis: Yeah exactly. Splitting the functionality up in some many PRs massively bloats the PR quite a bit so I didn't want to add code just to remove it tomorrow

src/squidpy/experimental/im/_feature.py
timtreis: Will defer this to the notebook that showcases the functionality. The names are fairly descriptive, these are distinct feature groups within the packages

src/squidpy/experimental/im/_calculate_image_features.py
timtreis: Yeah let's do that in a follow-up

src/squidpy/experimental/im/_calculate_image_features.py
timtreis: Yeah, I think once everyhting is in, I'll do another full optimisation scan

src/squidpy/experimental/im/_feature.py
timtreis: Yeah, I think I was doing it that way for the granularity in selecting features, but if it's a lot slower, I might engineer around that

selmanozleyen:
Hi one other comment I can make here is the codecov has many missing lines. Maybe we can add tests for those branches

timtreis:
Increased coverage in this PR to 87 % 👌

timtreis:
Also renamed `skimage:label` to `skimage:morphology` and `skimage:label+image` to `skimage:intensity`. The other terms came from using them as such in a project, but you're probably right, the other ones might be easier for a user seeing the function for the first time. Also made `image_key` optional. I guess I considered that option dead because they could just directly run regionprops then, better to defend against them running it through us 👌

selmanozleyen (approved):
looks good, thanks for incorperating the changes

PR #1191, opened 2026-05-28
Stain normalization: Reinhard color transfer

timtreis:
## Summary

First public surface for the experimental stain-normalization module (builds on #1178). Adds two method-dispatched entry points and implements the Reinhard (2001) color-transfer branch end-to-end:

- `fit_stain_reference(sdata, image_key, *, method="reinhard", scale="auto", method_params=None) -> StainReference`
- `apply_stain_normalization(sdata, image_key, reference, *, scale="auto", method_params=None, image_key_added=None) -> xr.DataArray | None`

Reinhard standardizes a source image by its own tissue statistics and rescales to a fitted reference, working in the Ruderman Lab space the foundations already provide. `macenko`/`vahadane` are recognized by the dispatch but raise `NotImplementedError`; the public signatures won't change when they land.


selmanozleyen (approved):
Honestly I am a bit far to the science here. Would take some time to be convinced of the functionality but the software side seems good. If you want I can take a deeper look but it seems good

timtreis:
The science behind all that is kind of "old news", plenty of reviews out there: https://www.sciencedirect.com/science/article/pii/S1566253523003135

HistomicsTK implements a lot of it but with a different, less flexible toolchain imho, hence the `SpatialData` version here.

For larger scale H&E FM training people even stopped stain normalisation I think because it was shown that just having millions of slides makes it unneccecary but I still want this for situations with smaller sample size, given that that's what we mostly encounter in our spatial experiments where the H&E used to be primarily eye candy

PR #1188, opened 2026-05-27
[1/3] Bundle TilingQCParams; add resolve_labels_array helper

timtreis:
PR 1/3 of the #1170 split. 

```python
# before
calculate_tiling_qc(sdata, "labels", distance_tol=1.0, min_area=30, max_contour_points=400)

# after
calculate_tiling_qc(sdata, "labels", tiling_qc_params={"distance_tol": 1.0, "min_area": 30, "max_contour_points": 400})
# or TilingQCParams(distance_tol=1.0, ...)
```

`adata.uns["tiling_qc"]` no longer carries flat `distance_tol`/`min_area`/`max_contour_points`; they live under `["tiling_qc_params"]`.

Also adds a private `resolve_labels_array` helper in `experimental/utils/_labels.py` (consumed here, reused by PR-B/PR-C). A strict variant of the existing `get_element_data` — fail-loud rather than fall-back, so downstream label pipelines can't silently pick the wrong scale.

PR-B will add `stitch_tile_cuts` + `StitchParams` and the QC-side warning hook. PR-C will add `make_stitched_labels`.

PR #1193, opened 2026-05-29
[2/3] Add assign_stitch_groups (tile-cut stitching algorithm)

timtreis:
## What

Second PR of the 3-way split of #1170. Adds `sq.experimental.tl.assign_stitch_groups` (+ `StitchParams`): given the `is_outlier=True` cells flagged by `calculate_tiling_qc` (PR-A, #1188), it pairs facing cut edges across tile boundaries, scores each candidate pair, and assembles high-confidence pairs into stitch groups via union-find.

It **only annotates** `.obs` (four columns) plus a `.uns["tiling_stitch"]` audit block; the labels element is **never modified**. Materialising a stitched labels element is PR-C (`make_stitched_labels`, follows this one).

```python
sq.experimental.tl.calculate_tiling_qc(sdata, "labels")      # PR-A (merged)
sq.experimental.tl.assign_stitch_groups(sdata, "labels")     # this PR
# sq.experimental.im.make_stitched_labels(sdata, "labels")   # PR-C (next)
```

## The score: a transparent flat mean, not a fitted model

@selmanozleyen asked to review the math/intuition of `stitch_confidence` and whether it is a novel/fitted formula. Short answer: **it is not fitted.** An earlier prototype on this branch (`ba60119f`) shipped an L2-regularised logistic regression with coefficients fit on 2197 synthetic disk pairs. We **removed it deliberately** — fixture-fit weights claim a calibration they cannot honour on real data, and squidpy should not silently encode a synthetic distribution. The score is now the **flat (unweighted) mean of five standard, bounded geometric descriptors**, each in `[0, 1]`. The features are recorded in `.uns["tiling_stitch"]`.

For a facing cut-edge pair (each a 1-D chord `extent=(lo, hi)` at perpendicular `coord`):

1. **`iou`** — 1-D extent IoU of the two chords: `overlap / (len_a + len_b - overlap)`. Do the two cut faces line up laterally?
2. **`endpoint_match`** — `max(0, 1 - (|a_lo-b_lo| + |a_hi-b_hi|) / max(len_a, len_b))`. Do the faces start/end at the same lateral positions?
3. **`merge_compactness`** — `min(4*pi*A / P^2, 1)` of the closed union mask's largest component. A re-joined real cell is compact; a false merge is elongated. Strongest single discriminator.
4. **`merge_solidity`** — `area / convex_hull_area` of the same component (clamped to 1). A false merge creates a concave waist at the join.
5. **`gap_proximity`** — `clip(1 - gap / (2*close_radius), 0, 1)`. The seam gap relative to the **closing reach** (the scale at which morphological closing could actually bridge it) — independent of the `max_gap` search radius; neutral (`1.0`) when closing is disabled.

`stitch_confidence(pair) = mean(the five features)`. Per-cell group confidence is the `min` over the group's pairwise confidences (weakest link).

### Runnable MREs (pure numpy/scipy/skimage, no squidpy)

The 1-D edge features:

```python
def extent_iou(a, b):
    o = max(0.0, min(a[1], b[1]) - max(a[0], b[0]))
    u = (a[1] - a[0]) + (b[1] - b[0]) - o
    return o / u if u > 0 else 0.0

def endpoint_match(a, b):
    d = abs(a[0] - b[0]) + abs(a[1] - b[1]); m = max(a[1] - a[0], b[1] - b[0])
    return max(0.0, 1.0 - d / m) if m > 0 else 0.0

# true cut: two faces span the same range          false pair: barely overlapping
assert extent_iou((0, 10), (0, 10)) == 1.0
assert extent_iou((0, 10), (8, 18)) < 0.2
assert endpoint_match((0, 10), (0, 10)) == 1.0
```

The 2-D merged-shape features (error handling elided):

```python
import numpy as np
from scipy.ndimage import binary_closing
from skimage.morphology import disk
from skimage.measure import label, regionprops

def merge_shape(mask, close_radius=3):
    closed = binary_closing(mask, structure=disk(close_radius))
    cc = label(closed, connectivity=2); s = np.bincount(cc.ravel()); s[0] = 0
    r = regionprops((cc == s.argmax()).astype(np.uint8))[0]; p = max(r.perimeter, 1.0)
    return min(r.solidity, 1.0), min(4 * np.pi * r.area / (p * p), 1.0)

# true cut: two halves of one disk, 2px seam -> high solidity + compactness
g = np.zeros((40, 40), bool); yy, xx = np.ogrid[:40, :40]
disk_mask = (yy - 20) ** 2 + (xx - 20) ** 2 <= 12 ** 2
true_cut = disk_mask.copy(); true_cut[19:21, :] = False     # cut into two halves
# false merge: two separate disks side by side -> concave waist, low solidity
false_merge = ((yy - 20) ** 2 + (xx - 12) ** 2 <= 7 ** 2) | ((yy - 20) ** 2 + (xx - 30) ** 2 <= 7 ** 2)
print("true  cut :", merge_shape(true_cut))      # ~ (solidity 0.96, compactness 0.92)
print("false pair:", merge_shape(false_merge))   # ~ (solidity 0.73, compactness 0.47) - concave join
```
shape_features

## Validation (sweep, not fit)

Weights are fixed by reasoning (flat-equal); the ground-truth fixture only **validates** them. Sweeping `min_confidence` over 3 synthetic layouts (recall = fraction of cut pieces stitched; precision = cut / (cut + intact-false-merge)):

| min_confidence | recall (5-feature) | precision | intact false-merges |
|----------------|--------------------|-----------|---------------------|
| 0.3 – 0.7      | 0.64               | 1.00      | 0                   |
| 0.8            | 0.58               | 1.00      | 0                   |
| 0.9            | 0.06               | 1.00      | 0                   |

(The recall ceiling is < 1 because not every cut piece has a stitchable partner.) **Default `min_confidence=0.7`** gives full attainable recall with zero false merges and headroom before the high-threshold drop-off. Users should still tune it for their data — the score is heuristic, not a calibrated probability.

validation_sweep

## Obs / uns contract (consumed additively by PR-C)

- `.obs`: `stitch_group_id` (int), `is_stitched` (bool), `n_pieces` (int32), `stitch_confidence` (float64; `NaN` = not an outlier, `1.0` = evaluated-solo, composite = stitched).
- `.uns["tiling_stitch"]`: params, `score_features`, and the run counts.

PR-C's `make_stitched_labels` reads only this contract (it does not import `_tiling_stitch`).

## Notes for review

- **Codecov**: expect the patch % to dip on `_tiling_stitch.py` (defensive branches in the geometry helpers); the algorithm paths are covered by `tests/experimental/test_tiling_stitch.py`.
- **`hatch-test.py3.13-pre`**: the pre-release-deps job is a known upstream flake (pre-existing on #1170/#1188), not caused by this PR.
- **Visual test**: `TestStitchVisual` renders a seam recolour (before by `label_id`, after by `stitch_group_id`); its baseline PNG is bootstrapped from the first CI run's artifact (not generated locally).
- Also extracts three shared helpers (`resolve_params`, `equivalent_diameter` + `largest_contour`, `iter_chunked_regionprops`) into `experimental/utils/`, reused by `_tiling_qc`/`_tiling` to remove duplication. QC numerics are unchanged — the committed QC/tiling visual baselines pass.

## Out of scope

- `make_stitched_labels` + materialisation — **PR-C** (next, stacks on this).


selmanozleyen:
I have a problem with the uns keys we give so far like: .uns["tiling_stitch"]. Can't we make a rule that whatever we write to uns as metadata we use the name of the function that wrote it? I would be more obvious imo. Also a test fails atm

src/squidpy/experimental/tl/_tiling_stitch.py
```diff
@@ -0,0 +1,878 @@
...
+    """Compute shape features per candidate, score, and keep pairs >= min_confidence.
+
+    One entry per ``(cell_a, cell_b, axis)`` (keeping max confidence on duplicates).
+    """
+    scored: list[_StitchPair] = []
+    for e, c, geom in candidates:
+        shape = _merge_shape_features(labels_da, [e.cell_id, c.cell_id], bboxes, close_radius=close_radius)
+        feats = {**geom, **shape, "gap_proximity": _gap_proximity(geom["gap"], close_radius)}
+        confidence = _score_pair_features(feats)
+        if confidence < min_confidence:
```
selmanozleyen, line 553: we can have early pruning here for cases when min_confidence is. For example:
```python
gap_prox = _gap_proximity(geom["gap"], close_radius)
max_possible = (geom["iou"] + geom["endpoint_match"] + gap_prox + 2.0) / 5.0
if max_possible < min_confidence:
    continue  # Prune early, skipping
```

we'd also need to add a test case to check if this works or not

src/squidpy/experimental/tl/_tiling_stitch.py
```diff
@@ -0,0 +1,878 @@
...
+
+
+def _merge_shape_features(
+    labels_da: xr.DataArray | np.ndarray,
+    cell_ids: Iterable[int],
+    bboxes: dict[int, tuple[int, int, int, int]],
+    close_radius: int = _STITCH_DEFAULTS.close_radius,
+) -> dict[str, float]:
+    """Materialise the union of given pieces, close the gap, and return shape stats.
+
```
selmanozleyen, line 370: I tested this function and it's quite slow because of fetching xarrays. This one recreates binary masks from the information already available 

```suggestion
def _merge_shape_features(
    cell_a: int,
    cell_b: int,
    bboxes: dict[int, tuple[int, int, int, int]],
    outlier_crops: dict[int, np.ndarray],
    close_radius: int,
    H: int,
    W: int,
) -> dict[str, float]:
    """Reconstruct a union mask locally from pre-fetched crops using spatial offsets.

    Includes boundary clamping to match the original global-clamping behavior
    at the image borders exactly.
    """
    pad = close_radius + 2
    r0_a, c0_a, r1_a, c1_a = bboxes[cell_a]
    r0_b, c0_b, r1_b, c1_b = bboxes[cell_b]

    # Calculate union bounding box and clamp exactly like the original code
    r0 = max(min(r0_a, r0_b) - pad, 0)
    c0 = max(min(c0_a, c0_b) - pad, 0)
    r1 = min(max(r1_a, r1_b) + pad, H)
    c1 = min(max(c1_a, c1_b) + pad, W)

    # Allocate a small blank union mask in memory
    union_mask = np.zeros((r1 - r0, c1 - c0), dtype=bool)

    # Place Cell A's mask relative to the union origin
    crop_a = outlier_crops[cell_a]
    mask_a = crop_a == cell_a
    union_mask[r0_a - r0 : r1_a - r0, c0_a - c0 : c1_a - c0] |= mask_a

    # Place Cell B's mask relative to the union origin
    crop_b = outlier_crops[cell_b]
    mask_b = crop_b == cell_b
    union_mask[r0_b - r0 : r1_b - r0, c0_b - c0 : c1_b - c0] |= mask_b

    # Perform morphological closing and feature extraction on the tiny local mask
    closed = binary_closing(union_mask, structure=morph_disk(close_radius))
    cc = cc_label(closed, connectivity=2)
    if cc.max() == 0:
        return {"merge_solidity": 0.0, "merge_compactness": 0.0}

    sizes = np.bincount(cc.ravel())
    sizes[0] = 0
    biggest = int(sizes.argmax())
    region = regionprops((cc == biggest).astype(np.uint8))[0]

    perimeter = max(region.perimeter, 1.0)
    compactness = float(min(4 * np.pi * region.area / (perimeter * perimeter), 1.0))
    solidity = float(min(region.solidity, 1.0))

    return {"merge_solidity": solidity, "merge_compactness": compactness}
```

selmanozleyen (approved):
There can be probably more performance issues but I think we just need to see how this performs in real life. So we can tackle the real bottlenecks

src/squidpy/experimental/tl/_tiling_stitch.py
timtreis: thanks!

timtreis:
> There can be probably more performance issues but I think we just need to see how this performs in real life. So we can tackle the real bottlenecks

Agree, the chopping into sub-PR thing bloated everything a bit. The final 3/3 PR now will hopefully tie everything back together so that we can actually optimise

selmanozleyen (approved):

PR #1195, opened 2026-06-01
feat: stalign api and landmark alignment

selmanozleyen:
Continuation of https://github.com/scverse/squidpy/pull/1162


src/squidpy/experimental/tl/__init__.py
```diff
@@ -1,6 +1,24 @@
...
-__all__ = ["StitchParams", "TilingQCParams", "assign_stitch_groups", "calculate_tiling_qc"]
+__all__ = [
+    "align",
+    "align_by_landmarks",
+    "calculate_tiling_qc",
+    "TilingQCParams",
+    "AlignResult",
+    "StitchParams",
+    "TilingQCParams",
+  "assign_stitch_groups",
```
timtreis: format

src/squidpy/experimental/tl/_align/_io.py
```diff
@@ -0,0 +1,170 @@
...
+        if isinstance(element, AnnData):
+            continue
+        if moving_cs not in get_transformation(element, get_all=True):
+            continue
+        if output_mode == "copy":
+            # `shallow_copy_sdata` shares element objects with the original; deep-copy each
+            # element we register a transform on so `output_mode="copy"` leaves the input untouched.
+            element = sd_deepcopy(element)
+            getattr(out, etype)[name] = element
+        set_transformation(element, sd_affine, to_coordinate_system=target_cs)
```
timtreis: breaks if the element already has a transformation, you need to `Sequence` them

```
existing = get_transformation(element, to_coordinate_system=moving_cs)
set_transformation(element, Sequence([existing, sd_affine]), to_coordinate_system=target_cs)
```

src/squidpy/experimental/methods/align_samples/_stalign_impl/_helpers.py
```diff
@@ -0,0 +1,126 @@
...
+    grid_y = np.arange(center_y - half_y, center_y + half_y + dx, dx, dtype=float)
+    if grid_x.size < 2 or grid_y.size < 2:
+        raise ValueError("Rasterized grid is too small. Increase the point spread or lower `dx`.")
+
+    mesh_x, mesh_y = np.meshgrid(grid_x, grid_y)
+    out = np.zeros((len(blur_values), grid_y.size, grid_x.size), dtype=float)
+    radius = int(np.ceil(float(np.max(blur_values)) * 4.0))
+    denom = 2.0 * (dx * blur_values * 2.0) ** 2
+
+    for x_i, y_i in zip(x, y, strict=False):
```
timtreis: How many points do we have? Should probably rasterize

src/squidpy/experimental/methods/align_samples/_stalign_impl/_core.py
```diff
@@ -0,0 +1,369 @@
...
+    *,
+    L: np.ndarray | jax.Array,
+    T: np.ndarray | jax.Array,
+    points_source: np.ndarray | jax.Array | None = None,
+    points_target: np.ndarray | jax.Array | None = None,
+    a: float = 500.0,
+    p: float = 2.0,
+    expand: float = 2.0,
+    nt: int = 3,
+    niter: int = 5000,
```
timtreis: Can we do some form of early stopping or do we always have to train for 5k steps? Might be too much/too little depending on data

src/squidpy/experimental/tl/_align/_io.py
```diff
@@ -0,0 +1,170 @@
...
+            f"Mixed AnnData/SpatialData inputs are not supported; `data_query` is {type(data_query).__name__}."
+        )
+    if ref_key is None or query_key is None:
+        raise ValueError("`ref_key` and `query_key` are required for SpatialData inputs.")
+    assert_key_in_sdata(data_ref, ref_key, attr="tables")
+    assert_key_in_sdata(sdata_query, query_key, attr="tables")
+    return data_ref.tables[ref_key], sdata_query.tables[query_key], sdata_query, query_key
+
+
+def get_coords(adata: AnnData, spatial_key: str) -> np.ndarray:
```
timtreis: Claude: 

> get_coords materializes coords up to 3× per align() (_io.py:71; called at _api.py:83,84 and again _io.py:97). np.asarray(..., dtype=float) forces a float64 copy each time. Thread the already-computed query_xy into writeback_obs and avoid the forced upcast when the array is already suitable.

pyproject.toml
```diff
@@ -76,6 +76,9 @@ dependencies = [
   "xarray>=2024.10",
   "zarr>=3",
 ]
+optional-dependencies.jax = [
+  "jax",
```
timtreis: Shouldn't we have some lower bound?

src/squidpy/experimental/tl/_align/_io.py
selmanozleyen: ok I removed the up casting

src/squidpy/experimental/methods/align_samples/_stalign_impl/_helpers.py
selmanozleyen: wdym? This is the function for rasterization?

src/squidpy/experimental/tl/_align/_io.py
selmanozleyen: thanks, its done!

src/squidpy/experimental/methods/align_landmarks/_landmark.py
```diff
@@ -0,0 +1,155 @@
...
+
+from collections.abc import Callable
+from dataclasses import dataclass, field
+from typing import Any
+
+import numpy as np
+import numpy.typing as npt
+
+from squidpy._utils import NDArrayA
+from squidpy.experimental.methods.registry import ALIGN_LANDMARKS
```
timtreis: Why are we using a registry here? It's unlikely this will get extended upon much and it introduces complexity

src/squidpy/experimental/methods/align_landmarks/_landmark.py
```diff
@@ -0,0 +1,155 @@
...
+            raise ValueError(f"Expected an (N, 2) coordinate array, found shape {coords.shape}.")
+        return coords @ self.matrix[:2, :2].T + self.matrix[:2, 2]
+
+
+def _fit_landmark_relation(
+    ref: np.ndarray,
+    query: np.ndarray,
+    *,
+    method: str,
+    solve_fn: Callable[[np.ndarray, np.ndarray], np.ndarray],
```
timtreis: Shouldn't it be `Callable[np.ndarray, np.ndarray]`?

selmanozleyen: nope the signature is like this `_fit_similarity(ref_xy: np.ndarray, query_xy: np.ndarray) -> np.ndarray:`

src/squidpy/experimental/tl/_align/_api.py
```diff
@@ -0,0 +1,867 @@
...
+        ),
+        initial_slice=initial_slice,
+        initial_rotation=initial_rotation,
+        initial_scale=initial_scale,
+        initial_affine=initial_affine,
+        **solver_kwargs,
+    )
+
+
+def stalign_apply_transform(
```
selmanozleyen: @flying-sheep I am not sure about the api design here. I try to be as notebook friendly and plain for squidpy sake but this seems like a clear case of fit-transform for me.

selmanozleyen: But also most important thing here. The presedence this will bring. Like if we decide on a class based api, what would the implications be for the rest of future and current implementations?

flying-sheep: > notebook friendly

I don’t know what you mean by that – a class instance with a good `repr` is as notebook-friendly as a dict.

flying-sheep: One `question about the “storing in .uns” part – JAX arrays can’t be serialized by AnnData, so that’s OK anyway right? Or does SpatialData support them?

If not serializing is OK, I’d say to just return a class instance. You could define a `Protocol` if you don’t want people to be able to import the class.

----

unrelated – why two independent TypedDicts with overlapping stuff instead of just [overriding](https://typing.python.org/en/latest/spec/typeddict.html#overriding-items) the fields that become required in the 3D variant?

selmanozleyen: > I don’t know what you mean by that – a class instance with a good repr is as notebook-friendly as a dict.

Me too honestly :D, Like from the chats I have with Tim, I feel like I need to restrict myself to an imaginary target audience (which I am not sure if it exists or matter) that doesn't know what a class or a for loop is. 

This is a longer discussion. I really am against this "mimicking R is simplicity for researchers" idea, which forgets there is a reason R is being replaced by python.

Thanks for the override suggestion I am looking at it, and yes not serializing is ok. Also could you have a look at https://github.com/scverse/squidpy/pull/1279? I made a proposal for a general repo-wide separation.

selmanozleyen: Like no need to fully review yet but I am just curious of your opinion on holding the defaults here like:

```python

class WekaParams(TypedDict, total=False):
    """Parameters for WEKA-like trainable segmentation."""

    sigma_min: Annotated[float, Default(1.0)]
    """Smallest scale in the multiscale feature bank."""

    sigma_max: Annotated[float, Default(16.0)]
    """Largest scale in the multiscale feature bank."""

    edges: Annotated[bool, Default(True)]
    """Include edge features."""

    pseudo_tissue_percentile: Annotated[float, Default(90.0)]
    """Percentile of distance-from-bg to label as tissue."""
```

PR #1207, opened 2026-06-14
[3/3] Add make_stitched_labels (materialise stitched labels)

timtreis:
Completes the tile-cut stitching arc (`[1/3]` #1188, `[2/3]` #1193) with the materialisation step.

## Usage

```python
sq.experimental.tl.calculate_tiling_qc(sdata, labels_key="labels")
sq.experimental.tl.assign_stitch_groups(sdata, labels_key="labels")
sq.experimental.im.make_stitched_labels(sdata, labels_key="labels")
# sdata.labels["labels_stitched"]      -> pieces of a cut cell share one ID (original untouched)
# sdata.tables["labels_stitched_table"] -> one row per stitched cell
```

Knobs: `merge_strategy` (how additive user features aggregate across pieces; default `"sum"`), `join_labels` (close the seam so a group is one connected component; stays lazy on dask), `write_table`, `inplace`.

## Also in this PR

- **De-bloat** of the merged arc: inlined the single-use `utils/_geometry.py` helpers into their one caller (module deleted) and routed `_resolve_qc_params` through the shared `resolve_params`. No behaviour change.

Closes the #1170 arc.


src/squidpy/experimental/tl/_tiling_stitch.py
```diff
@@ -302,10 +301,12 @@ def _extract_cut_edges(
         if not cell_mask.any():
             continue
         outlier_crops[lid] = cell_mask
+        # 1px zero-pad so cells filling their bbox still trace a closed contour.
         mask = np.pad(cell_mask.astype(np.float32), 1, mode="constant", constant_values=0)
-        contour = largest_contour(mask)
-        if contour is None:
+        contours = find_contours(mask, 0.5)
+        if not contours:  # degenerate mask traces nothing; skip it
```
selmanozleyen, line 307: if not contours is ambigious if it means it's 0 the check should be to if contours != 0 or something like this.

src/squidpy/experimental/im/_stitched_labels.py
```diff
@@ -0,0 +1,546 @@
...
+    qc_table_key
+        Key of the QC table.  Defaults to ``"{labels_key}_qc"``.
+    labels_key_added
+        Key for the new labels element.  Defaults to
+        ``"{labels_key}_stitched"``.  Existing element at this key is
+        overwritten with a warning.
+    table_key_added
+        Key for the optional collapsed AnnData (one row per unique
+        ``stitch_group_id``).  Defaults to ``"{labels_key_added}_table"``
+        (must differ from the labels element key -- SpatialData requires
```
selmanozleyen, line 433: "must differ from the labels element key"
This isn't enforced in the code I think

docs/release/notes-dev.md
```diff
@@ -2,6 +2,7 @@
 
 ## Features
 
+- Add {func}`squidpy.experimental.im.make_stitched_labels` to materialise a stitched labels element (and an optional collapsed table) from an {func}`squidpy.experimental.tl.assign_stitch_groups` result, completing the tile-cut stitching workflow.
```
selmanozleyen: we don't usually add these nowadays right?

tests/experimental/test_stitched_labels.py
```diff
@@ -0,0 +1,485 @@
...
+        after_arr = np.asarray(sdata.labels["labels"].values)
+        np.testing.assert_array_equal(original_arr, after_arr)
+
+    def test_remap_unifies_stitched_pieces(self, sdata_tile_boundary):
+        sdata, _ = sdata_tile_boundary
+        _qc_and_stitch(sdata, min_confidence=0.5)
+        adata = sdata.tables["labels_qc"]
+        stitched = adata.obs[adata.obs["is_stitched"].astype(bool)]
+        if len(stitched) == 0:
+            pytest.skip("no stitched cells in this fixture realisation")
```
selmanozleyen: why do we need this? Isn't the sdata_tile_boundary deterministic?

src/squidpy/experimental/im/_stitched_labels.py
```diff
@@ -0,0 +1,546 @@
...
+        :func:`squidpy.experimental.tl.assign_stitch_groups`.
+    labels_key
+        Key in ``sdata.labels`` of the original labels element.
+    qc_table_key
+        Key of the QC table.  Defaults to ``"{labels_key}_qc"``.
+    labels_key_added
+        Key for the new labels element.  Defaults to
+        ``"{labels_key}_stitched"``.  Existing element at this key is
+        overwritten with a warning.
+    table_key_added
```
selmanozleyen, line 430: I might be wrong on this, but can't we notice if table_key_added is not None, write_table would be automatically true? I think it's a default at the price of adding a bool argument.

selmanozleyen (commented):
Some nitpicks but it seems quite good!

src/squidpy/experimental/tl/_tiling_stitch.py
timtreis: `if contours != 0` wouldn't work here because upon failure, it'd return an empty list

docs/release/notes-dev.md
timtreis: Yeah was drive-by-edit, I thought I had deleted it already

tests/experimental/test_stitched_labels.py
timtreis: true

src/squidpy/experimental/im/_stitched_labels.py
timtreis: Almost 🤔 We have these states:

A) `write_table=True` + `table_key_added=None` --> write table with default key (`{out_key}_table`)
B) `write_table=True` + `table_key_added="foo"` --> write table with key (`foo`)
C) `write_table=False` + `table_key_added=None` --> no table
D) `write_table=False` + `table_key_added="foo"` --> error

For that granularity of "I don't care about the name, just use defaults", we need to enable the permutations

selmanozleyen: yeah thats what I mean, we add one argument `write_table` for just supporting: "I don't care about the name, just use defaults". This is a preference thing so it's fine I guess

src/squidpy/experimental/tl/_tiling_stitch.py
selmanozleyen: yeah then len(contours) == 0 then right? that's what I mean by being unclear

selmanozleyen (approved):

timtreis:
yey! Will send a tutorial to squidpy_notebooks next

PR #1209, opened 2026-06-14
Multi-slide pooled fit_stain_reference

timtreis:
`fit_stain_reference` now accepts `image_key: str | list[str]` — a list pools several images' tissue pixels into one fit (one reference for a cohort in one SpatialData).

```python
ref = sq.experimental.im.fit_stain_reference(sdata, ["s1", "s2"], method="macenko")
```

`tissue_mask_key` takes an order-matched list (or `None` for the `{key}_tissue` convention). Single-image fit unchanged.


tests/experimental/test_stain_normalize.py
```diff
@@ -251,3 +251,103 @@ def test_plot_reinhard_before_after(self, sdata_hne) -> None:
...
+    )
+    def test_validation(self, image_key, tissue_mask_key, match: str) -> None:
+        sdata, _ = self._cohort(n=2)
+        with pytest.raises(ValueError, match=match):
+            fit_stain_reference(sdata, image_key, method="macenko", tissue_mask_key=tissue_mask_key)
+
+    def test_list_mask_with_str_image_raises(self) -> None:
+        sdata, _ = self._cohort(n=1)
+        with pytest.raises(ValueError, match="single `tissue_mask_key`"):
+            fit_stain_reference(sdata, "img0", method="macenko", tissue_mask_key=["img0_tissue"])
```
selmanozleyen: can these tests be merged into a parameterized one?

src/squidpy/experimental/im/_stain/_normalize.py
```diff
@@ -292,6 +312,63 @@ def fit_stain_reference(
...
+
+    Each image is resolved + validated and masked by its own `tissue_mask_key`
+    (order-matched) or the `{key}_tissue` convention; decomposition pools tissue
+    OD, Reinhard pools tissue Lab pixels. Images must share a dtype so the white
+    point is well-defined. A blank slide raises (named); no silent skip.
+    """
+    if not image_keys:
+        raise ValueError("`image_key` list is empty; pass at least one image key.")
+    if len(set(image_keys)) != len(image_keys):
+        raise ValueError("`image_key` list has duplicate keys.")
```
selmanozleyen: Can't we use _unique_order_preserving here somehow?

src/squidpy/experimental/im/_stain/_normalize.py
selmanozleyen: There are lines duplicated here:

You should normalize to [str] if it's str. Then you'd do this procedure for free on both
```python
   das, masks = [], []
    for k, mk in zip(image_keys, mask_keys, strict=True):
        da = _resolve_image(sdata, k, scale, prefer="coarsest")
        validate_rgb_range(da)
        das.append(da)
        masks.append(_resolve_tissue_bool_mask(sdata, k, da, mk))
    dtypes = {str(da.dtype) for da in das}
    if len(dtypes) != 1:
        raise ValueError(f"pooled images must share a dtype; got {sorted(dtypes)}.")
```

```
    if method == "reinhard":
        return fit_reinhard_pooled(das, masks, params, image_keys)
    bg = default_white_point(das[0]) if white_point is None else np.asarray(white_point, np.float64)
    reference = RUIFROK_HE if canonical_reference is None else dict(canonical_reference)
    return fit_decomposition_pooled(
        das, masks, image_keys, method, params, bg, reference=reference, max_angle_deg=max_angle_deg
    )
```

I'd also do

```
def fit_reinhard_pooled(
    das: list[xr.DataArray],
    masks: list[np.ndarray | None],
    params: ReinhardParams,
    image_keys: list[str | None],
) -> StainReference:
    """Fit Reinhard stats from the pooled tissue Lab pixels of one or more slides."""
    cols = [_tissue_lab_pixels(da, params, m, image_key=k) for da, m, k in zip(das, masks, image_keys, strict=True)]
    return _reinhard_from_pixels(np.concatenate(cols, axis=1))


def fit_reinhard(
    image_rgb: xr.DataArray, params: ReinhardParams, *, tissue_mask: np.ndarray | None = None
) -> StainReference:
    """Single-image Reinhard fit — a pool of one."""
    return fit_reinhard_pooled([image_rgb], [tissue_mask], params, [None])
```
or just have `fit_reainhard(str|list[str])` tbh

selmanozleyen (changes requested):
please do the changes I mentioned. Overall writing the code as if all the inputs are [str] is a good idea and normalizing str to [str] will save duplicate computation and code

src/squidpy/experimental/im/_stain/_normalize.py
timtreis: Done in 3a026a5 — `image_key`/`tissue_mask_key` are normalised to lists at the top of `fit_stain_reference`, one pipeline for both; `_fit_pooled` and both `*_pooled` functions are gone. `fit_reinhard`/`fit_decomposition` take one image or a sequence (pool of one), per your sketch with `_tissue_lab_pixels`.

src/squidpy/experimental/im/_stain/_normalize.py
timtreis: Done — duplicate check now goes through `_unique_order_preserving`.

tests/experimental/test_stain_normalize.py
timtreis: Folded into two parametrized tests: `test_pooled_fit` (runs + pool-of-one == single, exact except vahadane where NMF is not bit-reproducible even seeded) and `test_validation` (all raise cases as one table). `test_order_matched_non_convention_masks` stays separate since it asserts values, not an exception.

selmanozleyen (commented):
I like to first assess this before going on with the experimental functions: https://github.com/scverse/squidpy/pull/1279.

It's about the design and the structure. I can look into this PR meanwhile separately but I'd prefer to do after

PR #1213, opened 2026-06-15
refactor(datasets): use the shared scverse-misc dataset registry + downloader

timtreis:
Replaces squidpy's internal pooch-based dataset registry/downloader with the shared infrastructure from [scverse-misc](https://github.com/scverse/scverse-misc) (`scverse-misc[datasets]>=0.1.1`).

- Public API unchanged: `sq.datasets.*` keep the same signatures and behavior.
- ~750 lines of duplicated registry/download code deleted; `pooch` is now transitive.
- `datasets.yaml` stays in squidpy; the `anndata` / `image` / `visium_10x` loaders register via `register_loader`.


src/squidpy/datasets/_datasets.py
```diff
@@ -122,19 +120,11 @@ def visium(
...
-        raise ValueError(msg)
-
-    # Use scanpy.settings.datasetdir/visium if base_dir not specified
-    if base_dir is None:
-        base_dir = Path(settings.datasetdir) / "visium"
+    if sample_id not in get_registry():
+        raise ValueError(f"Unknown Visium sample: {sample_id}. Available samples: {dataset_names('visium_10x')}")
 
-    return downloader.download(sample_id, base_dir, include_hires_tiff=include_hires_tiff)
+    # downloads land in <datasetdir>/visium_10x/<sample_id>/
```
selmanozleyen, line 130: this used to be `/x/visium/x/` on old code but it's fine I guess to move the cache path. It's also seen in 
```
-    expected_image_path = (Path(settings.datasetdir) / "visium" / sample / "image.tif").resolve()
+    expected_image_path = (Path(settings.datasetdir) / "visium_10x" / sample / "image.tif").resolve()
```
but good to document

src/squidpy/datasets/_registry.py
```diff
@@ -1,198 +1,39 @@
...
-    @property
-    def visium_10x_datasets(self) -> list[str]:
-        """Return names of all 10x Genomics Visium datasets."""
-        return [name for name, entry in self.datasets.items() if entry.type == DatasetType.VISIUM_10X]
 
-    @property
-    def visium_datasets(self) -> list[str]:
-        """Return names of all Visium datasets (alias for visium_10x_datasets)."""
-        return self.visium_10x_datasets
+def get_registry() -> dict[str, DatasetEntry]:
```
selmanozleyen: now we can have a prettier name for _parsed() and get rid of these two functions. No need to have two separate functions get_registry() and get_base_url()

selmanozleyen: `dict[str, DatasetEntry]:` should be MappingProxy because we don't want it to be mutable

pyproject.toml
```diff
@@ -59,13 +59,14 @@ dependencies = [
...
-  "pyyaml>=6",
+  "pooch>=1.6",                    # used directly (pooch.Untar) in the visium loader
   "scanpy>=1.9.3",
   "scikit-image>=0.25",
   # due to https://github.com/scikit-image/scikit-image/issues/6850 breaks rescale ufunc
   "scikit-learn>=0.24",
-  "spatialdata>=0.7.2",          # 0.7.2 dropped xarray-schema (pkg_resources break, #1115)
+  # dataset registry + downloader now live in scverse-misc
+  "scverse-misc[datasets]>=0.1.1",
+  "spatialdata>=0.7.2",            # 0.7.2 dropped xarray-schema (pkg_resources break, #1115)
```
selmanozleyen: why the diff here, can you undo it?
```
"spatialdata>=0.7.2",            # 0.7.2 dropped xarray-schema (pkg_resources break, #1115)
```

src/squidpy/datasets/_datasets.py
```diff
@@ -122,19 +120,11 @@ def visium(
...
-    if sample_id not in downloader.registry:
-        msg = f"Unknown Visium sample: {sample_id}. "
-        msg += f"Available samples: {downloader.registry.visium_datasets}"
-        raise ValueError(msg)
-
-    # Use scanpy.settings.datasetdir/visium if base_dir not specified
-    if base_dir is None:
-        base_dir = Path(settings.datasetdir) / "visium"
+    if sample_id not in get_registry():
+        raise ValueError(f"Unknown Visium sample: {sample_id}. Available samples: {dataset_names('visium_10x')}")
```
selmanozleyen: valid-but-wrong-type name (e.g. imc, an AnnData dataset) passes the guard, then fails deep inside the anndata loader with a confusing TypeError: read_h5ad() got an unexpected keyword argument 'include_hires_tiff' instead of a clear error.
```suggestion
    visium_samples = dataset_names("visium_10x")
    if sample_id not in visium_samples:
        raise ValueError(f"Unknown Visium sample: {sample_id}. Available samples: {visium_samples}")
```

tests/datasets/test_registry.py
```diff
@@ -1,235 +1,61 @@
...
-        assert registry1 is registry2
+    def test_visium_10x_entry(self):
+        sample = get_registry()["V1_Adult_Mouse_Brain"]
+        assert sample.type == "visium_10x"
+        assert len(sample.files) == 3
+        assert any(f.name.startswith("image.") for f in sample.files)
+
+    def test_unknown_dataset(self):
+        assert "nonexistent" not in get_registry()
+
```
selmanozleyen, line 57: Maybe let's add a test like this given that it fails
```python
class TestVisiumValidation:
    """Offline validation of the ``visium()`` sample-name guard (no network)."""

    def test_unknown_sample_raises(self):
        with pytest.raises(ValueError, match="Unknown Visium sample"):
            visium("definitely_not_a_sample")

    def test_valid_but_wrong_type_rejected(self):
        # `imc` is a valid registry entry but an AnnData dataset, not a Visium 10x
        # sample. It must be rejected before any download is attempted.
        with pytest.raises(ValueError, match="Unknown Visium sample"):
            visium("imc")
```

selmanozleyen (changes requested):
See my comments. But this is a nice win!

docs/release/notes-dev.md
selmanozleyen, line 1: We don't really use this file. Maybe we should remove it? Or make a note each release to somehow add these notes into the autogenerated logs

selmanozleyen (approved):
LGTM

PR #1219, opened 2026-06-19
test: replace deprecated spatial_neighbors with mode-specific functions

selmanozleyen:
Converts incidental uses of the deprecated spatial_neighbors to the new mode-specific functions (spatial_neighbors_grid/_knn/_radius). Leaves tests/graph/test_spatial_neighbors.py untouched since it intentionally exercises the deprecated path.


timtreis:
Missed one I think (see pre-commit)

timtreis (approved):

PR #1220, opened 2026-06-19
use numba progress in threadpool

selmanozleyen:
Using numba-progress for the default case. In case of custom progress bar handling custom logic can be used. But that case can be explored in future PR's


src/squidpy/_utils.py
```diff
@@ -19,7 +19,11 @@
 import numpy as np
 import xarray as xr
 from spatialdata.models import Image2DModel, Labels2DModel
-from tqdm.auto import tqdm
+
+if TYPE_CHECKING:
+    from contextlib import AbstractContextManager
```
timtreis: Why do we need this?

selmanozleyen: to annotate the return type

timtreis: I find the pattern a bit too complicated for what we're doing there? Maybe just refactor as a context manager, then we avoid that issue and have simpler typing too?

timtreis: like

```
@contextmanager
def progress_bar(
    total: int,
    *,
    show_progress_bar: bool = True,
    unit: str = "item",
    desc: str | None = None,
) -> Iterator[ProgressBar | None]:
    """Create a progress bar usable both inside and outside :mod:`numba` functions."""
    if not show_progress_bar:
        yield None
        return

    from numba_progress import ProgressBar

    kwargs: dict[str, Any] = {"total": total, "unit": unit}
    if desc is not None:
        kwargs["desc"] = desc
    with ProgressBar(**kwargs) as pbar:
        yield pbar
```

timtreis:
Why did we remove the reusable machinery here?

selmanozleyen:
`thread_map` was already reused. I just made it use `numba-progress` to unify on output format

timtreis (approved):

PR #1223, opened 2026-06-20
Drop GraphMatrixT export and make GraphPostprocessor a PEP 695 type alias

selmanozleyen:
Follow-up to #1222. That PR deliberately left out two breaking changes (see its "Deliberately not done" section); now that the graph-builder API is still experimental, this applies both.

## Changes
- **Drop `GraphMatrixT` from the public API.** Removed from `__all__` in `squidpy.gr` and `squidpy.gr.neighbors`, plus the two autosummary entries in `docs/api.md`. It was only exported to back the `GraphPostprocessor` alias.
- **Convert `GraphPostprocessor` to a PEP 695 `type` alias** whose type parameter is scoped to the alias itself. This removes the now-unused module-level `GraphMatrixT = TypeVar(...)` and the `TypeVar` import.

## Breaking changes
- `from squidpy.gr import GraphMatrixT` (and `from squidpy.gr.neighbors import GraphMatrixT`) no longer works — the name is gone.
- `GraphPostprocessor`'s runtime type changes from a plain `typing` `Callable` form to `TypeAliasType`. Type-level usage (`GraphPostprocessor[csr_matrix]`, annotations) is unchanged.

The `GraphBuilder[CoordT, GraphMatrixT]` class param and `SpatialNeighborsResult[GraphMatrixT]` are independent PEP 695 params and are unaffected.

## Verification
- `ruff check src/squidpy/gr/` clean.
- Runtime asserts: `type(GraphPostprocessor).__name__ == "TypeAliasType"`, `GraphPostprocessor[csr_matrix]` still subscriptable, `GraphMatrixT` absent from both `__all__`s and no longer an attribute of `squidpy.gr.neighbors`.
- `tests/graph/test_spatial_neighbors.py`: 50 passed.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -23,7 +23,6 @@ import squidpy as sq
     gr.spatial_neighbors_radius
     gr.spatial_neighbors_delaunay
     gr.spatial_neighbors_grid
-    gr.GraphMatrixT
     gr.SpatialNeighborsResult
     gr.mask_graph
     gr.nhood_enrichment
@@ -129,7 +128,6 @@ See the {doc}`extensibility guide </extensibility>` for how to implement a custo
 
     gr.neighbors.GraphBuilder
     gr.neighbors.GraphBuilderCSR
-    gr.neighbors.GraphMatrixT
     gr.neighbors.GraphPostprocessor
     gr.neighbors.DistanceIntervalPostprocessor
     gr.neighbors.PercentilePostprocessor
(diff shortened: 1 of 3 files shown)


timtreis (approved):

selmanozleyen:
I will close this because we need to pin these for the problem that arise in the doc builds
```
"sphinx>=9",                          # was sphinx>=5.3 — PEP 695 alias autodoc support
"sphinx-autodoc-typehints>=3.10.3",   # was >=1.10.3 — renders `type X = Y` aliases correctly
```
However the extensions we use require:
- sphinx-rtd-theme (the active theme) requires <9
- nbsphinx requires <8.2
- myst-parser/myst-nb, autodocsumm, furo all require <9

So it's not possible to fix this in short term and it isn't worth it

Issue #1192, opened 2026-05-29
Refactor gr.calculate_niche into separate functions


grst:
### Description of feature

In https://github.com/scverse/squidpy/pull/1147 we refactored the `spatial_neighbors` function to address concerns raised in https://github.com/scverse/squidpy/issues/1047 and https://github.com/scverse/squidpy/issues/1102. 

I think it's a big win both in terms of documentation and extensibility. 

IMO the `gr.calculate_niche` function suffers from the same problems and would strongly benefit from exactly the same changes.

shashkat:
Totally agree! I would definitely like to contribute to this. Been exploring the niche code since some days now.

grst:
Using #1147 as a blueprint, I think it's pretty clear what to do. So I'd say please go ahead and prepare a  PR, I'm happy to help reviewing. 

Any objections @timtreis @selmanozleyen?

selmanozleyen:
> Using https://github.com/scverse/squidpy/pull/1147 as a blueprint, I think it's pretty clear what to do. So I'd say please go ahead and prepare a PR, I'm happy to help reviewing.

Same, I think we have a nice blueprint thanks to you all! I can help with the review as well

grst:
Another thought here: 
I think at least utag/neighborhood/cellcharter all have in common that they first derive some sort of feature space and then apply clustering on that feature space. 

I think it could make sense to decouple these steps for more flexibility. E.g. 
 * derive a feature space once and then apply clustering with different parameters without recomputing the feature space. 
 * plug in you own feature space
 * mix and match methodologies, e.g. use the GMM from cellcharter with the feature space from utag etc. 
 * use completely other methodologies for clustering, e.g. the [milo](https://pertpy.readthedocs.io/en/stable/tutorials/notebooks/milo.html) based clustering implemented by [quiche](https://github.com/jranek/quiche)

shashkat:
Yeah that makes sense, I will incorporate that too!

selmanozleyen:
hi @shashkat , thanks a lot but is it ok if I take over? Or do you have time for a meeting so we can cork on it together concurrently? Because we might not go the same class way here.

shashkat:
Hi @selmanozleyen , sorry about the delay. I am almost there with the submission of a draft PR. I have a class structure planned out and would love to discuss that with you too. The finer details are what I am working with right now. Is time taken more important here or the implementation structure according to you?

selmanozleyen:
[...]

PR #1229, opened 2026-06-25
bugfix in _calculate_neighborhood_profile in gr/_niche.py

shashkat:
## Description
I am currently working on refactoring the niche code and API to make more modular similar to how it was done for graph construction strategies in #1147 and discussed in #1192. However, before that, in a separate, smaller PR, I wanted to make aware of a bug in the niche code, and fix it. This is that PR.

The bug was in the function `_calculate_neighborhood_profile()` in gr/_niche.py. This is called when `calculate_niche()` is called with `flavor = 'neighborhood'`, inside `_get_nhood_profile_niches()`. The goal of `_calculate_neighborhood_profile()` is to get a dataframe of shape `num_cells x num_categories` (categories could be something like celltype), where each entry stores the number of neighbors of a cell of a particular category. However, it wasn't working correctly when different cells had different number of neighbors. I believe it was written to be meant for grid-based data only. 
Now, it handles correctly the case when cells have different number of neighbors. 

## How has this been tested?
I created a new dummy adata object (`dummy_adata2` in `tests/conftest.py`) with explicitly defined locations of the cells, not necessarily organized in a grid. Then, I determined how the neighborhood profile for this object would look like and wrote a test function accordingly in `tests/graph/test_niche.py` called `test_calculate_neighborhood_profile()`. It failed with the original version of `_calculate_neighborhood_profile()` due to incorrect values and passed with the new version.

## Other points
- The purpose of this PR is to just make aware of this bug and fix it first. But it is not gonna be the final version of this function as there would be many changes with the niche-refactor PR.
- I also believe that dummy_adata2 object can be useful for writing other tests cases too in the future for niche code and possibly graph building code too.


selmanozleyen:
closing this since it's included in https://github.com/scverse/squidpy/pull/1245

PR #1234, opened 2026-06-26
Fix: seed sequence permutation tests and make results independent of n_jobs 

### LINKED ISSUE #1232: ripley reuses same seed for every simulation once a seed is provided
When an explicit `seed` is passed, `ripley` reuses that same seed for every simulation, so all `n_simulations` draw the **identical** point pattern. The simulation envelope collapses and the p-values are degenerate (every simulation equally exceeds/does not exceed the observed statistic).

### LINKED ISSUE #1233: reproducable results indepent of `n_jobs`
### Description of feature

This feature would mean we wouldn't create same results as old versions but it would fix:
- https://github.com/scverse/squidpy/issues/1232
- Would not use sequential seeds (which might cause correlated random results)
- Threading changing the results reproducibility

Would be also the best way to switch to numba since we could expect same results when numbafied also (since each seed is determined by the source seed)

selmanozleyen:
Fixes: https://github.com/scverse/squidpy/issues/1233 and fixes https://github.com/scverse/squidpy/issues/1232 as a side effect.

For each permutation we have a separate seed. The seeds are generated by `SeeqSequence(root_seed).spawn(x)` routine. This will prevent having correlated results from sequential seeds mentioned in the [numpy docs](https://numpy.org/doc/stable/reference/random/parallel.html). I tried to use the best practice but since we need to use numba in the future our seeds can be only `Sequence[int],` instead of Sequence[np.random.SeedSequence]. Hence:
```python
def _spawn_seeds(seed: int | None, n: int) -> list[int]:
    # uint32 integer seeds, derived from independent SeedSequence children (avoids the correlation
    # of sequential seeds). uint32 is required to stay reproducible with numba/legacy RandomState,
    # which only accept uint32 integer seeds.
    return [int(s.generate_state(1)[0]) for s in np.random.SeedSequence(seed).spawn(n)]
```

Once you confirm. I will write warnings in docstrings and in notebooks about this behavioral change to users.


flying-sheep:
> I tried to use the best practice but since we need to use numba in the future our seeds can be only Sequence[int], instead of Sequence[np.random.SeedSequence]. Hence:

please explain: why does the seed have to go inside of numba? Numba supports `numpy.random.Generator` instances (to a degree).

selmanozleyen:
> please explain: why does the seed have to go inside of numba? Numba supports numpy.random.Generator instances (to a degree).

Oof, I didn't know about this. Thank you! Then we can fully modernize as well

flying-sheep:
first check if the methods you need work, I tried before and it failed since whatever distribution I tried didn’t work

selmanozleyen:
now I remember why I discarded generators. It was because it was documented that Generator was not thread safe but I guess it should be fine if we have a generator for each permutation that's used once I guess...

But another point is `ripley` uses `.uniform()`. Numba doesn't guarantee bit by bit alignment with numpy on that. But I guess we can document it once we numbafy ripley. As long as it stays reproducable per version it's good enough for me. (Which is the modern numpy Generator guarantee now anyway from what I understand...)

src/squidpy/gr/_ligrec.py
```diff
@@ -820,9 +830,8 @@ def _analysis_helper(
         - `'pvalues'` - array of shape `(n_interactions, n_interaction_clusters)`  containing `np.sum(T0 > T)`
           where `T0` is the test statistic under null hypothesis and `T` is the true test statistic.
     """
-    rs = np.random.RandomState(None if seed is None else perms[0] + seed)
-
-    clustering = clustering.copy()
+    # used as a read-only base; each permutation shuffles its own copy (see the loop below)
```
selmanozleyen, line 833: I checked if this copy would be costly or not. From the run results it shouldn't be. Also if we want to make it `n_jobs` independent it's inevitable

src/squidpy/gr/_ripley.py
```diff
@@ -48,6 +47,15 @@ def ripley(
...
     `{rp.F.s!r}`, `{rp.G.s!r}` or `{rp.L.s!r}` statistics.
 
+    .. versionchanged:: 1.8.4
+        Every simulation now uses an independent :class:`numpy.random.Generator` spawned from a
+        :class:`numpy.random.SeedSequence`, following `NumPy's guidance on parallel random
+        number generation <https://numpy.org/doc/stable/reference/random/parallel.html>`_.
+        Previously every simulation reused the same ``seed`` and therefore produced an identical
+        point pattern; the simulated statistics (and hence the p-values) are now genuinely
+        independent across simulations, but results obtained with a given ``seed`` differ from
+        those produced by squidpy < 1.8.4.
```
flying-sheep: is it intentional that this differs from the templated version?

flying-sheep (approved):

src/squidpy/gr/_ripley.py
selmanozleyen: It was, because this was a bug. But I can rephrase it perhaps to make it clear that the old version was broken statistically and this isn't just about a behaviour change warning.

selmanozleyen: ok I remembered that I also link to the ripley issue so it should be fine to just use the template

PR #1266, opened 2026-08-18
fix: reproducible niche Leiden clustering + expose flavor/random_state/n_iterations

### LINKED ISSUE #1260: Integration Testing CI Test Failure on python 3.12
The daily CI failed on test for squidpy failed.  Please go to [the logs of the integration testing repo](https://github.com/scverse/integration-testing/actions/runs/31554369860) to review. @scverse/anndata

timtreis:
Fixes the squidpy integration-CI failures in `tests/graph/test_niche.py` (scverse/squidpy#1260).

Two root causes:
- `calculate_niche_neighborhood`/`_utag` silently dropped `random_state`/`n_iterations` on the Leiden path and used the deprecated `leidenalg` backend, so niche labels were unseeded and unstable across `leidenalg`/`igraph` versions (the `assert np.False_` failure).
- The `spatialleiden` tests hard-failed with `ImportError` when the optional dep was absent.

Changes:
- Expose `flavor` (default `"igraph"`), `n_iterations` (default `-1`), `random_state` (default `0`) on `calculate_niche_neighborhood`/`calculate_niche_utag`; thread them through `_LeidenClusterer` and the deprecated `calculate_niche` umbrella (where they were previously marked "unused").
- Default the Leiden backend to `igraph` with a fixed seed -> reproducible labels. **Changes default niche labels** vs earlier releases.
- Guard the two `spatialleiden` tests with `importorskip`; regenerate the pinned exact labels for the nhood/utag/library/multipostprocessor tests.

Closes #1260

### DIFF
--- a/src/squidpy/_docs.py
+++ b/src/squidpy/_docs.py
@@ -46,16 +46,25 @@ _numba_parallel = """\
 numba_parallel
     Whether to use :func:`numba.prange` or not. If `None`, it is determined automatically.
     For small datasets or small number of interactions, it's recommended to set this to `False`."""
-_seed = """\
-seed
-    Random seed for reproducibility."""
+_rng = """\
+rng
+    Pseudorandom number generator state, following
+    `SPEC 7 <https://scientific-python.org/specs/spec-0007/>`_. When `None`, a new
+    :class:`numpy.random.Generator` is created using entropy from the operating system.
+    Types other than :class:`numpy.random.Generator` are passed to
+    :func:`numpy.random.default_rng` to instantiate a generator."""
 _seed_versionchanged = """\
 .. versionchanged:: 1.8.4
     Every permutation now uses an independent :class:`numpy.random.Generator` spawned from a
     :class:`numpy.random.SeedSequence`. Consequently the permutation-based results no
-    longer depend on ``n_jobs`` / ``backend``, but results obtained with a given ``seed``
+    longer depend on ``n_jobs`` / ``backend``, but results obtained with a given seed
     differ from those produced by squidpy < 1.8.4. See `#1232 <https://github.com/scverse/squidpy/issues/1232>`_ and
     `#1233 <https://github.com/scverse/squidpy/issues/1233>`_."""
+_rng_versionchanged = """\
+.. versionchanged:: 1.8.4
+    ``seed`` / ``random_state`` were renamed to ``rng``, which also accepts a
+    :class:`numpy.random.Generator` (`SPEC 7 <https://scientific-python.org/specs/spec-0007/>`_).
+    The old names still work but emit a :class:`FutureWarning`."""
 _n_perms = """\
 n_perms
     Number of permutations for the permutation test."""
@@ -247,14 +256,14 @@ _niche_common_params = f"""\
 {_niche_mask}
 {_library_key}
 {_niche_inplace}"""
-_niche_leiden_params = """\
+_niche_leiden_params = f"""\
 flavor
     Leiden backend passed to :func:`scanpy.tl.leiden`. Defaults to ``'igraph'``
     (the ``'leidenalg'`` backend is deprecated in scanpy).
 n_iterations
     Number of Leiden iterations. ``-1`` iterates until convergence.
-random_state
-    Random seed for the Leiden clustering, for reproducible niche labels."""
+{_rng}
+    Every resolution is clustered with an independent rng derived from it."""
 
 # static plotting docs
 _plotting_kwargs_static = """\
@@ -467,8 +476,9 @@ d = DocstringProcessor(
     copy=_copy,
     copy_cont=_copy_cont,
     numba_parallel=_numba_parallel,
-    seed=_seed,
+    rng=_rng,
     seed_versionchanged=_seed_versionchanged,
+    rng_versionchanged=_rng_versionchanged,
     n_perms=_n_perms,
     img_layer=_img_layer,
     feature_name=_feature_name,
--- a/src/squidpy/_docs.py
+++ b/src/squidpy/_docs.py
@@ -246,16 +246,12 @@ _niche_min_niche_size = """\
 min_niche_size
     Minimum number of observations required for a niche. Niches with fewer observations
     are relabeled ``'not_a_niche'``."""
-_niche_inplace = """\
-inplace
-    If `True`, modify the table in place and return `None`.
-    If `False`, return a modified copy and leave the input unchanged."""
 # the postprocessing + output params every user-facing niche function shares, in signature order
 _niche_common_params = f"""\
 {_niche_min_niche_size}
 {_niche_mask}
 {_library_key}
-{_niche_inplace}"""
+{_copy}"""
 _niche_leiden_params = f"""\
 flavor
     Leiden backend passed to :func:`scanpy.tl.leiden`. Defaults to ``'igraph'``
@@ -520,7 +516,6 @@ d = DocstringProcessor(
     niche_spatial_conn_key=_niche_spatial_conn_key,
     niche_mask=_niche_mask,
     niche_min_niche_size=_niche_min_niche_size,
-    niche_inplace=_niche_inplace,
     niche_common_params=_niche_common_params,
     niche_leiden_params=_niche_leiden_params,
     sdata_params=_sdata_params,
--- a/src/squidpy/_docs.py
+++ b/src/squidpy/_docs.py
@@ -247,6 +247,14 @@ _niche_common_params = f"""\
 {_niche_mask}
 {_library_key}
 {_niche_inplace}"""
+_niche_leiden_params = """\
+flavor
+    Leiden backend passed to :func:`scanpy.tl.leiden`. Defaults to ``'igraph'``
+    (the ``'leidenalg'`` backend is deprecated in :mod:`scanpy`).
+n_iterations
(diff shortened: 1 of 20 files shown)


src/squidpy/experimental/im/_tiling.py
```diff
@@ -414,6 +415,34 @@ def _has_distributed_client() -> bool:
     return True
 
 
+@contextlib.contextmanager
+def _local_cluster(n_workers: int) -> Iterator[Any]:
```
selmanozleyen, line 419: Hi thanks but was the cluster related additions intentional? Or just leftover since I didn't see any mentions of it

src/squidpy/_docs.py
```diff
@@ -247,6 +247,14 @@ def decorator2(obj: Any) -> Any:
 {_niche_mask}
 {_library_key}
 {_niche_inplace}"""
+_niche_leiden_params = """\
+flavor
+    Leiden backend passed to :func:`scanpy.tl.leiden`. Defaults to ``'igraph'``
+    (the ``'leidenalg'`` backend is deprecated in scanpy).
+n_iterations
+    Number of Leiden iterations. ``-1`` iterates until convergence.
+random_state
```
selmanozleyen: should we call this perhaps a seed? Because it would be nice to have consistency here no? I will also rename the other random_state appearance in compute_niche to seed before a release as well

src/squidpy/experimental/im/_tiling.py
timtreis: Should probably be redundant with #1268 now

src/squidpy/_docs.py
timtreis: I think I got random_state from scanpy/sklearn - ideally we'd use the same. But even internally, Squidpy is inconsistent right now. I'm fine with whatever but I'd learn towards ecosystem consistency.

timtreis: @flying-sheep @ilan-gold any plans on changing scanpys `random_state` any time soon?

selmanozleyen: I don't think they will change it since Phil just wrote a legacy support but we only use random_state under our experimental module and the calculate_niche which I plan to fix. I know scanpy takes it but I think it's better if we control the seed perhaps.

timtreis: I do think we should control the seed but I feel like ideally as many scverse packages share the same keywords for the same things, no?

selmanozleyen: Yeah I agree but we already have `seed` in ripley, spatial_autocorr, ligrec, and nhood_enrichment. We should be first package consistent IMO

timtreis: But if we move everything towards `seed`, we diverge more no? I'd rather use the breaking change acceptance of a 2.0 to align with the rest

selmanozleyen: > But if we move everything towards seed, we diverge more no? 

How? Nothing released uses `random_state` atm (plus the experimental module). Like divergence across packages is not big of a deal as divergence inside a package. I'd say we don't need to be like scanpy in that regard because `random_state` implies legacy np random state guarantees to me. 

> I'd rather use the breaking change acceptance of a 2.0 to align with the rest

Yes but that's a different topic no? Since random_state is something that implies legacy numpy reproducibility I am not sure if it's a good idea for us. Scanpy still gives the old reproducibility guarantees because it supports the legacy random_state. 

Like why not make it a seed now and do the breaking change all at once in a separate PR after we have this decision?

timtreis: Sorry, what's the difference between random_state and seed here? Aren't they both just numbers we propagate down through the relevant RNGs?

selmanozleyen: https://numpy.org/doc/stable/reference/random/legacy.html#numpy.random.RandomState

timtreis: hm, okay. It seems that scanpy and sklearn accept both an int or the actual state object to that parameter though, and, if passed an int just generate the obj from it. Shouldn't we just do the same then?

image

flying-sheep: > Nothing released uses `random_state` atm

Then you shouldn’t use it at all, only `rng`, see below. The only place where `random_state` should appear is when you’re interacting with `sklearn`, which is still old and bad.

> sklearn and scanpy

… behave completely differently. `sklearn` is horrible at modernizing here and annoyingly still hasn’t migrated. At this point it feels concerning that they still haven’t figured this out.

Scanpy and scipy support both legacy `random_state` (with a `FutureWarning`) and the new `rng` parameter.

> Aren't they both just numbers we propagate down through the relevant RNGs?

This shouldn’t be done for new functions. The correct final state is to only have `rng: SeedLike | RngLike | None = None`, then in each user-facing function call `rng = np.random.default_rng(rng)` to convert it into a `np.random.Generator`, which is then passed to private functions, which only accept that. The default, `None`, means “behave nondeterministically” and stability is only guaranteed when all library versions are the same.

```py
type SeedLike = int | np.integer | Sequence[int] | np.random.SeedSequence
type RNGLike = np.random.Generator | np.random.BitGenerator
```

<details><summary>If you would need to migrate functions</summary>

If you don’t want a breaking change immediately, you can as an intermediate also accept have `random_state: _LegacyRandom` (which includes seed-like and the old `RandomState`), and if the function defaults to a fixed seed, have it instead default to a `np.random.Generator` subclass derived from both parameters, behaving exactly as an old `np.random.RandomState`.

It’s quite complicated to migrate this way, as the old way of just passing seeds down the callstack and sometimes even using `np.random.seed(...)` to mutate global state is hard to emulate exactly, but both scanpy and rapids_singlecell have done it by now and you can look up how it works.

</details>

timtreis: > > sklearn and scanpy
> 
> … behave completely differently. 

from https://scanpy.readthedocs.io/en/stable/generated/scanpy.tl.umap.html
```
random_state: int | RandomState | None (default: 0)
If int, random_state is the seed used by the random number generator; If 
RandomState or Generator, random_state is the random number generator; 
If None, the random number generator is the RandomState instance used 
by np.random.
```

from https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.train_test_split.html
```
random_state: int, RandomState instance or None, default=None
Controls the shuffling applied to the data before applying the split. Pass an 
int for reproducible output across multiple function calls. 
```

I didn't bother to look under the hood but at first glance both patterns look like they capture the same "if number, make generator. if generator, use generator" idea that

> The correct final state is to only have rng: SeedLike | RngLike | None = None, then in each user-facing function call rng = np.random.default_rng(rng) to convert it into a np.random.Generator, which is then passed to private functions, which only accept that. 

would also implement? I don't really have strong opinions on how exactly this pattern is implemented, but it seems that in Squidpy we're in a situation where we partially use `seed`, partially use `random_state` and neither is what scanpy, rapids and scirpy are using. So my hope would be to just use the opportunity of our upcoming 2.0 release to transition to whatever we decide is the better cross-package way forward here 🤔 No better chance to ignore numerical backwards stability than that one 😅

selmanozleyen: again, we don't partially use `random_state`, they are all under experimental/unexposed. It's my fault I didn't bring those up in the PR's but we were actively discussing about the new rng changes meanwhile (also in annbatch). 

I know ideally rng would be takes as an arg but at least `seed` is a safe phrase we can just use it to create rng's. We never got `RandomState` classes so let's use that to our advantage

Also this is the state of `random_state` as Phil said
> Scanpy and scipy support both legacy random_state (with a FutureWarning) and the new rng parameter.

So sounds like the divergence here is "rng vs seed" rather than "random_state vs seed" even...

timtreis: Okay, but then let's look forward and just align everything to the new target pattern (with a dep warning on everything else until we move to 2.0) no? Why do it twice if we can just adapt the new standard now (not necessarily in this PR but in the very near feature).

flying-sheep: https://scanpy.scverse.org/en/latest/api/generated/scanpy.tl.umap.html

wild that this isn’t in a stable release, feels like I did this forever ago

selmanozleyen: > Why do it twice if we can just adapt the new standard now (not necessarily in this PR but in the very near feature).

Yeah I agree, I am for`rng`s but didn't think you'd care or be high priority. But one thing is for sure, we shouldn't introduce `random_state`. `rng`'s itself can be another topic/PR

timtreis: I only care because it surfaced as a point of discussion now and I'd rather deal with it now'ish or we'll forget about it and do it in 3 years 😅 So in this PR we go with `seed` to be consistent with the rest of Squidpy and then ASAP move to the new `rng` logic?

selmanozleyen: > So in this PR we go with seed to be consistent with the rest of Squidpy and then ASAP move to the new rng logic?

Internally I try to do this as much as possible. The only thing I use seed in ligrec for example is to create an rng first in the beginning. So it should be straightforward to do all these breaking changes at once. Or basically we can give a warning saying: hey seed here is only being used to create `np.default_rng(seed)` atm seed will be dropped in the future etc.

> I'd rather deal with it now'ish or we'll forget about it and do it in 3 years

Ok then let's merge https://github.com/scverse/squidpy/pull/1270 first then we can do all the public facing breaking changes in one PR

selmanozleyen (approved):

PR #1268, opened 2026-08-18
fix: reap leaked dask workers in process-based featurization

### LINKED ISSUE #1267: Process-based featurization leaks dask workers on teardown, hanging CI coverage job
### Description

The process-based featurization engine leaks worker processes on teardown. `_run_tiled(..., kind="processes")` (`src/squidpy/experimental/im/_tiling.py:486`) spins up a `dask.distributed.LocalCluster(processes=True)` in a `with` block. On exit the cluster/client teardown does not reliably reap its worker/nanny/`resource_tracker` processes. The orphans keep the parent's stdout pipe open, so any consumer waiting for EOF on that pipe never returns.

In CI this wedges the `py3.14-stable` coverage job: pytest completes normally (`1229 passed in 621.53s`), then the `run coverage tests` step hangs on the piped stdout until the job is cancelled (observed ~2.5h). Cleanup killed 4 orphan `python3` workers plus `coverage`/`uv`.

### Reproduction

```python
from squidpy.experimental.im._tiling import _run_tiled

def work(spec, *scatter):
    return spec * spec

# pipe stdout so the reader waits for EOF, mirroring the CI step
res = _run_tiled(list(range(8)), work, n_jobs=4, kind="processes", desc="tiles")
print("done", res)
```

Run piped: `python repro.py | cat`. The child returns the correct result and prints `done`, but the process does not exit and `| cat` never gets EOF (reproduced locally, 2-min hang).

### What I already ruled out

- Not a missing `.close()`: the code already uses `with LocalCluster(...) as cluster, Client(cluster) as client:`.
- `client.shutdown()` does **not** fix it; neither does explicit `client.close()` + `cluster.close()`.
- Platform-divergent manifestation: on Linux CI the main process finishes and 4 worker orphans survive holding the pipe; on macOS/py3.14 the teardown call itself blocks. Same culprit (distributed worker/nanny/`resource_tracker` teardown under `processes=True`).

### Environment

- python 3.14.6
- dask 2026.1.1
- distributed 2026.1.1

### Candidate directions

- `nanny=False` on the cluster (drop the extra nanny process layer).
- Force-terminate the worker process group after `close()`.
- Investigate a distributed/py3.14 teardown regression; pin or bump.

Not blocking a specific PR by itself, but it destabilizes the coverage job. Filing separately so the fix stays focused.

timtreis:
Fixes #1267.

`_run_tiled(..., kind="processes")` runs work on a `LocalCluster(processes=True)`. Under Python 3.14 the cluster teardown can leave worker/nanny processes alive; an orphan that inherited stdout/stderr keeps the pipe open, so a consumer waiting for EOF (the CI coverage step) hangs indefinitely.

Fix: record the cluster's own child PIDs and reap any that outlive `close()`, leaving a host app's pre-existing children untouched. `psutil` is already a `distributed` dependency.

### DIFF
--- a/src/squidpy/experimental/im/_tiling.py
+++ b/src/squidpy/experimental/im/_tiling.py
@@ -12,6 +12,7 @@ never materialize the full image or label array.
 
 from __future__ import annotations
 
+import contextlib
 from collections.abc import Callable, Sequence
 from dataclasses import dataclass
 from typing import Any, Literal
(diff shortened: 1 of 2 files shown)


selmanozleyen (approved):
process based parallelism will always be more brittle and slower than thread based one. Processes shouldn't be used in scientific algoritm computations.

That being said it's nice to have a fix

PR #1270, opened 2026-08-20
chore: switch to rngs

selmanozleyen:
This pull request refactors and standardizes the use of random seeds for reproducibility across several modules, replacing ambiguous or inconsistent `random_state` parameters with a uniform `seed` parameter. It also introduces utility functions for managing random number generators, and updates the documentation and validation logic accordingly.

These changes improve reproducibility, consistency, and clarity in the codebase's handling of random seeds and random number generation.

### DIFF
--- a/src/squidpy/_utils.py
+++ b/src/squidpy/_utils.py
@@ -237,8 +237,10 @@ def parallelize(
     return wrapper
 
 
-type SeedLike = int | np.integer | Sequence[int] | np.random.SeedSequence
-type RNGLike = np.random.Generator | np.random.BitGenerator
+# plain assignments, not PEP-695 `type` statements: those stay opaque to sphinx and would
+# render as the bare alias names on the `*Params.rng` attribute pages
+SeedLike = int | np.integer | Sequence[int] | np.random.SeedSequence
+RNGLike = np.random.Generator | np.random.BitGenerator
 
 
 def rng_to_random_state(rng: np.random.Generator) -> int:
--- a/src/squidpy/_utils.py
+++ b/src/squidpy/_utils.py
@@ -380,7 +380,7 @@ def verbosity(level: int) -> Generator[None, None, None]:
         sc.settings.verbosity = verbosity
 
 
-def deprecated_rng_param(func: Callable[..., Any]) -> Callable[..., Any]:
+def deprecated_randomness_param(func: Callable[..., Any]) -> Callable[..., Any]:
     @functools.wraps(func)
     def wrapper(*args: Any, **kwargs: Any) -> Any:
         for old in ("seed", "random_state"):
--- a/src/squidpy/_utils.py
+++ b/src/squidpy/_utils.py
@@ -243,7 +243,7 @@ SeedLike = int | np.integer | Sequence[int] | np.random.SeedSequence
 RNGLike = np.random.Generator | np.random.BitGenerator
 
 
-def rng_to_random_state(rng: np.random.Generator) -> int:
+def legacy_random(rng: np.random.Generator) -> int:
     """Draw an int seed for third-party APIs that only accept ``random_state``."""
     return int(rng.integers(np.iinfo(np.int32).max))
 
--- a/src/squidpy/_utils.py
+++ b/src/squidpy/_utils.py
@@ -237,11 +237,12 @@ def parallelize(
     return wrapper
 
 
-def spawn_generators(seed: int | None, n: int) -> list[np.random.Generator]:
-    return [np.random.default_rng(s) for s in np.random.SeedSequence(seed).spawn(n)]
+type SeedLike = int | np.integer | Sequence[int] | np.random.SeedSequence
+type RNGLike = np.random.Generator | np.random.BitGenerator
 
 
 def rng_to_random_state(rng: np.random.Generator) -> int:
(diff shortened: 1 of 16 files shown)


flying-sheep (commented):
Everything should be `rng: SeedLike | RngLike | None = None`, not `seed: Something = 0`, with the behavior and types specified here: https://scientific-python.org/specs/spec-0007/

Incompatible 3rd party APIs that only take a seed should be seeded via `third_party_api(seed=rng.bytes(16))` (in case the 3rd party API takes a 128 bit seed sequence)

For backwards compatibility, you could do the same as I did for scanpy, i.e. if nothing is passed, add a decorator that passes a compatibility `Generator` which behaves exactly as a `RandomState`. New APIs (i.e. ones that haven’t been released yet or future ones) should have the standard `None` default, i.e. behave nondeterministically.

selmanozleyen:
~~I agree but my intention isn't to convert to the SPEC-0007 conversion here. I just want to get rid of the unexposed `random_state` usages first~~

you know what, while I am at it I might as well do the full switch here

flying-sheep:
> you know what, while I am at it I might as well do the full switch here

Yeah, why rename these twice? Once to `seed` and then to `rng`. Sounds like two opportunities for bugs to sneak in.

src/squidpy/gr/_niche.py
```diff
@@ -538,14 +540,22 @@ def calculate_niche_spatialleiden(
     else:
         adata = orig_adata.copy()
 
+    # normalise once here; everything below this point works with generators only
+    generator = np.random.default_rng(rng)
```
flying-sheep: Why name it differently here? are you using `rng` elsewhere while also using `generator`? If so, why?

src/squidpy/gr/_ripley.py
```diff
@@ -45,6 +46,8 @@ def ripley(
 
     %(seed_versionchanged)s
 
+    %(rng_versionchanged)s
```
flying-sheep, line 49: are there other places where this should go? Also, is `%(seed_versionchanged)s` above still relevant?

src/squidpy/_utils.py
```diff
@@ -237,8 +237,15 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
 
-def spawn_generators(seed: int | None, n: int) -> list[np.random.Generator]:
-    return [np.random.default_rng(s) for s in np.random.SeedSequence(seed).spawn(n)]
+# plain assignments, not PEP-695 `type` statements: those stay opaque to sphinx and would
+# render as the bare alias names on the `*Params.rng` attribute pages
+SeedLike = int | np.integer | Sequence[int] | np.random.SeedSequence
+RNGLike = np.random.Generator | np.random.BitGenerator
+
+
+def rng_to_random_state(rng: np.random.Generator) -> int:
```
flying-sheep: The name implies that the return value is `np.random.RandomState`. Maybe:

```suggestion
def legacy_random(rng: np.random.Generator) -> int:
```

src/squidpy/_utils.py
```diff
@@ -373,6 +380,28 @@ def verbosity(level: int) -> Generator[None, None, None]:
         sc.settings.verbosity = verbosity
 
 
+def deprecated_rng_param(func: Callable[..., Any]) -> Callable[..., Any]:
```
flying-sheep: Not an ideal name either as it implies that `rng` is deprecated, but I can’t come up with a better one.

tests/conftest.py
```diff
@@ -339,7 +339,7 @@ def ligrec_result() -> Mapping[str, pd.DataFrame]:
         n_jobs=1,
         show_progress_bar=False,
         copy=True,
-        seed=0,
+        rng=np.random.default_rng(0),
```
flying-sheep: Why call `default_rng`? You should just be able to pass `0` right?

flying-sheep (commented):
Looks great! If you don’t have to maintain backwards compat, this is exactly how is should be done. I only have some nitpicks, mostly about naming.

src/squidpy/gr/_ripley.py
selmanozleyen: yes it is relevant because I fixed a bug which changed the rng behaviour: https://github.com/scverse/squidpy/issues/1232. So it's not only about reproducibility but warning the users that the old way might be broken. 

I am not sure if it should stay or not though. Seems like it belongs to the changelogs now.

flying-sheep: That’s what I meant!

src/squidpy/gr/_niche.py
```diff
@@ -1474,7 +1474,7 @@ class _GMMClusterer(_NicheClusterer):
     def __init__(
         self,
         n_components: int,
-        rng: np.random.Generator,
+        rng: np.random.rng,
```
flying-sheep: that’s wrong now.

```suggestion
        rng: np.random.Generator,
```

flying-sheep (approved):
Apart from one bug: looks great!

src/squidpy/gr/_niche.py
selmanozleyen: Damn sorry, that wasn't a careful find and replace

PR #1276, opened 2026-08-22
feat: automatic K selection (CellCharter's ClusterAutoK) 

### LINKED ISSUE #1244: Add automatic K selection (CellCharter's ClusterAutoK) for the cellcharter niche flavor
## Context

`sq.gr.calculate_niche(flavor="cellcharter", ...)` fits a GMM with a fixed, user-supplied `n_components`. CellCharter's own [`ClusterAutoK`](https://cellcharter.readthedocs.io/en/latest/generated/cellcharter.tl.ClusterAutoK.html) instead searches a range of K, refits the model `max_runs` times per K, scores stability via Fowlkes-Mallows similarity between labelings at K/K-1/K+1, and picks the K with highest stability (`best_k`, plus a `peaks` attribute for runner-up candidates). Squidpy's `_get_GMM_clusters` currently skips this -- its own docstring notes it's "simplified... without stability analysis."

## Proposal

Port `ClusterAutoK`-style automatic K selection, with an input signature that mirrors CellCharter's own API rather than a magic-string trigger -- i.e. take a `n_clusters` range plus the same knobs CellCharter exposes (`max_runs`, `convergence_tol`, `similarity_function`, `model_params`), so behavior and defaults line up with the original.

**Open question:** integrate this as an option inside `calculate_niche` vs. a separate standalone function (e.g. `sq.gr.cluster_auto_k`) that returns `best_k`/stability curve/peaks, to be fed into `calculate_niche` afterward. Leaning toward separate/composable, since it's a stochastic multi-run search (many GMM refits across a K range) rather than a single deterministic fit, and bolting it into `calculate_niche`'s already-branchy per-flavor validation adds complexity. Not settled yet -- needs discussion.

## Dependency

Blocked on / should follow #1192 (refactor of `calculate_niche` into separate functions) -- final placement of auto-K selection depends on how that refactor shakes out.

cc: @sarajimenez

selmanozleyen:
fixes: https://github.com/scverse/squidpy/issues/1244

will add more description but I want to start somewhere @SaraJimenez, @FrancescaDr these are all the things I could find in https://github.com/Jieran-S/hitchhikers_guide_to_spatial_project/blob/spatial_clustering/notebooks/Spatially_aware_clustering/sac.ipynb.

Anything else needed? If everything is there, how does this state look like for you?


grst:
I think a Bayesian GMM could also be very interesting for the same purpose: https://scikit-learn.org/stable/modules/generated/sklearn.mixture.BayesianGaussianMixture.html

> This class allows to infer an approximate posterior distribution over the parameters of a Gaussian mixture distribution. The effective number of components can be inferred from the data.

The downside is that it's quite slow - not sure how hard it would be to make a fast GPU implementation for that @Intron7.

src/squidpy/gr/_cluster_auto_k.py
```diff
@@ -0,0 +1,207 @@
...
+    table_key: str | None = None,
+) -> AnnData | None:
+    """Select the number of clusters (K) of a Gaussian mixture by clustering stability.
+
+    Every candidate K is fitted ``max_runs`` times, and each K is scored by how similar its
+    labeling is to the labeling of the adjacent K across runs (Fowlkes-Mallows). The most
+    stable K is the one whose partition is least sensitive to adding a cluster.
+
+    Reimplements the K selection of CellCharter's ``ClusterAutoK``, using
+    :class:`~sklearn.mixture.GaussianMixture` for the fits.
```
grst, line 48: Does this have to be tied to the GaussianMixture? We deliberately split the niche calling procedure such that arbitrary embeddings and clustering algos can be combined... In principle, why not use the same approach for Leiden?

src/squidpy/gr/_autok.py
```diff
@@ -0,0 +1,291 @@
...
+    pairs = list(zip(ks[:-1], ks[1:], strict=True))
+
+    labels_per_run: list[dict[int, np.ndarray]] = []
+    best_nll: dict[int, float] = {}
+    best_labels: dict[int, np.ndarray] = {}
+    blocks: list[list[float]] = []
+    previous_curve: np.ndarray | None = None
+    converged = False
+
+    for run in range(max_runs):
```
grst, line 257: this is embarrassingly parallel. Probably the clustering algorithm has some sort of threading, too, but did you evaluate how well it scales to many cores? Maybe parallelizing at this level could make sense.

src/squidpy/gr/_autok.py
```diff
@@ -0,0 +1,291 @@
...
+        run_labels: dict[int, np.ndarray] = {}
+        for k in ks:
+            # keyed by (run, K) rather than drawn in sequence, so appending a candidate K
+            # does not re-seed every other K's fits
+            random_state = int(np.random.SeedSequence([seed, run, k]).generate_state(1)[0])
+            labels, nll = _fit_once(X, k, random_state, model_params)
```
grst, line 263: The runs only differ at the level of random seeds. Would it make sense to combine this with e.g. bootstrapping?

src/squidpy/gr/_autok.py
selmanozleyen: I think we rely on the clustering algorithm to be utilizing the cores as much as possible so that any other threading would oversubscribe only. Also we can only parallelize runs if tolerance is 0, ie we don't early stop.

But from my small benches I didn't see much improvement than 1.5x tbh. And BLAS was already enough.

src/squidpy/gr/_cluster_auto_k.py
selmanozleyen: I didn't want to define a clusterer protocol publicly. And to have a hard-coded list of clusterers I wanted some opinions first. Bayesian GMM is a good hint

src/squidpy/gr/_autok.py
grst: I was planning to open a separate issue the next days to discuss a public "extensibility" interface for the niche algos. We can discuss it again in that context.

src/squidpy/gr/_autok.py
selmanozleyen: I am not sure, it sounds interesting but also sounds out of scope for this PR. wdyt @sarajimenez @timtreis ?

PR #1279, opened 2026-08-28
refactor: convert experimental parameters to TypedDicts (1/4)

selmanozleyen:
Hi,

I want to bring a clear separation to the kwargs madness we have here. 

- **Params**: For kw only parameter bags we expect from the user: we use this PRs suggestion.
- **Fits**: For configurations obtained by other functions: like `StainReference` or `AlignmentResult` we use frozen data classes . **Even if they are given as input** later to other functions since we don't expect the user to type those it doesn't make sense to unpack. Based on this instinct I also have a follow up PR to stack on https://github.com/scverse/squidpy/pull/1280
- **Results**: If they are simple returns only because we don't want to modify the input we can return a  `NamedTuple` (like current SpatialNeighboursResults)

(about the nesting of functions by groups I am open to discussion, but it's really unreadable if we have a flat `tl.*`. I would also argue that we want to deprecate `.im` anyway. So we might as well put everything inside `experimental.im.*` under `tl` to make it less confusing if we will do this grouping in the docs.

Intended end result after one more pr in this stack:

## Fits
Screenshot 2026-08-31 at 12 54 36 AM

## Functions
Screenshot 2026-08-31 at 12 54 24 AM

## Results
Screenshot 2026-08-31 at 12 53 57 AM

## Params

Screenshot 2026-08-31 at 12 53 42 AM

image


flying-sheep:
Nice, I like the plan and consistency.

What I don’t get is the distinction between spelling kwargs out in the function and having them defined in a TypedDict: are they always shared? Is there a rule about what goes where?

selmanozleyen:
> What I don’t get is the distinction between spelling kwargs out in the function and having them defined in a TypedDict: are they always shared? Is there a rule about what goes where?

- For stalign there is a clear win with deduplication using the inheritence and overriding. We use Unpack here

- For the cases like in `detect_tissue`, we have a bag or kwargs we sont want to bloat the signature with in the docs and the code. Honestly I haven't look into the criteria exactly. It's where @timtreis used frozen dataclasses to not bload the signature basically :D. I just converted them mechanically. We also use Unpack here.

- for cases like in FelzenszwalbParams, we dont use Unpack. But keeping them as typed dict is better than either taking a frozen dataclass vs Mapping. Bc we used to also let the user give a mapping in case they didnt want to create the class themselves.  Plus you can annotate mappings while essentially giving the `Mapping` flexibility to the users as well

selmanozleyen:
hi @timtreis, so these are what I noticed on the kwargs bags. I think it's useful to rethink them:
- `TilingQCParams` is three keys against a 14-parameter signature, so no reason it should exist. If it should, it should encapsulate more of the params.
- `detect_tissue` keeps `auto_max_pixels`, `close_holes_smaller_than_frac` and `mask_smoothing_cycles` in the signature while bagging `corner_size_pct`, those seem equally obscure to me

flying-sheep:
Very nice design, I think this 3-way split makes the most sense.

- named tuples are great return types, especially when they have 2–4 elements and can therefore be either be used as-is or unpacked as `conns, dists = spatial_neighbors(...)`
- custom classes with `.transform` is a great clean API for model fitting: `stalign_align_obs(sdata_ref, ...).transform(sdata)` makes it very obvious what’s happening.
- using typed dicts for kwargs containers makes it very frictionless to use, as users can just rely on autocompletion, or use the imported `*Params` class to make a container that is then passed or unpacked later.

  Things like pydantic would add an early layer of validation here, but lose the “just use a dict literal” advantage. Ideally there would be a way to say “a pydantic instance or a dict with matching items”, but I don’t think Python’s type system is flexible enough for that, see https://github.com/python/typing/issues/1495

----

I think there should be an indicator that the `*Params` classes are typed dicts. You could just mention it, or use `sphinx_toolbox.more_autodoc`, which looks great:

image

They invent custom object types, which isn’t ideal for interoperability (i.e. if you use it, it’ll be hard for users to link to them), but your call if that’s worth it.

Issue #1159, opened 2026-04-14
n-hop neighbors in spatial graph


grst:
### Description of feature

I think it would be useful to add a function to derive n-hop neighbors. 
In some cases, that's better than just using a larger radius since it takes gaps in the tissue properly into account. 
For instance, in the case of core needle biopsies, we often have situations like this one:

<img width="1229" height="1210" alt="Image" src="https://github.com/user-attachments/assets/95f8d246-766d-4193-8d16-f6290240e1b4" />

IMO this would be useful irrespective of the method (radius, delauney), so probably a postprocessing function?

shashkat:
Hi, I would like to give this a shot. 
Would it be fine if I try to implement it in the same fashion as the Postprocessor classes in the PR #1147 ? Specifically the `PercentilePostprocessor` and `TransformPostprocessor` classes introduced in `neighbors.py` in this commit - https://github.com/scverse/squidpy/pull/1147/commits/744c757e2825ddc525420e4d00588b4f584b05a2

grst:
Agree, it makes a lot of sense to implement this as a Postprocessor class

shashkat:
Perfect, so I guess I will wait for PR #1147 to be accepted and then start work on this, if its fine.

timtreis:
Hey @grst, could you elaborate on your sketch? What are A, B and C?

grst:
updated above, hope it's clearer now.

PR #1284, opened 2026-09-01
Incorporating feedback on niche refactor

### LINKED ISSUE #1277: Feedback on new niche implementation
Hi @shashkat and @selmanozleyen, 

thanks for refactoring the niche stuff! I wasn't available to comment on https://github.com/scverse/squidpy/pull/1245 in time, so I'll leave some implementation-specific comments in this issue, to be addressed in a follow-up PR. 

### General
* please use kwargs-only for basically all arguments except `data` for all public functions
* in general, the docstrings of the different niche methods are a bit light on details. I think they should provide a bit more details on the methodology, how they compare one to another, and reference the original publications. The docstrings should also make it clear to the user e.g. when to prefer UTAG over Neighborhood. 
* You sometimes use `logg.warning`, and sometimes `warnings.warn`. Is that intentional? 

### calculate_niche_neighborhood
* docstring `calculate_niche_neighborhood` -> specify what happens to the output if multiple `resolutions` are specified. 

### Neighborhood embedder
 * if extending n_hop_weights, I think this would most likely not intentional. I'd *at least* emit a warning, not debug, but probably just fail. 
 * `_calculate_neighborhood_profile`. I think most of the calculation could be simplified into pd.dummies + matrix multiplication (
    `(pd.get_dummies(adata.obs[self.groups]).T.values @ adata.obsp[self.spatial_connectivities_key].T).T`)
 * I wouldn't prematurely convert the profile to a pandas df as long as you still perform arithmetic operations on it. 

### UTAG
 * it should be possible to specify an AnnData layer, do not blindly use `X`. 
 * Utag embedder: sc.tl.pca can take a sparse matrix directly and return the result matrix. No need to create an intermediate AnnData object. 

### Cellcharter

 * I would move `use_rep` outside the `_CellcharterEmbedder`, i.e. intantiate the embedder only if `use_rep is None`. The embedder could be renamed to something more generic, such as `NHopPCAEmbedder`. 
 * why the `_setdiag` with `.tolil()` -> `.tocsr()` roundtrip? CSR matrices have a native `.setdiag` function. 
 * not exactly sure what cellcharter's intention with the "visited" filter is. If  a k-hop space should strictly exclude nodes that are already included in a (k-1)-hop-space , this doesn't work since it is not accounted for node degrees. The comparison adj_hop = adj_hop > adj_visitied would need to use boolean matrices instead. 

### _postprocess_niche_results

* I'd convert the result columns back to categoricals after assigning the prefixes. 

### _calculate_niche_custom

Not entirely sure about this, but could it make sense, for performance reasons, to move the embedder out of the loop? The spatial neighbohood graph should already account for libraries. The embedding could then be stored globally in adata.obsm and be inspected or potentially be reused for other purposes.

shashkat:
Fixes: #1277

## Description

Big thanks to @grst for invaluable feedback (#1277) on the niche refactor in #1245. Incorporating those suggestions in this PR.


shashkat:
Hi, @grst and @selmanozleyen, this is ready for review. I think the suggestions were all quite useful and some of them I also thought about during the refactor, but didn't implement at that time. Here are some questions/points-of-discussion from me about the same:
- Right now, for the warnings, I am switching to logg, but I wanted to ask what would be the recommended approach to do logging. In _build.py too, I see both, `logg` and `warnings` used. One thing I noticed that even though the `logg` outputs look nicer, the `logg.warning` outputs don't appear when we run tests.
- I have formalized the hopping mechanism to be able to get an adjacency matrix of first-time-visited-nodes at hop n, like it was being attempted in `CellcharterEmbedder`. This I implemented in function `_compute_hop_adjacency_matrices`, and since there was an argument for hopping also in `NhoodProfileEmbedder`, I used  `_compute_hop_adjacency_matrices` there too. However, this lead to the hopping mechanism becoming slightly different from previous in `NhoodProfileEmbedder`. Previously for hopping in `NhoodProfileEmbedder`, just the adjacency matrix was being multiplied to itself, which would not account for degree, and lead to I believe unexpected aggregation. Now, that doesn't happen, and degree is accounted for.
- I think this formalization of hopping mechanism can be useful in the implementation for #1159 too.

selmanozleyen:
> Right now, for the warnings, I am switching to logg, but I wanted to ask what would be the recommended approach to do logging. In _build.py too, I see both, logg and warnings used. One thing I noticed that even though the logg outputs look nicer, the logg.warning outputs don't appear when we run tests.

For `FutureWarnings` and deprecation stuff warning. Rest -> logg

> I have formalized the hopping mechanism to be able to get an adjacency matrix of first-time-visited-nodes at hop n, like it was being attempted in CellcharterEmbedder. This I implemented in function _compute_hop_adjacency_matrices, and since there was an argument for hopping also in NhoodProfileEmbedder, I used _compute_hop_adjacency_matrices there too. However, this lead to the hopping mechanism becoming slightly different from previous in NhoodProfileEmbedder. Previously for hopping in NhoodProfileEmbedder, just the adjacency matrix was being multiplied to itself, which would not account for degree, and lead to I believe unexpected aggregation. Now, that doesn't happen, and degree is accounted for.
> I think this formalization of hopping mechanism can be useful in the implementation for https://github.com/scverse/squidpy/issues/1159 too.

Definetely agree and it's worth doing it here I think as an unexposed function perhaps. See: https://github.com/scverse/squidpy/issues/1285#issuecomment-5526263076 would a `nhood_aggregate(hops=(1,))`  function make sense to you for this?

selmanozleyen:
Also from https://github.com/scverse/squidpy/issues/1277 @grst 

> not exactly sure what cellcharter's intention with the "visited" filter is. If a k-hop space should strictly exclude nodes that are already included in a (k-1)-hop-space , this doesn't work since it is not accounted for node degrees. The comparison adj_hop = adj_hop > adj_visitied would need to use boolean matrices instead.

 "If a k-hop space should strictly exclude nodes that are already included in a (k-1)-hop-space" I think this should be done in cell-charter if one wants to concatenate the neigbourhoods instead of aggregating. Which is what cellcharter does. I didn't get the rest of the sentence but it made me realize there might've been a bug there so I will come up with a suggestion there

grst:
I think it's a bug, but one that already was present in the original cellcharter implementation.

selmanozleyen:
> I think it's a bug, but one that already was present in the original cellcharter implementation.

Fixing it would mean something different semantically though right? Because why would I add the features I already added when I am concatenating and not reducing?

grst:
I think the semantics are rings vs. circles.

selmanozleyen:
yes but to make it a ring you'd need visited checks, otherwise your ring would include inner nodes no? I don't think cellcharter upstream had a bug. But the implementation should be based on a BFS in numba IMO rather than using matrix multiplications. Which I will show as a suggestion in this PR.

selmanozleyen:
Hi opened a PR here for my suggestions here https://github.com/shashkat/squidpy/pull/1. Which involve only using functions as embedders to reduce number of lines and to have nhood_aggregate which we can perhaps expose in the future. I'd also like to refactor the clusterers to match the sklearn API after the discussions in https://github.com/scverse/squidpy/issues/1285. Which I would also include these changes in #1276 .

I will later reuse nhood_aggregate for entropy metrics probably as well

src/squidpy/gr/_niche.py
```diff
@@ -157,12 +156,10 @@ def calculate_niche(
         If 'False', return a new AnnData object with the niche labels.
     """
 
-    warnings.warn(
```
selmanozleyen, line 160: For FutureWarnings and deprecation stuff warning. Rest -> logg

shashkat:
Hi guys, sorry about my delayed response.

I really like the idea of `nhood_aggregate`. It can effectively combine the work done by all the different embedders in one top-level function, which can possibly be made public later. This would allow us to have a "spatial embedding function" analogous to `sc.tl.pca` in single cell as mentioned here https://github.com/scverse/squidpy/issues/1285#issuecomment-5616312253. I just think the implementation of `nhood_aggregate` can possibly be made more modular (speaking wrt f64ad55e579f306b6ef26ff2499eee93a2aa0b4f) for better developer experience, but other than that I like it.

The embedders hence change from classes to functions and call `nhood_aggregate` from within, as indicated in f64ad55e579f306b6ef26ff2499eee93a2aa0b4f.

Converting the clusterers into sklearn-compatible should be straightforward I think.

This should more or less satisfy the points raised in #1285.

shashkat:
@selmanozleyen, are you taking over this PR now. Feel free to do that if you want! Let me know if you need anything from me.

selmanozleyen:
Hi @shashkat , I'd like your feedback on my changes.

So here is what I changed on top of your reviews. Note: some of these items are from AI generated output.
-  labels are category on v1.8.3 and head. main returns str, so this restores v1.8.3.
- `nhood_aggregate` and `compute_hop_adjacency_matrices` in `_nhood.py`, a numba BFS producing disjoint boolean rings. Replaces the thresholded matrix powers, which leaked degree-proportional self-loops into every
ring. `neighborhood`, `utag` and `cellcharter` all derive from it now, and `GridBuilder` uses it
for its rings too.
- **A clusterer protocol** (`ed9002e6`) — structural and runtime-checkable, so sklearn's
`GaussianMixture` satisfies it unchanged and `_GMMClusterer` goes away; `_NicheClusterer` and its
ABC go with it. One clusterer labels one column, so a resolution sweep is a mapping rather than an
internal loop. This is grst's #1285 proposal.
- **Spatialleiden stops recursing into itself** (`257c8a56`) — which is what forced `prefix` onto the
public signature, and what made the re-entry redo the sdata extraction, rng normalisation,
resolution validation and `sanitize_table` once per library.
- **Spatialleiden stops recursing into itself** (`257c8a56`) — which is what forced `prefix` onto the
public signature, and what made the re-entry redo the sdata extraction, rng normalisation,
resolution validation and `sanitize_table` once per library.


besides the changes I made on your commits here are the changes vs main and the latest released version: https://github.com/scverse/squidpy/issues/1293

(still in progress because I found some more bugs in 1.8.3)

selmanozleyen:
@timtreis @grst @shashkat , could you tell me please if `library_key` inside cellcharter would make sense or not? In original cellcharter there is no stratification. And generating `n_clusters` per `library_key` doesn't seem to make sense to me. Maybe we shouldn't expose `library_key` for `*_cellcharter`?

timtreis:
I think it'd be nice to add a "share_niches: bool" parameter to the entire function that, if library_key is preset, decides whether things are jointly modeled. But that'd go beyond the scope of this PR. That's iirc also what official Cellcharter does

grst:
IMO the neighborhood graph should respect niches, but the embedding+clustering should be across samples. That's the whole point, to classify cells into shared labels that can be used to compare between groups of samples.

shashkat:
I think with the `library_key` argument, it is a bit confusing for end user to understand how library_key interacts with the batch-aware niche clustering of cellcharter, because as grst mentioned, cellcharter's whole point is to compute niches shared across samples. Having the `share_niches` arg could be good, as it allows for both options, and the documentation of this arg will help the user understand what `library_key` is actually doing, but I am not sure when someone would need `share_niches` to be false when using `calculate_niche_cellcharter`, as likely they would be using `calculate_niche_cellcharter` because they want to perform a combined niche clustering of the data which probably has batches.

Just to reiterate, even though `cellcharter.gr.aggregate_neighbors` has a sample_key arg, it's not necessary as the spatial graph should already encode the information about batches and hence aggregation will be batch-aware. So I believe library_key isn't a necessary arg for `calculate_niche_cellcharter` and can be removed.

selmanozleyen:
> I think it'd be nice to add a "share_niches: bool" parameter to the entire function that, if library_key is preset, decides whether things are jointly modeled. But that'd go beyond the scope of this PR. That's iirc also what official Cellcharter does

Where does cellcharter provide something like this?

> IMO the neighborhood graph should respect niches, but the embedding+clustering should be across samples. That's the whole point, to classify cells into shared labels that can be used to compare between groups of samples.

Is this in general? If so I don't see a reason to require library_key then (for other flavors except spatialleiden)? because all the embedding + clustering is already across samples by default

timtreis:
> > I think it'd be nice to add a "share_niches: bool" parameter to the entire function that, if library_key is preset, decides whether things are jointly modeled. But that'd go beyond the scope of this PR. That's iirc also what official Cellcharter does
> 
> Where does cellcharter provide something like this?

It doesn't provide the option, it just jointly models it by default.


> > IMO the neighborhood graph should respect niches, but the embedding+clustering should be across samples. That's the whole point, to classify cells into shared labels that can be used to compare between groups of samples.
> 
> Is this in general? If so I don't see a reason to require library_key then (for other flavors except spatialleiden)? because all the embedding + clustering is already across samples by default

Yeah, I agree. I also wouldn't require it, the baseline assumption should be the thing that makes the most sense (joint niches here)

selmanozleyen:
Ok any reason not to remove it then for utag, cellcharter, neighbours? Because slice and run is all we do there and we can just document that if someone really wants it. Because `library_key`  is a confusing arg here atm IMO

src/squidpy/gr/_clusterers.py
```diff
@@ -0,0 +1,79 @@
...
+        """Cluster *X*, observations as rows, and return one label per row."""
+        ...
+
+    def get_params(self, deep: bool = True) -> dict[str, object]:
+        """The constructor parameters, as :func:`~sklearn.base.clone` reads them."""
+        ...
+
+    def set_params(self, *, random_state: int) -> Self:
+        """Seed the next fit; returns the estimator."""
+        ...
```
shashkat, line 31: I think that `ClusterMixin` class only implements `fit_predict` and not `get_params` and `set_params`. So the type checker is not able to categorize `LeidenClusterer` as a `Clusterer` class.

src/squidpy/_docs.py
```diff
@@ -237,20 +237,37 @@ def decorator2(obj: Any) -> Any:
...
+_niche_cluster_mask = """\
+cluster_mask
+    Boolean :class:`pandas.Series` indexed like :attr:`anndata.AnnData.obs`. ``False``
+    observations are labeled ``'not_a_niche'`` and take no part in the clustering, though they
+    still reach their neighbors through the graph. Observations it omits are kept. To drop them
+    from the graph as well, subset and rebuild it with :func:`~squidpy.gr.spatial_neighbors`."""
+_niche_key_added_stem = """\
+key_added
+    Stem of the :attr:`anndata.AnnData.obs` columns the labels are written to, one per resolution,
+    named ``f"{key_added}_res={resolution}"``."""
```
shashkat, line 266: Not sure why, but I got the following warnings from docrep when calling `calculate_niche_neighborhood()`
```py
/Users/shashankkatiyar/Library/Application Support/hatch/env/virtual/squidpy/0otTDom-/squidpy/lib/python3.12/site-packages/docrep/decorators.py:43: SyntaxWarning: 'niche_library_key' is not a valid key!
doc = func(self, args[0].__doc__, *args[1:], **kwargs)
/Users/shashankkatiyar/Library/Application Support/hatch/env/virtual/squidpy/0otTDom-/squidpy/lib/python3.12/site-packages/docrep/decorators.py:43: SyntaxWarning: 'niche_cluster_mask' is not a valid key!
	doc = func(self, args[0].__doc__, *args[1:], **kwargs)
/Users/shashankkatiyar/Library/Application Support/hatch/env/virtual/squidpy/0otTDom-/squidpy/lib/python3.12/site-packages/docrep/decorators.py:43: SyntaxWarning: 'niche_key_added_stem' is not a valid key!
	doc = func(self, args[0].__doc__, *args[1:], **kwargs)
```

shashkat (commented):
Other than these things, it looks good to me!

Just one more thing: `fast-array-utils.types` has `HasArrayNamespace` only from v1.5 onwards. So we may need to add the corresponding pin in the pyproject.yaml file.

src/squidpy/gr/_clusterers.py
grst: But `BaseEstimator` implements them, doesn't it?

src/squidpy/gr/_clusterers.py
```diff
@@ -0,0 +1,79 @@
...
+            kwargs["directed"] = False
+        sc.tl.leiden(
+            shell,
+            resolution=self.resolution,
+            key_added="niche",
+            random_state=self.random_state,
+            **kwargs,
+        )
+
+        self.labels_ = shell.obs["niche"].to_numpy()
```
grst, line 78: This creates labels with object dtype. Maybe `.cat.codes.to_numpy()` would be more appropriate?

src/squidpy/gr/_nhood.py
```diff
@@ -822,3 +826,258 @@ def _build_shuffle_groups(
     group_indices = np.argsort(codes, kind="stable").astype(np.int64)
     group_offsets = np.concatenate(([0], np.cumsum(np.bincount(codes, minlength=n_groups)))).astype(np.int64)
     return group_offsets, group_indices
+
+
+@njit(inline="always", cache=True)
```
grst, line 831: Have you benchmarked if the numba approach is actually faster than the matrix multiplication approach?

src/squidpy/gr/_nhood.py
```diff
@@ -822,3 +826,258 @@ def _build_shuffle_groups(
...
+            new_tail += 1
+    return new_tail
+
+
+@njit(parallel=True, cache=True)
+def _bfs_shells(
+    indptr: NDArrayA,
+    indices: NDArrayA,
+    max_hop: int,
+    n_threads: int,
```
grst, line 864: is it required to pass that? should be already covered by `with numba_threads` below?

src/squidpy/gr/_nhood.py
```diff
@@ -822,3 +826,258 @@ def _build_shuffle_groups(
...
+
+def compute_hop_adjacency_matrices(
+    adjacency_matrix_orig: CSBase | Array,
+    max_hop: int,
+    n_jobs: int | None = None,
+) -> list[CSBase]:
+    """Disjoint adjacency rings, one per hop up to *max_hop*.
+
+    Ring ``k`` holds the pairs first reached at hop ``k + 1``, so the rings never restate
+    each other. Ring 0 is the input as booleans, self-loops included.
```
grst, line 905: @marcovarrone, reaching out to you as one of the original cellcharter authors here. 

Context: we are working on native implementation of common niche-calling algorithms in squidpy, including cell-charter. The plan is to make the implementation modular by decomposing the neighborhood embedding and the clustering step, as these are often shared between different approaches. 

I understand cellcharter builds an embedding by aggregating the spatial neighborhood at different levels. What I'm unsure: did you intend to aggregate growing circles, or disjoint rings? I.e. does the layer 2 include all cells from layer1 and so on?

src/squidpy/gr/_nhood.py
```diff
@@ -822,3 +826,258 @@ def _build_shuffle_groups(
...
+def _assert_hop_request(adata: AnnData, connectivity_key: str, hops: Sequence[int]) -> None:
+    """Verify a hop request against the graph it is about to run on."""
+    _assert_connectivity_key(adata, connectivity_key)
+    if len(hops) == 0:
+        raise ValueError("'hops' must name at least one hop")
+    if any(hop < 0 for hop in hops):
+        raise ValueError(f"'hops' must be non-negative, got {list(hops)!r}")
+
+
+def nhood_aggregate(
```
grst, line 1017: I would suggest to expose this function publicly. It's useful for custom niche building workflows and I could also imagine an embedded neighborhood space to be useful for other things, like batch correction, similar to the latent space that is used in single-cell workflows.

src/squidpy/gr/_nhood.py
```diff
@@ -822,3 +826,258 @@ def _build_shuffle_groups(
...
+
+def _onehot(labels: pd.Series) -> csr_matrix:
+    """Indicator matrix of ``labels``, one column per category."""
+    cat = labels.astype("category")
+    codes = cat.cat.codes.to_numpy()
+    keep = codes >= 0
+    return csr_matrix(
+        (np.ones(keep.sum(), dtype=np.float64), (np.flatnonzero(keep), codes[keep])),
+        shape=(len(codes), len(cat.cat.categories)),
+    )
```
grst, line 968: pd.get_dummies(..., sparse=True) should do the same as this function, no?

src/squidpy/gr/_nhood.py
```diff
@@ -822,3 +826,258 @@ def _build_shuffle_groups(
...
+    if any(hop < 0 for hop in hops):
+        raise ValueError(f"'hops' must be non-negative, got {list(hops)!r}")
+
+
+def nhood_aggregate(
+    adata: AnnData,
+    *,
+    groups: str | None = None,
+    use_rep: str | None = None,
+    layer: str | None = None,
```
grst, line 1022: This could probably be replaced by one single parameter with an AnnData accessor? Do we want to do that already now? Would require AnnData 0.13 though

src/squidpy/gr/_nhood.py
```diff
@@ -822,3 +826,258 @@ def _build_shuffle_groups(
...
+    if len(given) > 1:
+        raise ValueError(f"pass at most one of 'groups', 'use_rep' and 'layer', got {given}")
+
+    # `has_value` says which observations have something to contribute; only a category can
+    # be unassigned, so a feature matrix leaves every row valid
+    has_value = None
+    if groups is not None:
+        assert_key_in_adata(adata, groups, attr="obs")
+        features = _onehot(adata.obs[groups])
+        has_value = np.asarray(features.sum(axis=1)).ravel() != 0
```
grst, line 1052: wouldn't that always be true for all rows of a one-hot-coded array?

src/squidpy/gr/_nhood.py
```diff
@@ -822,3 +826,258 @@ def _build_shuffle_groups(
...
+
+def _power_adjacencies(adj: CSBase, max_hop: int) -> list[CSBase]:
+    if max_hop < 1:
+        raise ValueError(f"max_hop must be >= 1, got {max_hop}.")
+
+    adjacencies, power = [adj], adj
+    for _ in range(1, max_hop):
+        power = power @ adj
+        adjacencies.append(power)
+    return adjacencies
```
grst, line 957: Just one thought, but that's more likely for a future iteration: The n-hop adjacency matrices can become quite large as the sparsity reduces... Maybe we can come up with a smarter solution (maybe numba-based) that doesn't require to keep the entire adjaceny matrix in memory at the same time.

src/squidpy/gr/_nhood.py
```diff
@@ -822,3 +826,258 @@ def _build_shuffle_groups(
...
+        # summing, so by the time an integer product exists the counts are already gone
+        if not np.issubdtype(x.dtype, np.floating):
+            x = x.astype(np.float64)
+        product = adj @ x
+        if issparse(product):
+            product = product.tocsr()
+            scale = np.repeat(inv.ravel(), np.diff(product.indptr))  # each entry against its own row
+            product.data = np.multiply(product.data, scale, dtype=np.float64).astype(product.dtype, copy=False)
+            return product
+        return np.multiply(product, inv, out=product, casting="same_kind")
```
grst, line 997: Could this be replaced by fast_array_utils.stats.mean(..., axis=1)?

src/squidpy/gr/_nhood.py
```diff
@@ -822,3 +826,258 @@ def _build_shuffle_groups(
...
+
+    if aggregation == "mean":
+        return mean_over(features)
+    if aggregation == "variance":
+        mean = to_dense(mean_over(features))
+        dense = to_dense(features)
+        return to_dense(mean_over(dense * dense)) - mean * mean
```
grst, line 1004: ... and possibly `fast_array_utils.stats.mean_var`?

grst (commented):
First round of comments, for now focussing on `_nhood.py`. Still need to take a look at `_niche.py`.

selmanozleyen:
Thanks for the reviews guys! After talking to @timtreis yesterday I realized spatiallieden doesn't also need `library_key`, so updates are on its way

src/squidpy/_docs.py
selmanozleyen: I can't reproduce this at 66769650. Importing squidpy and calling `calculate_niche_neighborhood` is clean even with `-W error::SyntaxWarning`, and no `%(...)s` is left unsubstituted in the rendered docstrings.

The three keys in your warnings are the three templates added most recently to `_docs.py`, so my guess is the env is importing an older `_docs.py` than the `_niche.py` it decorates. Could you check `squidpy._docs.__file__`, and whether that file has `_niche_cluster_mask`? Recreating the hatch env should fix it if so.

src/squidpy/gr/_nhood.py
selmanozleyen: I will leave the accessor usage for later as it needs it's own discussion

src/squidpy/gr/_nhood.py
selmanozleyen: But the mean over here is not an axis, it's per each cells' neighbours

grst: But isn't that a mean over an axis of the `adj @ x` product? 
Agree my code selection was too broad, was mostly wondering if we could get rid of the sparse special casing here by delegating to fau.

src/squidpy/gr/_nhood.py
selmanozleyen: It does but it's just so slow

```python
n          k    _onehot   get_dummies
200k       10     1.1ms      25.9ms
200k       50     1.1ms      28.1ms
1M        200     5.4ms     193.2ms
```

Maybe I can add it to fau later?

src/squidpy/gr/_nhood.py
selmanozleyen: good catch! it was required for the buffer size but we can use numba.get_num_threads inside

src/squidpy/gr/_nhood.py
grst: forget what I said...

